PR #61, opened 2023-09-27
Fix wandb

jewelltaylor:
# PR Type
[Feature | **Fix** | Documentation | Other() ]

# Short Description
[Clickup Story](https://app.clickup.com/t/8685pw7u1):  In the refactor of the client, our WANDB integration broke. This is a simple PR to add it back. It is based on the original implementation from the fedprox client but I updated it to decouple logging and reporting. 

Did a quick run of the fedprox example with both steps and epochs and the WANDB reporter seems to be working just like before. 

# Tests Added
None for this PR! If you have any ideas, let me know!

### DIFF
--- a/examples/fedopt_example/metrics.py
+++ b/examples/fedopt_example/metrics.py
@@ -1,11 +1,14 @@
 import json
-from typing import Dict, List
+from typing import Dict, List, TypeVar
 
-import torch
+import numpy as np
 from flwr.common.typing import Metrics
 from sklearn.metrics import confusion_matrix
+from torch import Tensor
 
 from examples.fedopt_example.client_data import LabelEncoder
+
+T = TypeVar("T", np.ndarray, Tensor)
 
 
 class Outcome:
@@ -107,8 +110,7 @@ class ClientMetrics:
         log_string = f"{log_string}\naverage_f1:{str(sum_f1/n_topics)}"
         return log_string
 
-    def update_performance(self, predictions: torch.Tensor, labels: torch.Tensor) -> None:
-        predictions, labels = predictions.numpy(), labels.numpy()
+    def update_performance(self, predictions: T, labels: T) -> None:
         confusion = confusion_matrix(labels, predictions, labels=range(self.n_classes))
         for i in range(self.n_classes):
             true_class = self.label_to_class[i]
--- a/examples/fedopt_example/metrics.py
+++ b/examples/fedopt_example/metrics.py
@@ -1,7 +1,7 @@
 import json
 from typing import Dict, List
 
-import numpy as np
+import torch
 from flwr.common.typing import Metrics
 from sklearn.metrics import confusion_matrix
 
@@ -107,7 +107,7 @@ class ClientMetrics:
         log_string = f"{log_string}\naverage_f1:{str(sum_f1/n_topics)}"
         return log_string
 
-    def update_performance(self, predictions: np.ndarray, labels: np.ndarray) -> None:
+    def update_performance(self, predictions: torch.Tensor, labels: torch.Tensor) -> None:
         confusion = confusion_matrix(labels, predictions, labels=range(self.n_classes))
         for i in range(self.n_classes):
             true_class = self.label_to_class[i]
--- a/examples/fedopt_example/metrics.py
+++ b/examples/fedopt_example/metrics.py
@@ -108,6 +108,7 @@ class ClientMetrics:
         return log_string
 
     def update_performance(self, predictions: torch.Tensor, labels: torch.Tensor) -> None:
+        predictions, labels = predictions.numpy(), labels.numpy()
         confusion = confusion_matrix(labels, predictions, labels=range(self.n_classes))
         for i in range(self.n_classes):
             true_class = self.label_to_class[i]
--- a/examples/fedprox_example/client.py
+++ b/examples/fedprox_example/client.py
@@ -52,7 +52,7 @@ if __name__ == "__main__":
     log(INFO, f"Device to be used: {DEVICE}")
     log(INFO, f"Server Address: {args.server_address}")
 
-    client = MnistFedProxClient(data_path, [Accuracy()], DEVICE)
+    client = MnistFedProxClient(data_path, [Accuracy()], DEVICE, use_wandb_reporter=True)
     fl.client.start_numpy_client(server_address=args.server_address, client=client)
 
     # Shutdown the client gracefully
--- a/examples/fedprox_example/client.py
+++ b/examples/fedprox_example/client.py
@@ -52,7 +52,7 @@ if __name__ == "__main__":
     log(INFO, f"Device to be used: {DEVICE}")
     log(INFO, f"Server Address: {args.server_address}")
 
-    client = MnistFedProxClient(data_path, [Accuracy()], DEVICE, use_wandb_reporter=True)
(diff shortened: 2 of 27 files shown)


examples/fedprox_example/client.py
```diff
@@ -52,7 +52,7 @@ def get_criterion(self, config: Config) -> _Loss:
     log(INFO, f"Device to be used: {DEVICE}")
     log(INFO, f"Server Address: {args.server_address}")
 
-    client = MnistFedProxClient(data_path, [Accuracy()], DEVICE)
+    client = MnistFedProxClient(data_path, [Accuracy()], DEVICE, use_wandb_reporter=True)
```
emersodb, line 55: Rather than having to set the `use_wandb_reporter` I think we can remove this boolean from the `BasicClient` and rely on the server config to send over the proper components. That is, in BasicClient, rather than doing 
```
if self.use_wandb_reporter:
```
we can just do 
```
```
Since the `from_config` function handles creating from the passed config. We could simply augment the from config function to be more robust. That is, 
```
@classmethod
def from_config(cls, client_name: str, config: Dict[str, Any]) -> Optional["ClientWandBReporter"]:
        if "reporting_enabled" in config and config["reporting_enabled"]
            return ClientWandBReporter(client_name, config["project_name"], config["group_name"], config["entity"])
        else:
            return None
```

fl4health/clients/basic_client.py
```diff
@@ -67,9 +73,13 @@ def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, N
         if ("local_epochs" in config) and ("local_steps" in config):
             raise ValueError("Config cannot contain both local_epochs and local_steps. Please specify only one.")
         elif "local_epochs" in config:
+            # Initialize total_epochs attribute with 0 if None
+            self.total_epochs = 0 if self.total_epochs is None else self.total_epochs
```
emersodb, line 77: I didn't really handle this in my original implementation, but it occurred to me that with this setup, we're assuming that the config never mixes epochs and steps. That is, we could do 2 epochs for the first 10 server rounds and then 100 steps for the next 10. I think the rest of the code functions perfectly fine with that, but we'll have some issues here. 

Maybe we should simply keep track of total steps and log that regardless of whether we're doing epochs or steps?

fl4health/clients/basic_client.py
```diff
@@ -140,6 +150,30 @@ def _handle_logging(
...
+        reporting_dict: Dict[str, Any] = {"server_round": current_round}
+        reporting_dict = (
+            {**reporting_dict, "epoch": self.total_epochs} if self.total_epochs is not None else reporting_dict
+        )
+        reporting_dict = (
+            {**reporting_dict, "step": self.total_steps} if self.total_steps is not None else reporting_dict
+        )
+        reporting_dict.update(loss_dict)
+        reporting_dict.update(metric_dict)
+        self._maybe_report_metrics(reporting_dict)
```
emersodb, line 175: Because logging and reporting were coupled together, we were forming this dictionary regardless of whether the reporter was defined or not. I think we can bring the `_maybe_report_metrics(...)` up into this class. So this function would be 
```
    def _maybe_report_metrics(
        self,
        loss_dict: Dict[str, float],
        metric_dict: Dict[str, Scalar],
        current_round: Optional[int] = None,
    ) -> None:
```
and we'd only do all of this work to create the dictionary if the reporter is not None. What do you think?

fl4health/clients/basic_client.py
```diff
@@ -187,7 +221,13 @@ def train_by_epochs(
             losses = self.train_loss_meter.compute()
             loss_dict = losses.as_dict()
 
-            self._handle_logging(loss_dict, metrics, current_epoch=local_epoch, current_round=current_round)
+            # Ensure total_epochs is not None and increment
+            assert self.total_epochs is not None
+            self.total_epochs += 1
```
emersodb, line 226: If we were to unify on `self.total_steps` as our tracker, this would just be bumped into the `train_loader` loop to increment `train_steps`

emersodb:
I'm not sure how viable it would be, but one small test that we could do is to create a WandB reporter and ensure that the `local_log_directory` is created in a temp directory along with the proper artifacts? This would be similar to the flow for our checkpointing tests where we use something like `def test_save_and_load_checkpoint(tmp_path: Path) -> None:` 

If you don't think that's a good idea or useful, it need not be implemented.

examples/fedprox_example/client.py
jewelltaylor: Updated in [c9a023a](https://github.com/VectorInstitute/FL4Health/pull/61/commits/c9a023ab5d1f19dce69e969d263f81b47b112cc8)!

fl4health/clients/basic_client.py
jewelltaylor: Updated in [31d94bf](https://github.com/VectorInstitute/FL4Health/pull/61/commits/31d94bf2464f27115b4672a5fae1a25a9c95e7fb)!

fl4health/clients/basic_client.py
jewelltaylor: Updated in [028422b](https://github.com/VectorInstitute/FL4Health/pull/61/commits/028422b3cc4efdce5bf2e37afcbc3ef98bfd4e24)!

fl4health/clients/basic_client.py
jewelltaylor: Updated in [31d94bf](https://github.com/VectorInstitute/FL4Health/pull/61/commits/31d94bf2464f27115b4672a5fae1a25a9c95e7fb)!

jewelltaylor:
> I'm not sure how viable it would be, but one small test that we could do is to create a WandB reporter and ensure that the `local_log_directory` is created in a temp directory along with the proper artifacts? This would be similar to the flow for our checkpointing tests where we use something like `def test_save_and_load_checkpoint(tmp_path: Path) -> None:`
> 
> If you don't think that's a good idea or useful, it need not be implemented.

Good idea! There was a little bit of trickiness in handling the constructor because we can't run it without a valid wandb config. So I patched the constructor and we can test certain methods individually. Should be good for another pass through now!

examples/fedopt_example/metrics.py
```diff
@@ -107,7 +107,8 @@ def summarize(self) -> str:
         log_string = f"{log_string}\naverage_f1:{str(sum_f1/n_topics)}"
         return log_string
 
-    def update_performance(self, predictions: np.ndarray, labels: np.ndarray) -> None:
+    def update_performance(self, predictions: torch.Tensor, labels: torch.Tensor) -> None:
```
jewelltaylor, line 110: For some reason, this started throwing pre-commit issue (correctly asserting this function is being passed torch tensor but the types of the argument were numpy ndarrays) even though it had previously passed and there was nothing changed in my code that would cause it to fail all of the sudden. I noticed this came up in Claire's PR's pre-commit checks on the github server too. I suspect since we dont have versions pinned, when we setup the environment, its possible to have inconsistent versions, leading to potentially inconsistent pre-commit behaviour. I could be wrong though, if you have any ideas let me know.

emersodb: Yeah I think it has to do with the version not being pinned. It's shown up for my PR and Yuchong's as well. However, it's hard to argue that it doesn't make sense. Yuchong's fix is slightly more general. Maybe we can just adopt that one?

emersodb (approved):
Changes look good to go for me.

PR #62, opened 2023-09-29
Update apfl client

jewelltaylor:
# PR Type
[**Feature**| Fix | Documentation | Other() ]

# Short Description
[Clickup Story](https://app.clickup.com/t/8685prd50): This story is a follow up from the client refactor to have `ApflClient` inherit from `BasicClient`. I made the call not to record the losses and metrics of the global and local models, only the personal model. This is in line with FENDA and simplifies things a lot.  I had to make some small changes to the `ApflModule` along the way.

Just wanted to get this up quickly before going on vacation next week :) No rush reviewing it. 

# Tests Added
- Test to ensure the local and global optimizers are being set correctly 
- Modifying and adding to test verifying the output of the personal, global and local models for Apfl


examples/apfl_example/client.py
```diff
@@ -1,41 +1,38 @@
...
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler)
+        return train_loader, val_loader
 
-        self.train_loader, self.val_loader, self.num_examples = load_mnist_data(self.data_path, batch_size, sampler)
-        self.parameter_exchanger = FixedLayerExchanger(self.model.layers_to_exchange())
+    def get_model(self, config: Config) -> nn.Module:
+        return APFLModule(MnistNetWithBnAndFrozen()).to(self.device)
 
-        super().setup_client(config)
+    def get_optimizer(self, config: Config) -> Optimizer:
```
emersodb, line 31: Theoretically, might the global and local optimizers be different for APFL? Maybe that's not admissible under their theory, but our previous implementation didn't enforce that they are the same.

One way we could handle this with a bit more generality is to rename `get_optimizer` to `get_optimizers` and have it return a dictionary of optimizers. For single optimizer implementations, this would just be keyed something like "optimizer" and for apfl (or other double optimizer approaches) this would be "global_optimizer" and "local_optimizer". That way we don't have to overload the idea of optimizer and they could be different. 

The property `self.optimizer` could also be renamed to `self.optimizers: Dict[str, torch.nn.Optimizer]` and basic client would simply used self.optimizers['optimizer'] to do the backprop, where as APFL would make use of two optimizers stored in the dictionary.

We could also get a little trickier and create classes with a base class returned for get_optimizer

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,101 @@
...
+        self.local_optimizer: Optimizer
+
+    def is_start_of_local_training(self, step: int) -> bool:
+        return step == 0
+
+    def update_after_step(self, step: int) -> None:
+        if self.is_start_of_local_training(step) and self.model.adaptive_alpha:
+            self.model.update_alpha()
+
+    def split_optimizer(self, global_optimizer: Optimizer) -> Tuple[Optimizer, Optimizer]:
```
emersodb, line 46: I like this idea and it's a great manipulation of the optimizers. However, I wonder if we're overloading the self.optimizer property here a bit by "assuming" or forcing it to be the global optimizer. That is, from an un-aware user, who knows APFL but not our library, it might be surprising to define only a single optimizer where they know that two optimizers **must** be used for APFL. I made a suggestion that we shift from having get_optimizer from returning a single optimizer to a dictionary. We could also consider defining a custom class that may have more than one optimizer in it.

examples/apfl_example/client.py
```diff
@@ -1,41 +1,38 @@
 import argparse
```
emersodb, line 1: I didn't realize this until looking at it, but we can eliminate the custom `def fit` method in FedProx by defining 
```
# Store current loss which is the vanilla loss without the proximal term added in
self.current_loss = loss_dict["checkpoint"]
```
in an `update_after_train` function correct?

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,101 @@
...
-        personal_loss = self.criterion(personal_pred, target)
-        personal_loss.backward()
+        losses = self.compute_loss(personal_pred, target)
+        losses.backward.backward()
         self.local_optimizer.step()
 
-        with torch.no_grad():
-            local_loss = self.criterion(local_pred, target)
-
-        return local_loss, global_loss, personal_loss, local_pred, global_pred, personal_pred
```
emersodb, line 129: Could we not fold the local and global losses into the Losses class as auxiliary losses as done with the proximal term for FedProx? I found these losses to be quite useful in debugging and understanding the training of APFL for the FLamby experiments.

emersodb: That is, defining a custom `compute_loss` as done for FedProx?

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,101 @@
...
-
-    def evaluate(self, parameters: NDArrays, config: Config) -> Tuple[float, int, Dict[str, Scalar]]:
-        if not self.initialized:
-            self.setup_client(config)
-
-        self.set_parameters(parameters, config)
-        # Default APFL uses an average meter
-        global_meter = MetricAverageMeter(self.metrics, "global")
-        local_meter = MetricAverageMeter(self.metrics, "local")
-        personal_meter = MetricAverageMeter(self.metrics, "personal")
```
emersodb, line 75: Losing these individual meters is a bit of a shame. As stated for the global, local, and personal losses below, these meters helped me understand what was happening for APFL under the hood, so to speak, during training (i.e. whether it made sense that alpha was moving one way or another).

Could we define a `Meters` class, similar to the Losses class, but that contains Meter objects and work with that in the BasicClient?

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,101 @@
...
         # to update the scalar alpha (see update_alpha() where .grad is called.)
-        self.global_optimizer.zero_grad()
+        self.optimizer.zero_grad()
         self.local_optimizer.zero_grad()
 
         # Personal predictions are generated as a convex combination of the output
         # of local and global models
-        pred_dict = self.model(input, personal=True)
-        personal_pred, local_pred = pred_dict["personal"], pred_dict["local"]
+        personal_pred = self.predict(input)
```
emersodb, line 91: I think this might be another situation where that might be a little confusing to a user who doesn't know the internals of our APFL implementation. They might be asked to define a predict but know that there are a few "predict" modes for APFL. One thing we could do, is have predict take a config dictionary that defaults to `{}`. Then we could implement it here where it is something like
```
def predict(input: torch.Tensor, auxiliary_args: Dict[str, Scalar]) -> torch.Tensor:
    if auxiliary_args['global_forward']:
        return self.model.global_forward(input)
    else:
        return self.model(input)
```
Just a thought.

fl4health/clients/basic_client.py
```diff
@@ -178,11 +178,13 @@ def train_by_epochs(
         for local_epoch in range(epochs):
             self.train_metric_meter.clear()
             self.train_loss_meter.clear()
-            for input, target in self.train_loader:
+            for step, (input, target) in enumerate(self.train_loader):
                 input, target = input.to(self.device), target.to(self.device)
                 losses, preds = self.train_step(input, target)
                 self.train_loss_meter.update(losses)
                 self.train_metric_meter.update(preds, target)
+                actual_step = int(local_epoch * len(self.train_loader) + step)
```
emersodb, line 186: I don't think this calculation is wrong or anything, but, rather than performing this calculation each time, would it not be simpler to implement a step counter that is, above line 178 do `local_steps=0`
and inside the for loop just do `local_steps+=1` and send local_steps to `self.update_after_step(local_steps)`

emersodb (commented):
I think most of the changes look good, but I think we might want to generalize several of the functions to better admit the auxiliary pieces that APFL produces, rather than cutting them out. For example the auxiliary losses could be computed in a similar way to FedProx. I also think there are a few places we can expand the definitions of BasicClient functions (get_optimizer, predict) that would help APFL fit better into the flow, while minimally affecting our other implementations.

examples/apfl_example/client.py
jewelltaylor: Good catch, updated in [d50ae1e](https://github.com/VectorInstitute/FL4Health/pull/62/commits/d50ae1eb23a57673481fd7322694a0d0d11ed0b9)!

fl4health/clients/apfl_client.py
jewelltaylor: Updated in https://github.com/VectorInstitute/FL4Health/pull/62/commits/f462d69c915f7f19a2d97d808e685d1d0ad39dd6

fl4health/clients/apfl_client.py
jewelltaylor: Updated in https://github.com/VectorInstitute/FL4Health/pull/62/commits/f462d69c915f7f19a2d97d808e685d1d0ad39dd6

fl4health/clients/apfl_client.py
jewelltaylor: Updated predict to return either torch tensor or dict of torch tensor if there is multiple prediction types. This ended up working nicely in having changing minimal code. Let me know if this is okay f462d69](https://github.com/VectorInstitute/FL4Health/pull/62/commits/f462d69c915f7f19a2d97d808e685d1d0ad39dd6)

fl4health/clients/basic_client.py
jewelltaylor: Updated in [349c7cb](https://github.com/VectorInstitute/FL4Health/pull/62/commits/349c7cb7ba64abfdb91b1e5c77eca6d7ac30c789)

fl4health/clients/apfl_client.py
jewelltaylor: Updated in [61693e7](https://github.com/VectorInstitute/FL4Health/pull/62/commits/61693e7a371f5a3b4a0ab9237a87cc0dcf339b2d)

examples/apfl_example/client.py
jewelltaylor: Updated in [61693e7](https://github.com/VectorInstitute/FL4Health/pull/62/commits/61693e7a371f5a3b4a0ab9237a87cc0dcf339b2d). Settled on a Union of optimizer and dict of str Optimizer as per our discussion on slack. Let me know if you have any issues with this.

examples/fedprox_example/config.yaml
```diff
@@ -16,7 +16,7 @@ local_epochs: 1 # The number of epochs to complete for client
 batch_size: 128 # The batch size for client training
 
 reporting_config:
-  enabled: True
+  enabled: False
```
emersodb, line 19: I assume you're setting this to false to make testing a bit less annoying? I'm good with that, just wanted to make sure it was intentional.

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,117 @@
...
-    def train_step(self, input: torch.Tensor, target: torch.Tensor) -> ApflTrainStepOutputs:
+
+        # local_optimizer is used on the local model
+        # Usual self.optimizer is used for global model
+        self.local_optimizer: Optimizer
+
+    def is_start_of_local_training(self, step: int) -> bool:
+        return step == 0
+
+    def update_after_step(self, step: int) -> None:
```
emersodb, line 39: Do you mind filling in a bit of a comment here? Maybe note that the `step` passed here is assumed to represent the overall number of local training steps taken by the client in the current server round?

examples/fedprox_example/config.yaml
jewelltaylor: Yeah, just because by default we dont have a valid entity name. I just set it to False so it does not run.

fl4health/clients/apfl_client.py
jewelltaylor: Added/Updated some more comments in [e15e2f8](https://github.com/VectorInstitute/FL4Health/pull/62/commits/e15e2f8a3dac6f637c956cdbabf3265a1af6bd41)

jewelltaylor:
@emersodb I made the changes to the MetricMeterManager and its initialization in the clients as we discussed this morning. Let me know if thing look okay

examples/apfl_example/client.py
```diff
@@ -1,41 +1,44 @@
...
-        data_path: Path,
-        metrics: Sequence[Metric],
-        device: torch.device,
-    ) -> None:
-        super().__init__(data_path=data_path, metrics=metrics, device=device)
-
-    def setup_client(self, config: Config) -> None:
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
         batch_size = self.narrow_config_type(config, "batch_size", int)
         self.model: ApflModule = ApflModule(MnistNetWithBnAndFrozen()).to(self.device)
```
emersodb, line 24: Correct me if I'm wrong, but it looks like we're setting a bit of extra stuff here in get_data_loaders that we shouldn't be?

jewelltaylor: Good catch , just updated!

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,149 @@
...
+            "global": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "train_meter_global"),
+            "personal": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "train_meter_personal"),
+        }
+        self.train_metric_meter_mngr = MetricMeterManager(train_key_to_meter_map)
+        val_key_to_meter_map = {
+            "local": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "val_meter_local"),
+            "global": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "val_meter_global"),
+            "personal": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "val_meter_personal"),
+        }
+        self.val_metric_meter_mngr = MetricMeterManager(val_key_to_meter_map)
```
emersodb, line 46: We don't have to do it in this PR, since we're going to follow this up with more changes to fold in the `MetricMeterManager`, but a clean way to do these constructions would be useful 🙂

fl4health/clients/apfl_client.py
```diff
@@ -1,255 +1,149 @@
...
+        self.train_loader = train_loader
+        self.val_loader = val_loader
+
+        # The following lines are type ignored because torch datasets are not "Sized"
+        # IE __len__ is considered optionally defined. In practice, it is almost always defined
+        # and as such, we will make that assumption.
+        self.num_train_samples = len(self.train_loader.dataset)  # type: ignore
+        self.num_val_samples = len(self.val_loader.dataset)  # type: ignore
+
+        optimizer_dict = self.get_optimizer(config)
```
emersodb, line 89: One way that we could get around having to redefine the setup_client for this is to have a `set_optimizers(config)` function which by default does what the BasicClient does, but in APFL is overridden to perform this operation instead. Then setup client could remain un-duplicated (the `assert isinstance(model, ApflModule)` could just be called before the `super().setup_client(config)` call

fl4health/clients/apfl_client.py
```diff
@@ -24,7 +25,36 @@ def __init__(
         metric_meter_type: MetricMeterType = MetricMeterType.AVERAGE,
         checkpointer: Optional[TorchCheckpointer] = None,
     ) -> None:
-        super().__init__(data_path, metrics, device, loss_meter_type, metric_meter_type, checkpointer)
+        super(BasicClient, self).__init__(data_path, device)
```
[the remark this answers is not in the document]
emersodb: Rather than having the user construct MetricMeterManagers, we could have them just pass a set of Meter objects and have the constructor run a function called `setup_metric_meter_managers(meters: Sequence[Meters])`. By default it runs the setup in BasicClient and in APFL, we would just override the function. If you like the idea of constructing MetricMeterManagers though, that would work too.

fl4health/clients/basic_client.py
```diff
@@ -310,18 +324,27 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
         """
-        Return predictions when given input. User can override for more complex logic.
+        Return dict of str and torch.Tensor contaiing predictions when given input.
+        In the default case, the dict has a single item with key prediction.
+        In more complicated approaches such as APFL, the dict has as many items as prediction types
+        User can override for more complex logic.
         """
-        return self.model(input)
+        preds = self.model(input)
+        preds = preds if isinstance(preds, dict) else {"prediction": preds}
```
emersodb, line 335: We might want to do something like 
```
if isinstance(preds, dict):
    return preds
elif isinstance(preds, torch.Tensor):
    return {"prediction": preds}
else:
    raise ValueError("Model forward did not return a tensor or dictionary or tensors")


```

fl4health/clients/basic_client.py
```diff
@@ -310,18 +324,27 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         """
         return FullParameterExchanger()
 
-    def predict(self, input: torch.Tensor) -> torch.Tensor:
+    def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
         """
-        Return predictions when given input. User can override for more complex logic.
+        Return dict of str and torch.Tensor contaiing predictions when given input.
```
emersodb, line 329: containing 🙂

fl4health/utils/metrics.py
```diff
@@ -228,3 +228,29 @@ def compute(self) -> Dict[str, Scalar]:
...
+class MetricMeterManager:
+    """
+    Class to manage one or metric meters.
+    """
+
+    def __init__(self, key_to_meter_map: Dict[str, MetricMeter]):
+        self.key_to_meter_map = key_to_meter_map
+
+    def update(self, preds: Dict[str, torch.Tensor], target: torch.Tensor) -> None:
+        for map_key, pred_key in zip(sorted(self.key_to_meter_map.keys()), sorted(preds.keys())):
```
emersodb, line 242: So we may have a bit of an issue that we've seen before with zip. That is, if the keys are not of equal lengths, zip will still function fine and maybe even pass the assert below if `pred.keys()` is a subset of `key_to_meter_map.keys()`, which will cause some unexpected behavior. So we should also assert that they key sets are the same length

fl4health/clients/apfl_client.py
jewelltaylor: Yeah i will reserve that for the next PR since we are having the user pass MetricMeterManager and they way we construct them is likely to change

research/flamby/fed_isic2019/apfl/client.py
```diff
@@ -1,60 +1,76 @@
...
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        train_dataset, validation_dataset = construct_fedisic_train_val_datasets(
+            self.client_number, str(self.data_path)
+        )
+        train_loader = DataLoader(train_dataset, batch_size=BATCH_SIZE, shuffle=True)
+        val_loader = DataLoader(validation_dataset, batch_size=BATCH_SIZE, shuffle=False)
+        return train_loader, val_loader
 
-        # NOTE: The class weights specified by alpha in this baseline loss are precomputed based on the weights of
-        # the pool dataset. This is a bit of cheating but FLamby does it in their paper.
```
emersodb, line 46: We lost this comment in the refactor. Do you mind adding it back to `get_criterion` It's just a note about the baseline loss for FedIsic

fl4health/utils/metrics.py
jewelltaylor: I was assuming that these would be equal lengths and contain the same keys for this but I will add that edge case

emersodb: Yeah it's just an extra sanity check. We've been burned by the zip on lists of unequal length thing before so I'm paranoid 😂

jewelltaylor: Yeah, I was planning on including more thorough checks in the follow up PR where we have the user define the metermanager because then we have no guarantees about what the keys they pass are and if they map back to what is returned by the predict method. Now we at least have some control because we are defining the manager in the constructor. But I can add some initial checks now but it will probably change a bit next PR

fl4health/clients/apfl_client.py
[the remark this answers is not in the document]
jewelltaylor: Yeah, that is also another option to consider! I will see what ends up being the cleanest and go with that. Will focus on this in the next PR

research/flamby/fed_isic2019/apfl/client.py
jewelltaylor: Updated in [9959c89](https://github.com/VectorInstitute/FL4Health/pull/62/commits/9959c896eba8b13846f86a313be23b34619dc1cb)

fl4health/utils/metrics.py
jewelltaylor: Updated in [9959c89](https://github.com/VectorInstitute/FL4Health/pull/62/commits/9959c896eba8b13846f86a313be23b34619dc1cb)

fl4health/clients/basic_client.py
jewelltaylor: Updated in [9959c89](https://github.com/VectorInstitute/FL4Health/pull/62/commits/9959c896eba8b13846f86a313be23b34619dc1cb)

fl4health/clients/basic_client.py
jewelltaylor: Updated in [9959c89](https://github.com/VectorInstitute/FL4Health/pull/62/commits/9959c896eba8b13846f86a313be23b34619dc1cb)

fl4health/clients/apfl_client.py
jewelltaylor: Good call! Updated in [9959c89](https://github.com/VectorInstitute/FL4Health/pull/62/commits/9959c896eba8b13846f86a313be23b34619dc1cb)

emersodb (approved):
Good to go. Great changes!

PR #63, opened 2023-10-04
Feature alignment

yc7z:
# PR Type
[Feature | Fix | Documentation | Other() ]
Feature

Data preprocessing pipelines for tabular features alignment. 
ClickUp: https://app.clickup.com/t/860q2qc3c

There is also an example under the examples folder which demonstrates the usage of this pipeline.

The ClickUp also includes two new subtasks that I think are the natural next steps following this PR.

# Tests Added
`test_alignment_pipeline.py`


examples/feature_alignment_example/misalign_data.py
```diff
@@ -0,0 +1,40 @@
...
+def random_split_data(df: pd.DataFrame, n: int) -> List[pd.DataFrame]:
+    df_rand = df.sample(frac=1.0, random_state=42)
+    num_rows_per_df = len(df_rand) // n
+    smaller_dfs = [df_rand.iloc[i * num_rows_per_df : (i + 1) * num_rows_per_df] for i in range(n - 1)]
+    smaller_dfs.append(df_rand.iloc[(n - 1) * num_rows_per_df :])
+    return smaller_dfs
+
+
+if __name__ == "__main__":
+    data_path = Path("examples/feature_alignment_example/mimic3d.csv")
```
emersodb, line 17: Since we have the folder `examples/datasets/` can we put the mimic3 dataset csv in there?

examples/feature_alignment_example/misalign_data.py
```diff
@@ -0,0 +1,40 @@
...
+    target_datapath = "examples/feature_alignment_example"
```
emersodb, line 18: If we're going to hardcode this destination, can we also direct it to the datasets folder? It would also be worthwhile to add these files to the gitignore file so that they don't accidentally get checked in.

examples/feature_alignment_example/README.md
```diff
@@ -0,0 +1,43 @@
...
+* `source_specified`: whether the server is provided with the information necessary to perform alignment or not. If this field is `True`, then it means the server already has the information needed for alignment. So the server will send this information to all the clients and wait for them to perform feature alignment. If this field is `False`, then the server will randomly select one of the clients and asks it to provide information about its features, then send this information to all clients so they can use it to perform feature alignment.
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the `n_clients`
+clients expected by the server. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.feature_alignment_example.client --dataset_path <path_to_dataset>
+```
+* `path_to_dataset` is the path towards the directory where the dataset is stored. To see the feature alignment capability in this example, you should launch clients with misaligned datasets.
```
emersodb, line 43: Maybe add an example. That is, launch two clients one pointing to `mimic3d_hospital1.csv` and the other pointing to `mimic3d_hospital2.csv`after running the `misalign_data.py` script.

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.metrics import Metric
+
+
+class TabularDataClient(BasicClient):
+    def __init__(
+        self, data_path: Path, metrics: Sequence[Metric], device: torch.device, id_column: str, target_column: str
+    ) -> None:
+        super().__init__(data_path, metrics, device)
+        self.parameter_exchanger = FullParameterExchanger()
```
emersodb, line 30: We now have a function that you can override called `get_parameter_exchanger`. If you don't override it, by default it will provide a FullParameterExchanger(). So I think we can delete this line.

examples/feature_alignment_example/client.py
```diff
@@ -0,0 +1,98 @@
...
+
+class Mimic3TabularDataClient(TabularDataClient):
+    def __init__(
+        self, data_path: Path, metrics: Sequence[Metric], device: torch.device, id_column: str, target_column: str
+    ) -> None:
+        super().__init__(data_path, metrics, device, id_column, target_column)
+
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        batch_size = self.narrow_config_type(config, "batch_size", int)
+        indices = np.random.permutation(self.aligned_features.shape[0])
```
emersodb, line 30: Maybe a comment here to discuss what you're doing. I _think_ you're permuting the DF columns to ensure the columns don't align.

emersodb, line 30: Maybe a comment here to detail what you're doing with this shuffle and split

examples/feature_alignment_example/client.py
```diff
@@ -0,0 +1,98 @@
...
+        shuffled_targets = self.aligned_targets[indices]
+        split_percentage = 0.9
+        split_point = int(shuffled_data.shape[0] * split_percentage)
+
+        train_data = shuffled_data[:split_point]
+        val_data = shuffled_data[split_point:]
+        train_targets = shuffled_targets[:split_point]
+        val_targets = shuffled_targets[split_point:]
+
+        tensor_train_data = torch.from_numpy(train_data.toarray())
```
emersodb, line 41: I think you can do something like `torch.Tensor(train_data.values())` rather than convert to numpy then from numpy to tensor. I'm not 100% confident on that, so let me know if I'm mistaken.

examples/feature_alignment_example/client.py
```diff
@@ -0,0 +1,98 @@
...
+        tensor_train_targets = torch.from_numpy(train_targets)
+        tensor_val_data = torch.from_numpy(val_data.toarray())
+        tensor_val_targets = torch.from_numpy(val_targets)
+
+        tensor_train_data = tensor_train_data.float()
```
emersodb, line 46: Rather than storing the intermediate `tensor_train_data` and then converting, I might recommend just taking the `.float()` onto the conversion above, since its pretty simple and straightforward what's happening and reduces lines.

examples/feature_alignment_example/client.py
```diff
@@ -0,0 +1,98 @@
...
+        tensor_train_targets = torch.squeeze(tensor_train_targets.long(), dim=1)
+        tensor_val_data = tensor_val_data.float()
+        tensor_val_targets = torch.squeeze(tensor_val_targets.long(), dim=1)
+
+        train_loader = DataLoader(
+            TensorDataset(tensor_train_data, tensor_train_targets), batch_size=batch_size, shuffle=True
+        )
+
+        val_loader = DataLoader(
```
emersodb, line 55: It's not super important, but we can set shuffle to false here (to skip that operation) since it's just a val loader and order doesn't matter.

examples/feature_alignment_example/client.py
```diff
@@ -0,0 +1,98 @@
...
+            TensorDataset(tensor_val_data, tensor_val_targets), batch_size=batch_size, shuffle=True
+        )
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        return LogisticRegression(self.input_dimension, self.output_dimension)
```
emersodb, line 61: In my PR, I'm forcing our BasicClient `setup_client` to move the model to the proper device so we don't have to in the `get_model` function, but as it is setup now, it's a good idea to move the model to `self.device` here. Otherwise you'll run into issues if this example is run on GPU

examples/feature_alignment_example/config.yaml
```diff
@@ -0,0 +1,9 @@
+# Parameters that describe server
+n_server_rounds: 5 # The number of rounds to run FL
+
+# Parameters that describe clients
+n_clients: 2 # The number of clients in the FL experiment
+local_epochs: 5 # The number of epochs to complete for client
+batch_size: 64 # The batch size for client training
+
+source_specified: False
```
emersodb, line 9: Maybe just add a comment as in the other config variables?

examples/feature_alignment_example/misalign_data.py
```diff
@@ -0,0 +1,40 @@
...
+
+    df = pd.read_csv(data_path)
+    df.dropna(inplace=True, how="any")
+    dfs = random_split_data(df, 2)
+    df1 = dfs[0].reset_index()
+    df2 = dfs[1].reset_index()
+
+    df2 = df2.drop(columns=["ExpiredHospital", "admit_type", "NumRx", "ethnicity"])
```
emersodb, line 26: A comment on what's being done here would be useful (Dropping columns to create misalignment. would be sufficient.)

examples/feature_alignment_example/misalign_data.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Now we randomly select 10 percent of the rows of df2
+    # and set its 'insurance' column to 'Unknown'
+    num_rows_to_select = int(0.10 * len(df2))
+    random_indices = random.sample(range(len(df2)), num_rows_to_select)
+
+    # Set a specific column ('B' in this example) to the specific value for the selected rows
```
emersodb, line 33: Should "B" here actually be "insurance"?

examples/feature_alignment_example/misalign_data.py
```diff
@@ -0,0 +1,40 @@
...
+    df2.loc[random_indices, "insurance"] = "Unknown"
+
+    print(f"hospital1 insurance values: {df1['insurance'].unique()}")
```
emersodb, line 36: If you think it's useful to have this in the script, let's use `log(INFO, f"...")` instead of print

examples/feature_alignment_example/server.py
```diff
@@ -0,0 +1,98 @@
...
+from flwr.common.typing import Metrics, Parameters
+
+from examples.models.logistic_regression import LogisticRegression
+from examples.simple_metric_aggregation import metric_aggregation, normalize_metrics
+from fl4health.client_managers.poisson_sampling_manager import PoissonSamplingClientManager
+from fl4health.feature_alignment.tab_features_info_encoder import TabularFeaturesInfoEncoder
+from fl4health.server.tabular_feature_alignment_server import TabularFeatureAlignmentServer
+from fl4health.strategies.basic_fedavg import BasicFedAvg
+from fl4health.utils.config import load_config
+
```
emersodb, line 17: A comment here might be good saying that this data path is used to create a "source of truth" on the server-side as an example. This is used if the config specifies `source_specified` as true

examples/feature_alignment_example/server.py
```diff
@@ -0,0 +1,98 @@
...
+DATA_PATH = "examples/feature_alignment_example/mimic3d_hospital1.csv"
+CONFIG_PATH = "examples/feature_alignment_example/config.yaml"
+
+
+def get_initial_model_parameters(input_dim: int, output_dim: int) -> Parameters:
+    initial_model = LogisticRegression(input_dim, output_dim)
+    return ndarrays_to_parameters([val.cpu().numpy() for _, val in initial_model.state_dict().items()])
+
+
+def fit_metrics_aggregation_fn(all_client_metrics: List[Tuple[int, Metrics]]) -> Metrics:
```
emersodb, line 27: Both of these are now reusable functions in `examples/simple_metric_aggregation.py` unless these functions differ from the ones implemented there. You can just import them from that file.

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+        self.initialize_parameters = initialize_parameters
+        self.format_info_gathered = False
+        # casting self.strategy to BasicFedAvg so its on_fit_config_fn can be specified.
+        self.strategy: BasicFedAvg
+        self.strategy.on_fit_config_fn = partial(fit_config, self.config, self.format_info_gathered)
+
+    def fit(self, num_rounds: int, timeout: Optional[float]) -> History:
+        """Run federated averaging for a number of rounds."""
+
+        assert isinstance(self.strategy, BasicFedAvg)
```
emersodb, line 69: I would recommend we test whether the `initial_parameters` of the strategy have been set here. If they have, we should warn the user that they will be ignored in favor of the `self.initialize_parameters` function

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+        config: Config,
+        initialize_parameters: Callable,
+        strategy: BasicFedAvg,
+        wandb_reporter: Optional[ServerWandBReporter] = None,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        tab_features_info: Optional[TabularFeaturesInfoEncoder] = None,
+    ) -> None:
+        assert isinstance(strategy, BasicFedAvg)
+        if strategy.on_fit_config_fn is not None:
+            warnings.warn("strategy.on_fit_config_fn will be overwritten.")
```
emersodb, line 49: I think these are valid warning logs, but we've been using the flwr logger from 
```
from flwr.common.logger import log
from logging import INFO, WARNING
```
as `log(WARNING, f"....")` for warnings in other places. So I think we should stick with that, unless you can think of a reason we shouldn't be?

examples/feature_alignment_example/server.py
```diff
@@ -0,0 +1,98 @@
...
+        # Server waits for min_available_clients before starting FL rounds
+        min_available_clients=config["n_clients"],
+        on_fit_config_fn=None,
+        on_evaluate_config_fn=None,
+        fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+        evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        initial_parameters=None,
+    )
+
+    source_specified = config["source_specified"]
```
emersodb, line 62: I'm a little surprised this isn't throwing a mypy arrow without type narrowing. I assume you've tested that the boolean value works here?

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def __init__(
+        self,
+        client_manager: ClientManager,
+        config: Config,
+        initialize_parameters: Callable,
+        strategy: BasicFedAvg,
+        wandb_reporter: Optional[ServerWandBReporter] = None,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        tab_features_info: Optional[TabularFeaturesInfoEncoder] = None,
```
emersodb, line 45: Maybe we can name this something slightly more informative like `tabular_features_source_of_truth`?

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+        and initializing the corresponding TabularFeaturesPreprocessor.
+
+        config[FORMAT_SPECIFIED] indicates whether the server has obtained
+        the source of information to perform feature alignment.
+        If it is True, it means the server has obtained such information
+        (either a priori or by polling a client).
+        So the client will encode that information and use it instead
+        to perform feature preprocessing.
+        """
+        assert FORMAT_SPECIFIED in config.keys()
```
emersodb, line 54: `narrow_config_type` will perform a similar assertion, so you don't need to do it here 🙂

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+        format_specified = self.narrow_config_type(config, FORMAT_SPECIFIED, bool)
+
+        # Encode the information of the client's local tabular data.
+        self.df = self.get_data_frame(config)
+        self.tabular_features_info_encoder = TabularFeaturesInfoEncoder.encoder_from_dataframe(
```
emersodb, line 59: If I'm reading it correctly, this process need not occur if `format_specified` is false. So I'd put this instantiation in an else block as the counter to the `if format_specified:` block to save the processing.

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+        Depending on whether the server has communicated the information
+        to be used for feature alignment, the client will send the input/output
+        dimensions so the server can use them to initialize the global model.
+
+        First initializes the client because this is called prior to the first
+        federated learning round.
+        """
+        self.setup_client(config)
+        format_specified = self.narrow_config_type(config, FORMAT_SPECIFIED, bool)
+        if not format_specified:
```
emersodb, line 106: This could be simplified slightly to
```
properties: Dict[str, Scalar] = {FEATURE_INFO: self.tabular_features_info_encoder.to_json()}
if format_specified:
    properties.update({INPUT_DIMENSION: self.input_dimension, OUTPUT_DIMENSION: self.output_dimension}
return properties
```

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+            return {
+                FEATURE_INFO: self.tabular_features_info_encoder.to_json(),
+            }
+        else:
+            return {
+                FEATURE_INFO: self.tabular_features_info_encoder.to_json(),
```
emersodb, line 112: I may get the answer later in the review, but if we already have `format_specified`, do we need the `tabular_features_info_encoder` to be sent back to the server?

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+        if format_specified:
+            # Since the server has obtained its source of information,
+            # the client will encode that instead.
+            self.tabular_features_info_encoder = TabularFeaturesInfoEncoder.from_json(
+                self.narrow_config_type(config, FEATURE_INFO, str)
+            )
+            self.tabular_features_preprocessor = TabularFeaturesPreprocessor(self.tabular_features_info_encoder)
+
+            # preprocess features.
+            self.aligned_features, self.aligned_targets = self.tabular_features_preprocessor.preprocess_features(
```
emersodb, line 72: If the `aligned_features` and `aligned_targets` are fairly large, we probably don't want to keep them around in the clients as these simply serve as intermediate objects to create the dataloaders. I would recommend that, after the 
```
super().setup_client(config)
```
below, we have something like
```
self.aligned_features = None
self.aligned_targets = None
```
or something like it.

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+                self.df
+            )
+
+            # Obtain the input and output dimensions to be sent to
+            # the server for global model initialization.
+            self.input_dimension = self.aligned_features.shape[1]
+            target_type = self.tabular_features_info_encoder.get_target_type()
+            if target_type == ORDINAL or target_type == BINARY:
+                self.output_dimension = len(self.tabular_features_info_encoder.get_target_categories())
```
emersodb, line 81: Rather than performing the `len(...)` here, maybe we can bury this as another function of `tabular_features_info_encoder` as `get_target_dimension()`

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        Return properties of client to be sent to the server.
+        Depending on whether the server has communicated the information
+        to be used for feature alignment, the client will send the input/output
+        dimensions so the server can use them to initialize the global model.
+
+        First initializes the client because this is called prior to the first
+        federated learning round.
+        """
+        self.setup_client(config)
```
emersodb, line 104: As a check, I still think we still want to do 
```
        if not self.initialized:
            self.setup_client(config)
```

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+        if not self.initial_polls_complete:
+
+            # If the server does not have the needed feature info a priori,
+            # then it requests such information from the clients before the
+            # very first fitting round.
+            if self.tab_features_info is None:
+                # A random client's feature information is selected as the source of truth for feature alignment.
+                feature_info = self.poll_clients_for_feature_info(timeout)
+
+                rand_idx = random.randint(0, len(feature_info) - 1)
```
emersodb, line 83: Rather than generating a random index, I believe you can use something like `random.choose(list, 1)` or close to it to just select an element uniformly from a list.

emersodb, line 83: Rather than generating a random index, I believe you can use something like `random.choose(list, 1)` or close to it to just select an element uniformly from a list.

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+            self.initial_polls_complete = True
+
+        # Normal federated learning rounds commence after all clients' features
+        # are aligned and global model is initialized.
+        return super().fit(num_rounds=num_rounds, timeout=timeout)
+
+    def poll_clients_for_feature_info(self, timeout: Optional[float]) -> List[str]:
+        log(INFO, "Feature information source unspecified. Polling clients for feature information.")
+        assert isinstance(self.strategy, BasicFedAvg)
+        client_instructions = self.strategy.configure_poll(server_round=1, client_manager=self._client_manager)
```
emersodb, line 115: This is a bit of a hack, but configure_poll selects all available clients. We actually don't want all clients, we want to randomly select a single on to get feature info from. So I would suggest that rather than having all clients send their feature information and we just select one, we randomly select one from the `client_instructions` using `random.choice`. This will then poll that randomly selected client for features, results can be asserted to be of length 1 and feature_info will just be a string rather than a list.

fl4health/server/tabular_feature_alignment_server.py
emersodb: See comment below about not getting features from all clients and then sampling.

fl4health/feature_alignment/constants.py
```diff
@@ -0,0 +1,39 @@
+from typing import Dict, Union
+
+from flwr.common.typing import Scalar
+from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer, TfidfTransformer, TfidfVectorizer
+
+TextFeatureTransformer = Union[CountVectorizer, TfidfTransformer, TfidfVectorizer, HashingVectorizer]
+
+# Constants that specify the data type of tabular data columns.
```
emersodb, line 8: So rather than having these be string constants and feature types be a list, I think we want to make this an enum with possible values as below. There are ways to iterate through the possible values of an enum,  so we wouldn't need the feature_types list anymore. That way we don't have to keep these in sync if we ever update anything and the types are strongly typed by the enum.

fl4health/feature_alignment/string_columns_transformer.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    The purpose of this class is to enable the application of text feature transformers from sklearn
+    to multiple string columns, which is not supported in the first place.
+    """
+
+    def __init__(self, transformer: TextFeatureTransformer):
+        self.transformer = transformer
+
+    def fit(self, X: pd.DataFrame, y: pd.DataFrame = None) -> "StringColumnTransformer":
+        joined_X = X.apply(lambda x: " ".join(x), axis=1)
```
emersodb, line 17: If, for example, we're using a TFIDF vectorizer, doesn't this mean that the TFIDF vectors will be informed by multiple columns and represent a single text vector for all text columns concatenated together? Would it not make more sense to specify a list of transformers (one per column) and fit them to individual columns, transform those columsn and concatenate them together to produce the transformed dataframe?

emersodb: I don't mind this idea as a preprocessing option, but I'm not sure we want it to be the default. Maybe if the list of transformers is only a single transformer but receives multiple columns this can be the default behavior?

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+    ----------
+    target: str
+        Target column.
+    target_type: str
+        Target column data type. Must be one of BINARY, NUMERICAL, ORDINAL, and STRING.
+    target_categories: List[Scalar]
+        The categories of the target column, provided it is of type ORDINAL.
+    """
+
+    def __init__(self, target: str, target_type: str, target_categories: List[Scalar]) -> None:
```
emersodb, line 28: I'm a little confused what the `target_categories` are, for example, if the datatype is STRING? Is that not allowed? If so, we should throw an error I think?

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+    that is necessary to perform feature alignment.
+
+    Parameters
+    ----------
+    target: str
+        Target column.
+    target_type: str
+        Target column data type. Must be one of BINARY, NUMERICAL, ORDINAL, and STRING.
+    target_categories: List[Scalar]
+        The categories of the target column, provided it is of type ORDINAL.
```
emersodb, line 25: This is also defined in the case of BINARY correct?

fl4health/feature_alignment/tab_features_info_encoder.py
emersodb: We should also have an assert somewhere that this list is empty for a NUMERICAL target and possible a STRING target if that's allowed

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+        self.target_type = target_type
+        self.target_categories = target_categories
+
+    def get_target(self) -> str:
+        return self.target
+
+    def get_target_type(self) -> str:
+        return self.target_type
+
+    def get_target_categories(self) -> List[Scalar]:
```
emersodb, line 39: Since `get_target`, `get_target_type`, and `get_target_categories` are just accessor functions with no logic, I think we can just access these values directly instead of calling these functions?

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+    features_to_types: Dict[str, str]
+        Dictionary that maps each feature name to its type.
+        We consider four types in tabular data:
+        BINARY, ORDINAL, NUMERICAL, and STRING.
+    categories: Dict[str, List[Scalar]]
+        Dictionary that maps each ordinal feature to its categories.
+    target_info: TargetInfoEncoder
+        Information about the target column(s).
+    vocabulary: Dict[str, int]
+        Vocabulary of all the STRING columns.
```
emersodb, line 77: I think, by default, we want to have a vocabulary per STRING column rather than one for all columns.

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+
+        if target_type == ORDINAL or target_type == BINARY:
+            target_categories = sorted(df[target_column].unique().tolist())
+        else:
+            target_categories = []
+
+        target_info = TargetInfoEncoder(target_column, target_type, target_categories)
+
+        # Extract vocabulary from the string columns of df
+        count_vectorizer = CountVectorizer()
```
emersodb, line 167: So it looks like we're restricting ourself to CountVectorizer() transforms for strings here. I think we can be a bit more dynamic here by exchanging information through bytes with pickle.dumps(). The poll_clients function can return a dictionary of type Dict[str, Scalar]. So bytes are an acceptable value we can pack in there. Thus, anything that is, for example, an sklearn object can be pickle.dumps(blah) and packed into those result to be materialized on the server side.

fl4health/feature_alignment/constants.py
```diff
@@ -0,0 +1,39 @@
...
+
+FEATURE_TYPES = [
+    NUMERIC,
+    BINARY,
+    STRING,
+    ORDINAL,
+]
+
+# The default fill-in values for entire missing columns of a certain data type.
+DEFAULT_FILL_VALUES: Dict[str, Scalar] = {BINARY: 0, NUMERIC: 0.0, ORDINAL: UNKNOWN, STRING: "N/A"}
```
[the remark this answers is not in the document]
emersodb: Below, I see why you defined a DEFAULT_FILL_VALUES dictionary. I think that's fine to do actually, rather than defining the static_method above. It can, however, be a Dict[TabularType, Scalar] instead of just a str->scalar dict.

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @staticmethod
+    def encoder_from_dataframe(
+        df: pd.DataFrame,
+        id_column: str,
+        target_column: str,
+        default_fill_values: Dict[str, Scalar] = DEFAULT_FILL_VALUES,
+    ) -> "TabularFeaturesInfoEncoder":
+        features_list = sorted(df.columns.values.tolist())
+        features_list.remove(id_column)
```
emersodb, line 140: This requires that there be an `id_column`, but what if there isn't? Maybe we make it optional?

emersodb: A similar question for `TabularFeatures` also exists. I'm not super familiar with that class in CyclOps

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+        features_to_types = tab_features.types
+        target_type = features_to_types[target_column]
+
+        # The target column is separated from the other columns
+        # so that targets and features remain separate after alignment
+        features_to_types.pop(target_column)
+
+        # extract categories information
+        ordinal_features: List[str] = sorted(tab_features.features_by_type(ORDINAL))
```
emersodb, line 152: super minor, but below we specify `type_ =` but we don't here. Just want to make sure it's not required.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,157 @@
...
+    tab_feature_encoder: TabularFeaturesInfoEncoder
+        Encodes the information necessary for constructing the column transformers.
+    """
+
+    def __init__(self, tab_feature_encoder: TabularFeaturesInfoEncoder) -> None:
+        self.transformers: Dict[str, Pipeline] = {}
+        self.initialize_default_data_transformer(tab_feature_encoder)
+        self.initialize_default_target_transformer(tab_feature_encoder)
+
+    def initialize_default_target_transformer(self, tab_feature_encoder: TabularFeaturesInfoEncoder) -> None:
```
emersodb, line 38: Rather than calling these initialize_default_target_transformer we should have something like
```
def initialize_target_transformer(self, tab_feature_encoder: TabularFeaturesInfoEncoder, column_transformer: Optional[ColumnTransformer] = None) -> None:
if column_transformer:
    self.target_column_transformer
else:
    logic for defaults based on target type
```

emersodb: I would also say we should encapsulate these different column transformer creation defaults into separate methods to make this easier to read

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,157 @@
...
+            transformers=[
+                ("num", self.transformers[NUMERIC], self.type_to_features[NUMERIC]),
+                ("bin", self.transformers[BINARY], self.type_to_features[BINARY]),
+                ("str_vectorizer", self.transformers[STRING], self.type_to_features[STRING]),
+                ("cat", self.transformers[ORDINAL], self.type_to_features[ORDINAL]),
+            ],
+            remainder="drop",
+        )
+
+    def set_data_transformer(self, data_type: str, transformer: Pipeline) -> None:
```
emersodb, line 111: It doesn't seem like either `set_data_transformer` or `set_data_transformer` are used anywhere, but I might be mistaken

fl4health/feature_alignment/tab_features_preprocessor.py
emersodb: Something similar could be down for the data_transformer

pyproject.toml
```diff
@@ -24,6 +24,7 @@ torchvision = "^0.14.1"
 torchtext = "0.14.1"
 torchdata = "0.5.1"
 torcheval = "^0.0.6"
+ipykernel = "^6.25.1"
```
emersodb, line 27: We use a little bit of cyclops in this PR, should that be added here along with into the requirements.txt?

examples/feature_alignment_example/client.py
yc7z: I think as long as `train_data` contains string columns, they will be transformed into a sparse matrix, so it seems that we need to convert it to a numpy array first.

emersodb: Ah, gotcha. That makes sense!

fl4health/server/tabular_feature_alignment_server.py
yc7z: No, I think it makes sense to stick with the flwr logger.

examples/feature_alignment_example/server.py
yc7z: I see what you mean, but yes this passed the mypy type checker somehow... I'll still add the type narrowing

fl4health/clients/tabular_data_client.py
yc7z: We don't really need it as it currently stands. So I'll remove the feature info from the else block for now.

fl4health/clients/tabular_data_client.py
yc7z: Yes, good call

fl4health/clients/tabular_data_client.py
yc7z: You are right that feature info doesn't need to be sent to the server if `format_specified` is True, so I just removed it from the dictionary returned here

fl4health/server/tabular_feature_alignment_server.py
yc7z: Yes I think this a good idea! Originally I polled all of the clients thinking we might want to select multiple ones and merge their information, but even in that case we can just use `random.choice` on `client_instructions`.

yc7z: I think this also applies to the `poll_clients_for_dimension_info` method, so I'll make these changes there as well.

fl4health/feature_alignment/constants.py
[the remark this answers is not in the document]
yc7z: I think this is a really good idea, I'll change the code to using enum instead.

fl4health/feature_alignment/tab_features_preprocessor.py
fatemetkl, line 111: Maybe we can extend this function to add transformations for specific column names beyond the data_type restriction. So we would be able to apply any customized pipeline on any of the feature columns.

fl4health/feature_alignment/tabular_type.py
yc7z, line 1: We now use Enum to represents the data type of tabular features as opposed to hard-coded constants.

fl4health/feature_alignment/tab_features_info_encoder.py
yc7z: This is still a relevant comment after the new design. CyclOps requires there to be an id column, and I don't think it's unreasonable that we require there to be one as well, since we are working with pandas DataFrame here and the user could very easily add an id column if needed.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,157 @@
...
+        and self.target_column_transformer to the target column to achieve
+        feature alignment.
+        """
+        # If the dataframe has an entire column missing, we need to fill it with some default value first.
+        df_filled = self.fill_in_missing_columns(df)
+        # After filling in missing columns, apply the feature alignment transform.
+        return self.data_column_transformer.fit_transform(
+            df_filled.drop(columns=[self.target_column])
+        ), self.target_column_transformer.fit_transform(df_filled[[self.target_column]])
+
```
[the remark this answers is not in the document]
yc7z: Yes, this is fine because the columns that do not have corresponding transformers would simply be dropped.

fl4health/feature_alignment/tab_features_info_encoder.py
emersodb: I think that makes sense to me. I suppose the table index could always be used.

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+    tab_features_info: the information that is required for aligning client features.
+    If it is not specified, then the server will randomly poll a client and gather
+    this information from its data source.
+    """
+
+    def __init__(
+        self,
+        client_manager: ClientManager,
+        config: Config,
+        initialize_parameters: Callable,
```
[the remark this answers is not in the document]
emersodb: I think this comment still applies after the refactor.

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,142 @@
...
+        # The server performs one or two rounds of polls before the normal federated training.
+        # The first one gathers feature information if the server does not already have it,
+        # and the second one gathers the input/output dimensions of the model.
+        self.initial_polls_complete = False
+        self.tab_features_info = tabular_features_source_of_truth
+        self.config = config
+        self.initialize_parameters = initialize_parameters
+        self.format_info_gathered = False
+        # casting self.strategy to BasicFedAvg so its on_fit_config_fn can be specified.
+        self.strategy: BasicFedAvg
```
emersodb, line 62: Rather than casting in this way, I think it's very slightly better to perform an assertion here.

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,105 @@
...
+            # the server for global model initialization.
+            self.input_dimension = self.aligned_features.shape[1]
+            self.output_dimension = self.tabular_features_info_encoder.get_target_dimension()
+            log(INFO, f"input dimension: {self.input_dimension}, output_dimension: {self.output_dimension}")
+
+            super().setup_client(config)
+
+            # freeing the memory of aligned features/targets.
+            del self.aligned_features
+            del self.aligned_targets
```
emersodb, line 71: I'd say it's worth freeing the memory associated with self.df here too?

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+            self.tabular_features_preprocessor = TabularFeaturesPreprocessor(self.tabular_features_info_encoder)
+
+            # preprocess features.
+            self.aligned_features, self.aligned_targets = self.tabular_features_preprocessor.preprocess_features(
+                self.df
+            )
+
+            # Obtain the input and output dimensions to be sent to
+            # the server for global model initialization.
```
emersodb, line 62: We're making an assumption here that the first dimension is the number of rows. Very reasonable, but maybe worth adding to the comment.

emersodb: Super minor but is it worth having a `get_feature_dimension` function in the `tabular_features_info_encoder` class that abstracts this away?

fl4health/clients/tabular_data_client.py
```diff
@@ -0,0 +1,105 @@
...
+            self.output_dimension = self.tabular_features_info_encoder.get_target_dimension()
+            log(INFO, f"input dimension: {self.input_dimension}, output_dimension: {self.output_dimension}")
+
+            super().setup_client(config)
+
+            # freeing the memory of aligned features/targets.
+            del self.aligned_features
+            del self.aligned_targets
+        else:
+            # Encode the information of the client's local tabular data.
```
emersodb, line 73: Probably worth commenting that this is expected to happen only once and only if the client is requested to provide alignment information?

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,143 @@
...
+            self.format_info_gathered = True
+
+            self.strategy.on_fit_config_fn = partial(fit_config, self.config, self.format_info_gathered)
+
+            # Now the server waits until feature alignment is performed on the clients' side
+            # and subsequently requests the input and output dimensions, which are needed for initializing
+            # the global model.
+            input_dimension, output_dimension = self.poll_clients_for_dimension_info(timeout)
+            log(INFO, f"input dimension: {input_dimension}, output dimension: {output_dimension}")
+            self.strategy.initial_parameters = self.initialize_parameters(input_dimension, output_dimension)
```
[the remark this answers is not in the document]
emersodb: I think this suggestion still applies. That way we don't have to overwrite and rely on the `initial_parameters` of the strategy

fl4health/feature_alignment/string_columns_transformer.py
```diff
@@ -0,0 +1,40 @@
...
+class TextColumnTransformer(BaseEstimator, TransformerMixin):
+    """
+    The purpose of this class is to enable the application of text feature transformers from sklearn
+    to a single-column pandas dataframe, which is not supported in the first place.
+    """
+
+    def __init__(self, transformer: TextFeatureTransformer):
+        self.transformer = transformer
+
+    def fit(self, X: pd.DataFrame, y: pd.DataFrame = None) -> "StringColumnTransformer":
```
emersodb, line 35: I think the return type for this should be `TextColumnTransformer`?

fl4health/feature_alignment/string_columns_transformer.py
```diff
@@ -0,0 +1,40 @@
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from fl4health.feature_alignment.constants import TextFeatureTransformer
+
+
+class StringColumnTransformer(BaseEstimator, TransformerMixin):
```
emersodb, line 7: This class doesn't seem to be used anywhere. If you think it's still useful, I'm okay with keeping it, but maybe we name it in a more detailed way. Something like `TextMulticolumnTransformer` maybe?

fl4health/feature_alignment/string_columns_transformer.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    The purpose of this class is to enable the application of text feature transformers from sklearn
+    to a single-column pandas dataframe, which is not supported in the first place.
+    """
+
+    def __init__(self, transformer: TextFeatureTransformer):
+        self.transformer = transformer
+
+    def fit(self, X: pd.DataFrame, y: pd.DataFrame = None) -> "StringColumnTransformer":
+        self.transformer.fit(X[X.columns[0]])
```
emersodb, line 36: If X is meant to be a DF with a single column, we should probably assert that this is true in both the fit and transform functions.

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,130 @@
...
+        self.tabular_targets = sorted(tabular_targets, key=TabularFeature.get_feature_name)
+
+    def get_tabular_features(self) -> List[TabularFeature]:
+        return self.tabular_features
+
+    def get_tabular_targets(self) -> List[TabularFeature]:
+        return self.tabular_targets
+
+    def get_feature_columns(self) -> List[str]:
+        return sorted([feature.get_feature_name() for feature in self.tabular_features])
```
emersodb, line 37: I'm probably missing something, but is there any reason why this list needs to be sorted?

emersodb: A similar question might apply to `get_target_columns`

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def get_target_dimension(self) -> int:
+        # Return the dimension of the target array *after* feature alignment is performed.
+        dimension = 0
+        for target in self.tabular_targets:
+            dimension += target.get_metadata_dimension()
+        return dimension
+
+    @staticmethod
+    def _construct_tab_feature(
```
emersodb, line 62: I think, but I'm not 100% sure (so let me know if I'm wrong), but I think the _ prefix in python is typically reserved for "private/protected" object methods (i.e. internal calls). However, I may be wrong and that this is also conventional for static calls.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+    then be applied to a pandas dataframe.
+
+    Each tabular feature, which corresponds to a column
+    in the pandas dataframe, has its own column transformer. A default
+    transformer is initialized for each feature based on its data type,
+    but the user may also manually specify a transformer for this
+    feature.
+
+    Parameters
+    ----------
```
emersodb, line 31: It looks like these comments, and a few others throughout are in the "numpy" docstring format. Can you convert them over to the "Google" format? Otherwise they are really great!

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+        self.data_column_transformer = self.return_column_transformer(self.features_to_pipelines)
+        self.target_column_transformer = self.return_column_transformer(self.targets_to_pipelines)
+
+    def get_default_numeric_pipeline(self) -> Pipeline:
+        return Pipeline(steps=[("imputer", SimpleImputer(strategy="mean")), ("scaler", MinMaxScaler())])
+
+    def get_default_binary_pipeline(self) -> Pipeline:
+        return Pipeline(steps=[("imputer", SimpleImputer(strategy="most_frequent")), ("encoder", OrdinalEncoder())])
+
+    def get_one_hot_pipeline(self, categories: MetaData) -> Pipeline:
```
emersodb, line 58: Just a suggestion, but for this function, and `get_ordinal_pipeline` below, maybe we follow the same naming convention as above and prefix them as `get_default_...` to be consistent?

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+                        categories=[categories],
+                    ),
+                )
+            ]
+        )
+
+    def get_default_string_pipeline(self, vocabulary: MetaData) -> Pipeline:
+        return Pipeline(steps=[("vectorizer", TextColumnTransformer(TfidfVectorizer(vocabulary=vocabulary)))])
+
+    def initialize_default_pipelines(
```
emersodb, line 78: Since this function is an "initialization function, do we need the `feature_to_transformer` dict as an argument. I think we could just fill an empty on here and return it. Then the call would just initialize the pipelines as
```
self.features_to_pipelines: Dict[str, Pipeline] = self.initialize_default_pipelines(self.tabular_features, self.features_to_pipelines, one_hot=True)
self.targets_to_pipelines: Dict[str, Pipeline] = self.initialize_default_pipelines(self.tabular_targets, self.targets_to_pipelines, one_hot=False)
```
in the constructor.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+            feature_to_transformer[feature_name] = feature_pipeline
+
+    def return_column_transformer(self, pipelines: Dict[str, Pipeline]) -> ColumnTransformer:
+        transformers = [
+            (f"{feature_name}_pipeline", pipelines[feature_name], [feature_name])
+            for feature_name in sorted(pipelines.keys())
+        ]
+        return ColumnTransformer(
+            transformers=transformers,
+            remainder="drop",
```
emersodb, line 114: Does this mean that if a column does not have an associated transformer then it is dropped from the df? That's okay, but we should probably mention it in a comment documentation for the `TabularFeaturesPreprocessor` class.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    def set_feature_pipeline(self, feature_name: str, pipeline: Pipeline, target: bool) -> None:
```
emersodb, line 117: In the example that you implemented, maybe we can show how this function is used somewhere? That is, we set a custom pipeline for the feature. It can be the same as the default, but demonstrating the use would be nice.

fl4health/feature_alignment/tab_features_preprocessor.py
```diff
@@ -0,0 +1,147 @@
...
+        # This method allows the user to customize a specific pipeline to be applied to a specific feature.
+        # For example, the user may want to use different scalers for two distinct numerical features.
+        if not target:
```
emersodb, line 120: Just a thought and you don't have to use it, but rather than have the user specify whether the feature is a target or not, could we simply just do
```
if feature_name in self.features_to_pipelines:
    self.features_to_pipelines[feature_name] = pipeline
    self.data_column_transformer = self.return_column_transformer(self.features_to_pipelines)
elif feature_name in self.targets_to_pipelines:
    self.targets_to_pipelines[feature_name] = pipeline
    self.target_column_transformer = self.return_column_transformer(self.targets_to_pipelines)
else:
    log(WARNING, f"{feature_name} is not in either the feature or target pipelines and the provided pipeline will be ignored."
```

fl4health/feature_alignment/tabular_feature.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_metadata(self) -> MetaData:
+        return self.metadata
+
+    def get_metadata_dimension(self) -> int:
+        if self.feature_type == TabularType.BINARY or TabularType.ORDINAL:
+            return len(self.metadata)
+        elif self.feature_type == TabularType.NUMERIC:
+            return 1
+        else:
```
emersodb, line 52: Rather than returning zero here, does it make sense to just raise an exception, as getting the dimension for other feature types isn't "supported"? Maybe there are downstream implications though, so correct me if I'm wrong.

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,142 @@
...
+        self,
+        client_manager: ClientManager,
+        config: Config,
+        initialize_parameters: Callable,
+        strategy: BasicFedAvg,
+        wandb_reporter: Optional[ServerWandBReporter] = None,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        tabular_features_source_of_truth: Optional[TabularFeaturesInfoEncoder] = None,
+    ) -> None:
+        assert isinstance(strategy, BasicFedAvg)
```
emersodb, line 46: Since we're explicitly typing this argument in the `__init__` signature, do we still need this assert?

pyproject.toml
emersodb: I think this comment still applies too?

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,142 @@
...
+        feature_info = str(get_properties_res.properties[FEATURE_INFO])
+        return feature_info
+
+    def poll_clients_for_dimension_info(self, timeout: Optional[float]) -> Tuple[int, int]:
+        log(INFO, "Waiting for Clients to align features and then polling for dimension information.")
+        assert isinstance(self.strategy, BasicFedAvg)
+        client_instructions = self.strategy.configure_poll(server_round=1, client_manager=self._client_manager)
+        # Since the features of all clients are aligned, we just select one client
+        # to obtain the input/output dimensions.
+
```
emersodb, line 128: Super minor, but the space between the comment above and this call to poll clients should be removed since the comment applies to the call I think?

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -0,0 +1,142 @@
...
+        results, _ = poll_clients(
+            client_instructions=client_instructions[:1], max_workers=self.max_workers, timeout=timeout
+        )
+        assert len(results) == 1
+        input_dimension = int(results[0][1].properties[INPUT_DIMENSION])
+        out_put_dimension = int(results[0][1].properties[OUTPUT_DIMENSION])
```
emersodb, line 134: I think we want this to be called `output_dimension` 🙂

emersodb (commented):
I think this is looking really good and I think you've improved the code readability a lot! Most of the comments that I left are pretty minor. One thing that we don't have any of is tests. I think it would be good for us to at least mock up some very small misaligned dataframes to make sure all of the transforms etc. work as expected.

fl4health/feature_alignment/tab_features_info_encoder.py
yc7z: Yes you are right, it is typically reserved for internal calls. I used it here because I thought of `_construct_tab_feature` as a helper method that is only used internally.

fl4health/feature_alignment/tab_features_preprocessor.py
yc7z: Yes good idea!

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -65,138 +17,114 @@ class TabularFeaturesInfoEncoder:
...
+        fill_values: Optional[Dict[str, Scalar]],
+    ) -> TabularFeature:
+        feature_type = TabularType(features_to_types[feature_name])
+
+        if fill_values is None or feature_name not in fill_values:
+            fill_value = TabularType.get_default_fill_value(feature_type)
+        else:
+            fill_value = fill_values[feature_name]
+
+        if feature_type == TabularType.ORDINAL or feature_type == TabularType.BINARY:
```
fatemetkl, line 73: Are we extracting categories for binary type as well?

fatemetkl (commented):
The structure of the code and its readability have improved considerably! 
Just had two minor comments, testing would also be a good addition, other than that the code looks great to me.

fl4health/feature_alignment/tab_features_info_encoder.py
yc7z: I guess it doesn't have to be sorted strictly speaking... It was a remanent from the previous iteration of the design. I kept it here because I thought it would make testing easier.

fl4health/feature_alignment/tab_features_info_encoder.py
```diff
@@ -65,138 +17,114 @@ class TabularFeaturesInfoEncoder:
...
-        string_col_transformer.fit(df[string_features])
-        vocabulary = count_vectorizer.vocabulary_
 
-        return TabularFeaturesInfoEncoder(features_to_types, categories, target_info, vocabulary, default_fill_values)
+        tabular_targets = []
+        tabular_features = []
+        # Construct TabularFeature objects.
+        for feature_name in features_to_types:
+            tabular_feature = TabularFeaturesInfoEncoder._construct_tab_feature(
+                df, feature_name, features_to_types, fill_values
```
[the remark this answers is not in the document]
yc7z: Good point!

fl4health/clients/tabular_data_client.py
yc7z: I looked into this and it might be a little tricky because the feature dimension is only known after the pipeline is applied to the dataframe... There might be a way to "predict" what the dimension would be based on the type of column transformer applied, but I'm not sure how to do that...

yc7z:
> I think this is looking really good and I think you've improved the code readability a lot! Most of the comments that I left are pretty minor. One thing that we don't have any of is tests. I think it would be good for us to at least mock up some very small misaligned dataframes to make sure all of the transforms etc. work as expected.

I've now added some tests to test the basic functionality on some small dataframes.

fl4health/feature_alignment/tab_features_info_encoder.py
yc7z: Yes. This is inherited from the previous iteration of the code because I thought it might be useful if the user wishes to impute missing value with a special unknown value, in which case the binary type would essentially become a categorical type. Do you think it would make sense to keep this?

fl4health/feature_alignment/tabular_feature.py
yc7z: It's true we are assuming that the target feature cannot be of type string, but since this method is for the class `TabularFeature` I think it's kind of odd that it would raise an error only for certain instances of the class but not others. Originally I thought to return the length of the vocabulary for string features here, but because string pipelines produce sparse matrices, such a length would not reflect the actual "dimension" of the output. So I just put zero here instead... I don't think there is any downstream implication though, so let me know what you think is the best.

fl4health/server/tabular_feature_alignment_server.py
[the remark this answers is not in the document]
yc7z: The changes have been applied now

fl4health/clients/tabular_data_client.py
emersodb: Gotcha. That's good with me. Just wanted to see if it were doable.

fl4health/feature_alignment/tabular_feature.py
emersodb: I would say that it's okay for us to raise an exception for only some `feature_types` here if using `get_metadata_dimension` on them is "improper" and should be avoided. Basically the exception would represent that something weird has happened that we're calling `get_metadata_dimension` on a feature type that doesn't really have one. However, I may be missing something.

fl4health/clients/tabular_data_client.py
```diff
@@ -53,31 +60,40 @@ def setup_client(self, config: Config) -> None:
...
             self.tabular_features_info_encoder = TabularFeaturesInfoEncoder.encoder_from_dataframe(
-                self.df, self.id_column, self.target_column
+                self.df, self.id_column, self.targets
             )
 
     def get_data_frame(self, config: Config) -> pd.DataFrame:
         """
         User defined method that returns a pandas dataframe.
+
+        Args:
```
emersodb, line 95: Looks like this comment has empty args?

examples/feature_alignment_example/misalign_data.py
```diff
@@ -27,6 +27,7 @@ def random_split_data(df: pd.DataFrame, n: int) -> List[pd.DataFrame]:
 
     # Dropping columns to create misalignment.
     df2 = df2.drop(columns=["ExpiredHospital", "admit_type", "NumRx", "ethnicity"])
+    log(INFO, "Hospital2 missing columns: ExpiredHospital, admit_type, NumRx, ethnicity")
```
emersodb, line 30: Very minor, but you could "automate" the log here as 
```
columns_to_drop = ["ExpiredHospital", "admit_type", "NumRx", "ethnicity"]
 df2 = df2.drop(columns=columns_to_drop)
log(INFO, "Hospital2 missing columns: {', '.join(columns_to_drop)}")
```

fl4health/feature_alignment/tabular_feature.py
```diff
@@ -45,7 +45,7 @@ def get_metadata(self) -> MetaData:
         return self.metadata
 
     def get_metadata_dimension(self) -> int:
-        if self.feature_type == TabularType.BINARY or TabularType.ORDINAL:
+        if self.feature_type == TabularType.BINARY or self.feature_type == TabularType.ORDINAL:
```
emersodb, line 48: Nice catch!

fl4health/server/tabular_feature_alignment_server.py
```diff
@@ -58,13 +70,23 @@ def __init__(
         self.config = config
         self.initialize_parameters = initialize_parameters
         self.format_info_gathered = False
+        self.dimension_info: Dict[str, int] = {}
         # casting self.strategy to BasicFedAvg so its on_fit_config_fn can be specified.
```
emersodb, line 74: I think we can remove this comment, since it's no longer a cast?

fl4health/feature_alignment/tab_features_info_encoder.py
fatemetkl: Right, that is something useful to have!

emersodb (approved):
I left some very minor comments that are certainly not "deal-breaking." Changes look great to me. Tests are awesome. I'm good to go. I'd just make sure @fatemetkl is good with the way you addressed her comments and we can merge.

fatemetkl (approved):
The changes look great to me, thanks for adding the test.

fl4health/feature_alignment/tabular_feature.py
yc7z: Sounds good!

PR #64, opened 2023-10-06
Dbe/setting doc string format

emersodb:
# PR Type
[**Feature** | Fix | **Documentation** | Other() ]

# Short Description

This PR addresses two tickets: [Ticket 1](https://app.clickup.com/t/8685w9eeb), [Ticket 2](https://app.clickup.com/t/8685rdwwk)

**The first ticket** is fairly mundane but important. It implements a particular code documentation style for our repository. The style that we'll be using is the Google docstring format. The structure can be automatically generated if using VSCode, which is quite convenient. I've added the guidelines to the ReadMe and updated the doc strings in a bunch of areas of the code as examples and a step towards more formal documentation of the library.

**The second ticket** is a quality of life ticket mainly addressing the approaches that exchange only a portion of the model weights (FENDA, APFL, Dynamic Weight Exchange, FedBN). On the first pass when the client models are being initialized by the server (i.e. the first time set_parameters is called) the clients use a FullParameterExchanger to set all weights in the network. This means that the server sends all weights in the model to be trained, regardless of the strategy, to sync all of the different client's weights. Thereafter, the set parameter exchangers are used to potentially exchange only sub-weights. This means that the server need not really be aware of how the parameter exchange is done under the hood when setting initial weights.

During the work for these two tickets I cleaned up the way SCAFFOLD warm-start happens and also fixed a small bug in the SCAFFOLD control variates calculations if they were being done on CPU (**Note**: This bug doesn't affect our SCAFFOLD experiments for the paper because they were done on GPU). As another QoL change for SCAFFOLD, and also to avoid inconsistencies with the initialization, I made it so that the user code provide initial values for the control variates and the client control variates were automatically set to these or the user could simply provide the model to the SCAFFOLD strategy and the model would be used to set the control variate to zero everywhere.

Finally, we had a few place where the class name casing didn't follow standard, so I just replaced them (Mostly just APFL... to Apfl...)

# Tests Added
Added a set of tests associated with the flow of first pass, set all model parameters, second pass, use the appropriate parameter exchanger.

Also re-ran all of the example to confirm that they all still work as expected.


.gitignore
```diff
@@ -143,6 +143,7 @@ settings.json
 **/datasets/news_classification/partitioned_datasets/**
 **/datasets/mnist_data/**
 **/datasets/MNIST/**
+**/datasets/agnews_data/**
```
emersodb, line 146: Just formally ignoring the agnews dataset used for the dynamic layer exchanger (if you download it as `agnews_data`)

examples/apfl_example/client.py
```diff
@@ -8,7 +8,7 @@
 
 from examples.models.cnn_model import MnistNetWithBnAndFrozen
 from fl4health.clients.apfl_client import ApflClient
-from fl4health.model_bases.apfl_base import APFLModule
+from fl4health.model_bases.apfl_base import ApflModule
```
emersodb, line 11: The standard casing (at least as I'm told, even for acronyms) is not to have all caps. So just changed here and in a bunch of other places to adhere to that run.

examples/federated_eval_example/server.py
```diff
@@ -39,7 +39,7 @@ def main(config: Dict[str, Any], server_checkpoint_path: Optional[Path]) -> None
         action="store",
         type=str,
         help="Path to configuration file.",
-        default="examples/basic_example/config.yaml",
+        default="examples/federated_eval_example/config.yaml",
```
emersodb, line 42: Just a typo fix in the default arg value

examples/fedprox_example/server.py
```diff
@@ -62,10 +54,6 @@ def main(config: Dict[str, Any], server_address: str) -> None:
         config["local_epochs"],
         config["batch_size"],
         config["n_server_rounds"],
-        config["adaptive_proximal_weight"],
```
emersodb, line 65: These weren't used anywhere in the config because they are part of the strategy. So I removed them.

examples/fl_plus_local_ft_example/client.py
```diff
@@ -44,7 +46,9 @@ def get_criterion(self, config: Config) -> _Loss:
     fl.client.start_numpy_client(server_address="0.0.0.0:8080", client=client)
 
     # Run further local training after the federated learning has finished
-    client.train_by_epochs(2)
+    local_epochs_to_perform = 2
+    log(INFO, f"Beginning {local_epochs_to_perform} local epochs of training on each client")
```
emersodb, line 50: Just adding a logging statement for clarity

examples/models/logistic_regression.py
```diff
@@ -7,5 +7,5 @@ def __init__(self, input_dim: int, output_dim: int) -> None:
         self.linear = torch.nn.Linear(input_dim, output_dim)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
-        outputs = torch.sigmoid(self.linear(x)).squeeze()
+        outputs = torch.sigmoid(self.linear(x)).reshape(-1)
```
emersodb, line 10: If the batch size is 1 (which can happen at the end of a dataloader if the len(dataloader) mod batch_size = 1) then squeeze will cause problems in the BCE loss function calculation. Using reshape rather than squeeze avoids this.

examples/scaffold_example/server.py
```diff
@@ -53,14 +49,14 @@ def main(config: Dict[str, Any]) -> None:
         # We use the same fit config function, as nothing changes for eval
         on_evaluate_config_fn=fit_config_fn,
         evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
-        initial_parameters=initial_parameters,
-        initial_control_variates=initial_control_variates,
+        initial_parameters=get_initial_model_parameters(model),
+        model=model,
```
emersodb, line 53: Model is used by the SCAFFOLD strategy to automatically initialize the control variates to zero.

fl4health/clients/basic_client.py
```diff
@@ -88,7 +84,7 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
 
         if local_epochs is not None:
             _, metrics = self.train_by_epochs(local_epochs, current_server_round)
-            local_steps = self.num_train_samples * local_epochs  # total steps over training round
+            local_steps = len(self.train_loader) * local_epochs  # total steps over training round
```
emersodb, line 87: @jewelltaylor: Correct me if I'm wrong, but I believe the LHS was a bug that popped up somewhere? The number of steps is the number of batches in the loader times the number of epochs rather than the number of sample times the number of epochs I think? So I fixed it here.

fl4health/clients/scaffold_client.py
```diff
@@ -48,7 +48,6 @@ def __init__(
         self.client_control_variates_updates: Optional[NDArrays] = None  # delta_c_i in paper
         self.server_control_variates: Optional[NDArrays] = None  # c in paper
         self.optimizer: torch.optim.SGD  # Scaffold require vanilla SGD as optimizer
-        self.server_model_state: Optional[NDArrays] = None  # model state from server
```
emersodb, line 51: This wasn't used anywhere, so I killed it off

fl4health/clients/scaffold_client.py
```diff
@@ -69,24 +68,34 @@ def get_parameters(self, config: Config) -> NDArrays:
...
         self.server_control_variates = server_control_variates
-        self.server_model_state = server_model_state
-        self.parameter_exchanger.pull_parameters(server_model_state, self.model, config)
+
+        super().set_parameters(server_model_state, config)
+
+        # Note that these are weights the require a gradient, because they are used to compute control variates
         self.server_model_weights = [
-            model_params.cpu().detach().numpy()
+            model_params.cpu().detach().clone().numpy()
```
emersodb, line 88: If we don't call clone here and the `model_params` are already on cpu, there is a memory optimization where this numpy array will **share the underlying memory of the parameters with the original tensor**. This is bad, because it means that when the model parameters are updated in training, the values in `self.server_model_weights` are **also updated**, which we really don't want.

fl4health/clients/scaffold_client.py
```diff
@@ -69,24 +68,34 @@ def get_parameters(self, config: Config) -> NDArrays:
...
         ]
 
-        # If client control variates do not exist, initialize with zeros as per paper
+        # If client control variates do not exist, initialize them to be the same as the server control variates.
+        # Server variates default to be 0, but as stated in the paper the control variates should be the uniform
+        # average of the client variates. So if server_control_variates are non-zero, this ensures that average
+        # still holds.
         if self.client_control_variates is None:
-            self.client_control_variates = [np.zeros_like(weight) for weight in self.server_control_variates]
+            self.client_control_variates = copy.deepcopy(self.server_control_variates)
```
emersodb, line 98: This used to be setting it to zeros, which if the user explicitly sets the initial server_control_variates to be something else, say all ones, would violate the assumption of the paper that the initial values for self.server_control_variates = average(client_control_variates). So here we're initializing the control variates to match the server ones

fl4health/clients/scaffold_client.py
```diff
@@ -100,7 +109,9 @@ def update_control_variates(self, local_steps: int) -> None:
         assert self.learning_rate is not None
 
         # y_i
-        client_model_weights = [val.cpu().detach().numpy() for val in self.model.parameters() if val.requires_grad]
+        client_model_weights = [
+            val.cpu().detach().clone().numpy() for val in self.model.parameters() if val.requires_grad
```
emersodb, line 113: This is a similar issue to what was happening above with `self.server_model_weights` if we're working exclusively on CPU. I don't think this causes any issues, as it happens after training, but still doing the clone to be safe.

fl4health/parameter_exchange/parameter_packer.py
```diff
@@ -51,7 +51,10 @@ def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, floa
         # The last entry is extra packed fedprox variable
         split_size = len(packed_parameters) - 1
         model_parameters = packed_parameters[:split_size]
-        extra_fedprox_variable = float(packed_parameters[split_size:][0])
+        # The packed contents should have length 1
+        packed_contents = packed_parameters[split_size:]
+        assert len(packed_contents) == 1
```
emersodb, line 56: Just adding an extra sanity check in case this catches something for us in the future

fl4health/server/scaffold_server.py
```diff
@@ -37,22 +53,30 @@ def __init__(
         )
         self.warm_start = warm_start
 
-    def initialize_paramameters(self, timeout: Optional[float]) -> None:
+    def _get_initial_parameters(self, timeout: Optional[float]) -> Parameters:
```
emersodb, line 56: Rather than having to worry about `_get_initial_parameters` being called after the custom `initialize_paramameters` function for warm-start, we just override the parent `_get_initial_parameters` so that we have control over when the warm start work happens.

requirements.txt
```diff
@@ -7,6 +7,7 @@ nltk
 numpy
 opacus
 pandas
+portalocker
```
emersodb, line 10: When I ran the dynamic layer exchanger example, my install was missing this library.

tests/server/test_base_server.py
```diff
@@ -14,7 +14,7 @@
 model = LinearTransform()
 
 
-class TestFLServer(FlServer):
+class DummyFLServer(FlServer):
```
emersodb, line 17: This was generating a warning in our test collection because it was fronted by the word test. Just changing it to suppress the error.

.github/workflows/static_code_checks.yaml
```diff
@@ -23,5 +23,5 @@ jobs:
       - name: precommit checks
         run: |
           pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install $(grep -v '^torchdata\|^torchtext\|^torcheval' requirements.txt)
```
emersodb, line 26: Skipping installing some packages that are only needed for the examples

.github/workflows/tests.yaml
```diff
@@ -26,6 +26,6 @@ jobs:
       - name: Install Requirements
         run: |
           pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install $(grep -v '^torchdata\|^torchtext\|^torcheval' requirements.txt)
```
emersodb, line 29: Skipping installing some packages that are only needed for the examples

.github/workflows/static_code_checks.yaml
```diff
@@ -23,5 +23,7 @@ jobs:
...
           pre-commit run --all-files
+      - name: Clearing pip for space
+        run: pip uninstall -y -r requirements.txt
```
emersodb, line 29: For whatever reason, we started running out of disk space to save the pip cache on the github runner. This removes the site-packages to save a bit of space and allow the caching to complete

examples/partial_weight_exchange_example/client.py
```diff
@@ -111,11 +111,24 @@ def get_parameters(self, config: Config) -> NDArrays:
         return self.parameter_exchanger.push_parameters(self.model, self.initial_model, config=config)
 
     def set_parameters(self, parameters: NDArrays, config: Config) -> None:
-        # Sets the local model parameters transfered from the server using a parameter exchanger to coordinate how
-        # parameters are set
-        assert self.model is not None and self.parameter_exchanger is not None
-        self.parameter_exchanger.pull_parameters(parameters, self.model, config)
-        # stores the values of the model parameters at the beginning of each training round.
+        """
```
yc7z, line 114: Not a suggestion for change but just a comment: I think it would make sense to create a client class that performs partial weight exchange (using whichever weight selection criterion). Right now there's no intermediate class between `BasicClient` and the very specific client in this example. 

Link to ClickUp: https://app.clickup.com/t/8685zgact

emersodb: I think that's a great idea. I actually also created a ticket for this a few days ago 🙂:

https://app.clickup.com/t/8685xd7uu

I'll merge the two tickets now!

fl4health/server/polling.py
```diff
@@ -37,7 +46,17 @@ def _handle_finished_future_after_poll(
...
+    Get Properties of client. This is run for each client to extract the properties from the target client
+
+    Args:
+        client (ClientProxy): Client proxy representing one of the clients managed by the server
+        ins (GetPropertiesIns): ins provides any configurations required to help the client retrieve the correct
+            properties
+
+    Returns:
+        Tuple[ClientProxy, GetPropertiesRes]: Returns the resulting properties from the client response.
+    """
```
fatemetkl, line 59: Some of the lines in the docstring do not end with a period. I was wondering if this is on purpose, and if so, why?

fatemetkl (approved):
Everything looks good to me!

fl4health/server/client_level_dp_fed_avg_server.py
```diff
@@ -58,6 +84,13 @@ def fit(self, num_rounds: int, timeout: Optional[float]) -> History:
         return super().fit(num_rounds=num_rounds, timeout=timeout)
 
     def setup_privacy_accountant(self, sample_counts: List[int]) -> None:
+        """
+        Sets up FL Accountant and computes privacy loss based on class attributes and retrived sample counts
```
yc7z, line 88: *retrieved

fl4health/strategies/basic_fedavg.py
```diff
@@ -59,51 +51,44 @@ def __init__(
...
-        evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn]
-            Metrics aggregation function, optional.
-        weighted_aggregation: bool, Optional.
-            Defaults to True, determines whether parameter aggregation is a linearly weighted average or a uniform
-            average. FedAvg default is weighted average by client dataset counts.
-        weighted_eval_losses: bool, Optional
-            Defaults to True, determines whether losses during evaluation are linearly weighted averages or a uniform
-            average. FedAvg default is weighted average of the losses by client dataset counts.
+        """
+        Federated Averaging with Flexible Sampling. This implementation extends that of Flower in two ways. The first
```
zxj-c, line 55: Wanted to double check we are intentionally removing the ArXiv reference (line 64 above).

zxj-c (approved):

fl4health/clients/scaffold_client.py
emersodb: Note: This does not cause the same problems if the model is on GPU, as model_params.cpu() makes a copy of the tensor on cpu, distinct from the tensor being trained on the GPU.

fl4health/server/polling.py
emersodb: Not intentional. Just going too fast and missed some punctuation. I'll go through and add them where I find them missing.

fl4health/strategies/basic_fedavg.py
emersodb: Nope. I've added it back. Thanks for the catch.

fl4health/clients/basic_client.py
jewelltaylor: Yeah, I must have accidentally computed the total samples saw through all the epochs in a given training round instead of the number of steps. Good catch!

fl4health/clients/scaffold_client.py
jewelltaylor: Good call, dug around and also came to the conclusion this isn't being used anywhere.

fl4health/clients/scaffold_client.py
jewelltaylor: Very good catch! I was not aware of this at all but will be sure to keep this in mind in the future.

fl4health/clients/scaffold_client.py
jewelltaylor: Nice that we can correctly handle init of control variates in all cases. Thanks for the update :)

jewelltaylor (approved):
Looked through the changes that pertained to Scaffold and everything looks great! One small comment about an unclear comment you had in the ScaffoldClient set_parameters method

PR #65, opened 2023-10-11
Privacy mechanisms & docs 

zxj-c:
# PR Type
[**Feature** | Fix | **Documentation** | Other() ]

# Short Description
This PR adds four privacy mechanisms in the `fl4health/privacy_mechanisms/` dir along with some brief documentation which can be started by running 

```
pip install -r requirements.txt
cd fl4health/privacy_mechanisms 
mkdocs serve
```

# Tests Added
Some simple test cases along with some type checking.


examples/fedopt_example/client.py
```diff
@@ -55,7 +55,7 @@ def train(
             # Report some batch loss statistics every so often to track decrease
             if batch_index % 100 == 0:
                 log(INFO, f"Batch Index {batch_index} of {n_batches}, Batch loss: {loss.item()}")
-            epoch_metrics.update_performance(predicted, labels)
+            epoch_metrics.update_performance(predicted, labels)  # type: ignore[arg-type]
```
emersodb, line 58: This issue should be fixed with the merging of a few other PRs. So if you merge in the latest changes from main, you should be able to get rid of these type ignores now.

examples/fedopt_example/client.py
```diff
@@ -93,7 +93,7 @@ def validate(
             out = model(data, (h0, c0))
             loss += criterion(out, labels).item()
             _, predicted = torch.max(out.data, 1)
-            epoch_metrics.update_performance(predicted, labels)
+            epoch_metrics.update_performance(predicted, labels)  # type: ignore[arg-type]
```
emersodb, line 96: See above

fl4health/privacy_mechanisms/dpBinomial.py
```diff
@@ -0,0 +1,38 @@
+from typing import List
+
+import torch
+
+
+# TODO why doesn't mechanism result in output type List[int]
+def BinomialMechanism(query_vector: List[int], N: int, p: float, j: int) -> List[int]:
```
emersodb, line 7: We've generally tried to adhere to the PEP 8 guidelines for casing of things like classes and functions (though we're definitely not perfect at it). If you're curious about the standard you can take a look [here](https://peps.python.org/pep-0008/) under sections like "[Class Names](https://peps.python.org/pep-0008/#class-names)". 

The two most common to know are 
- functions are lowercase with "words" separated by underscores (i.e. `binomial_mechanism`)
- class names are "camel cased" (i.e. `FedProxStrategy`)

fl4health/privacy_mechanisms/docs/index.md
```diff
@@ -0,0 +1,89 @@
...
+``` py title="dpDiscreteGaussian.py" linenums="1"
+def DiscreteGaussianMechanism(query_vector: List[int],
+                                variance: float) -> List[int]:
+
+```
+
+## 2. Poisson Binomial Mechanism
+
+
+!!! info inline end "Note"
```
emersodb, line 31: I'm not very familiar with this annotation for markdown, but it seems like maybe this is orphaned or in an odd spot, given where `!!! info ""` starts. I might be wrong though?

fl4health/privacy_mechanisms/docs/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+    This is the only non-additive privacy mechanism implemented. It gives an unbiased estimator of the mean.
+
+In the scalar version of this mechanism, a real valued input $x$ is assoicated with a
+binomial random variable $$Z\sim \text{Binom}(m, p(x))$$ where $m$ is the modulus for moular arithmetic used in SecAgg and the success probability $p(x)$ encodes user data $x \in \mathbb{R}$.
```
emersodb, line 36: *modular arithmetic

fl4health/privacy_mechanisms/dpBinomial.py
emersodb: Note that the casing generally overrides something being a proper noun (i.e. `gaussian_mechanism` over `Gaussian_mechanism`)

fl4health/privacy_mechanisms/dpBinomial.py
```diff
@@ -0,0 +1,38 @@
...
+    Args:
+        query_vector (List[int]): A vector with integer valued components.
+        N (int): The number N of Bernoulli trials which determines the distribution Bin(N, p).
+        p (float): The success probability p which determines the distribution Bin(N, p).
+        j (int): Determines the quantization scale s = 1/j.
+
+    Returns:
+        List[int]: Privatized query vector via the scaler binomial mechanism applied to each component of query_vector
+
+    Reference
```
emersodb, line 20: This reference should go above args in the description section, I think.

fl4health/privacy_mechanisms/dpBinomial.py
```diff
@@ -0,0 +1,38 @@
...
+        cpSGD: Communication-efficient and differentially-private distributed SGD
+        https://proceedings.neurips.cc/paper_files/paper/2018/file/21ce689121e39821d07d04faab328370-Paper.pdf
+    """
```
emersodb, line 23: @masi-sh: Please double check my understanding here

Based on the reference above, the release mechanism M is actually a map from f(D) -> R^d. So even though you get a discrete vector in the form of f(D), M(f(D)) is a real valued vector. So the annotation List[int] for the output isn't correct. 

I'm also a little confused as to why Z (which is the variable `binomial_samples` here) is being distributed as Bin(d, p), where d is the dimension of the query vector rather than Bin(N, p) here. My understanding of `torch.distributions.Binomial` is that the first argument is the number of trials to run for each p value in the vector passed to the second argument.

fl4health/privacy_mechanisms/dpBinomial.py
```diff
@@ -0,0 +1,38 @@
...
+
+    binomial_samples = torch.distributions.Binomial(dim, p * torch.ones(dim)).sample()
+    query = torch.tensor(query_vector)
+
+    privatized_vector = query + (binomial_samples - N * p) / j
+
+    return privatized_vector.tolist()
+
+
+if __name__ == "__main__":
```
emersodb, line 36: Rather than using a main here for testing, we should add formal unit tests to our tests suite (found in the `tests/` folder. In addition to checking the type here. I'd advise that you fix the torch random seed in the test and make sure that the call to the mechanism produces the correct values.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+
+import torch
+
+
+def Bernoulli_exp(gamma: float) -> int:
+    """
+    Draws a sample from Bernoulli(exp(-gamma))
+
+    Args:
+        gamma (float): A non-negative parameter of type float.
```
emersodb, line 12: No need to specify the type in the comment, as it is already annotated in the args typing. An intuitive description of what gamma is would be useful (if there is one)

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Returns:
+        int: A Bernoulli sample, either 0 or 1
+
+    Reference:
```
emersodb, line 17: Reference can go in the description above args.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+        Algorithm 2 of "The Discrete Gaussian for Differential Privacy"
+        https://proceedings.neurips.cc/paper/2020/file/b53b3a3d6ab90ce0268229151c9bde11-Paper.pdf
+    """
+
+    assert gamma >= 0  # ensures probabiity is not over 1
+
+    if 0 <= gamma <= 1:
```
emersodb, line 24: very minor, but this can be collapsed to `if gamma<=1` due to the assert above.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+            K += 1
+            sample_A = torch.bernoulli(g / K)
+
+        return K % 2
+
+    # if gamma > 1
+    G, p = math.floor(gamma), torch.exp(torch.tensor(-1))
+
+    for _ in range(G):
+        sample_B = torch.bernoulli(p)
```
emersodb, line 38: Based on my reading of Algorithm 2 from the reference paper, this is supposed to be a recursive call, similar to sample_C. That is, this line should be `Bernoulli_exp(p.item())` rather than an explicit bernoulli sample.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+        int: An integer sample from the centered discrete Gaussian distribution with given variance.
+
+    Reference
+        Algorithm 1 in "The Discrete Gaussian for Differential Privacy"
+        https://proceedings.neurips.cc/paper/2020/file/b53b3a3d6ab90ce0268229151c9bde11-Paper.pdf
+    """
+
+    assert variance > 0  # Requires variance > 0
+
+    t = 1 + math.floor(variance)
```
emersodb, line 66: In Algorithm 1 from this paper, it looks like t should be
```
t = math.floor(math.sqrt(variance)) + 1
```
That is, I believe you need t to be set to the floor of the standard deviation not the variance.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+
+    while True:
```
emersodb, line 68: This scares me a little bit 😂. Could we add a counter in the loop and warn the user if a large number of loops have been completed. Something like
```
if counter % 100:
    log(WARNING, f"Discrete Gaussian Sampling has completed {counter} samples without success.")
```

emersodb: Note, counter can start at 1 to avoid logging this error the first time.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+    Applies additive discrete Gaussian noise to query_vector.
+
+    Args:
+        query_vector (List[int]): This is the discretized gradient vector in the SecAgg context.
+        variance (float): Gaussian noise variance.
+
+    Returns:
+        List[int]: Privitized vector.
+
+    Reference
```
emersodb, line 112: Reference to method description 🙂

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+def DiscreteGaussianMechanism(query_vector: List[int], variance: float) -> List[int]:
+    """
+    Applies additive discrete Gaussian noise to query_vector.
+
+    Args:
+        query_vector (List[int]): This is the discretized gradient vector in the SecAgg context.
+        variance (float): Gaussian noise variance.
+
+    Returns:
+        List[int]: Privitized vector.
```
emersodb, line 110: Super minor, but I think Privitized should be Privatized? This spelling change would take place in a number of areas of the code.

fl4health/privacy_mechanisms/dpDiscreteGaussian.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Reference
+        The Distributed Discrete Gaussian Mechanism for Federated Learning with Secure Aggregation
+        http://proceedings.mlr.press/v139/kairouz21a/kairouz21a.pdf
+    """
+
+    dim = len(query_vector)
+    assert dim > 0  # query_vector needs to be nonempty
+
+    return [query_vector[i] + DiscreteGaussianSampler(variance) for i in range(dim)]
```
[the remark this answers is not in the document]
emersodb: I'm assuming that you'll do the modular arithmetic from the paper elsewhere? In the paper it looks like the noised values are
```
z_i = (\tilde{x}_i + y_i) mod m
```
Just wanted to check.

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+    Raises:
+        Exception: Invalid parameter N
+        Exception: Invaid radius
+        Exception: Invalid theta
+
+    Returns:
+        int: A random sample taken from the bionomial distribution with success probability
+        p = theta * x / radius + 1/2 and trial number N.
+
+    Reference
```
emersodb, line 28: Reference to method description section.

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Args:
+        x (float): Scalar data to be encoded.
+        N (int): Number of Bernoulli trials that yields the binomial variable (We have N = modulus - 1).
+        radius (float): The parameter x must be within the real interval [-radius, radius].
+        theta (float): a parameter within the real interval [0, 1/4].
+
+    Raises:
+        Exception: Invalid parameter N
+        Exception: Invaid radius
```
emersodb, line 21: Invalid* 🙂

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+from typing import List
+
+import torch
+from scipy.stats import ortho_group
+from torch.distributions import Binomial
+
+
+def ScalerPoissonBinomial(x: float, N: int, radius: float, theta: float) -> int:
+    """
+    Privacy mechanism which encodes scalar data x as a bionomial random variable.
```
emersodb, line 11: binomial*

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Args:
+        x (float): Scalar data to be encoded.
+        N (int): Number of Bernoulli trials that yields the binomial variable (We have N = modulus - 1).
```
emersodb, line 15: The "N=modulus -1" comment here is a bit off-hand. Is this referring to something in the paper? If so, I'd reference where.

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Reference
+        The Poisson Binomial Mechanism for Unbiased Federated Learning with Secure Aggregation
+        https://proceedings.mlr.press/v162/chen22s/chen22s.pdf
+    """
+
+    if N < 1 or type(N) != int:
+        raise Exception("The parameter N of Binom(N, p) must be a natural number")
+
+    if radius < 0:
```
emersodb, line 36: You also need to check that x is in the interval [-radius, radius] correct?

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+        raise Exception("The radius for the interval must be positive")
+
+    if not (0 <= theta <= 0.25):
+        raise Exception("theta must be within the real interval [0, 1/4]")
+
+    # rescaling
+    p = theta * x / radius + 1 / 2
```
emersodb, line 43: Super minor, but you might as well use `theta* x / radius + 0.5` rather than forcing the division here

fl4health/privacy_mechanisms/dpBinomial.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Reference
+        cpSGD: Communication-efficient and differentially-private distributed SGD
+        https://proceedings.neurips.cc/paper_files/paper/2018/file/21ce689121e39821d07d04faab328370-Paper.pdf
+    """
+
+    dim = len(query_vector)
+    assert dim > 0  # nonempty query_vector
+
+    binomial_samples = torch.distributions.Binomial(dim, p * torch.ones(dim)).sample()
```
emersodb, line 28: rather than using the full name here, you could just do
```
from torch.distributions import Binomial
```
as you do a bit later in the PR

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # privatization
+    B = Binomial(N, p)
```
emersodb, line 46: I don't think you need to store this. You can just do
```
return Binomial(N, p).sample().int().item()
```

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+    Args:
+        dim (int): Dimension.
+
+    Returns:
+        List[List[int]]: Replacement of the tight frame in Kashin representation.
+    """
+
+    sample_matrix = ortho_group(dim).rvs()
+
+    return sample_matrix.tolist()
```
emersodb, line 64: Any reason we're converting to a `List[List[int]]` here rather than using an `np.ndarray`?

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def PoissonBinomialMechanism_Client(
+    query_vector: List[float], radius: float, tight_frame: List[float], theta: float, K: float, modulus: int
+) -> List[int]:
+    """
+    Privatization of client vector with Poisson bionomial mechanism.
+
+    Args:
+        query_vector (List[float]): Client integer vector.
```
emersodb, line 74: The type annotation here is a float. Is that correct given the comment specifying an integer vector?

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+    return estimator.tolist()
+
+
+if __name__ == "__main__":
+    # # ScalerPoissonBinomial test cases
+    # x, N, radius, theta = 0, 20, 5, 0.1
+    # out = [ScalerPoissonBinomial(x, N, radius, theta) for _ in range(100)]
+    # print(out)
+
+    # PoissonBinomialMechanism test cases
```
emersodb, line 159: I would recommend making this a formal test and also running tests on the various helper functions to ensure that they are working properly.

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+        K (float): Constant parameter in Kashin representation, or clip.
+        modulus (int): The modulus for modular arithmetic used in SecAgg.
+
+    Raises:
+        Exception: Bad query.
+
+    Returns:
+        List[int]: The discretized and privatized query_vector as a list of integers.
+
+    Reference
```
emersodb, line 87: Move to description body

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+def PoissonBinomialMechanism_Client(
+    query_vector: List[float], radius: float, tight_frame: List[float], theta: float, K: float, modulus: int
+) -> List[int]:
+    """
+    Privatization of client vector with Poisson bionomial mechanism.
+
+    Args:
+        query_vector (List[float]): Client integer vector.
+        radius (float): The radius.
```
emersodb, line 75: Being a bit more explicit here would be helpful. From what I understand, it reflects the bound on the l_2 norm of the query_vector

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+    if dim == 0:
+        raise Exception("query_vector cannot be empty")
+
+    kashin_matrix = torch.tensor(tight_frame)
+    client_vector = torch.tensor(query_vector)
+
+    kashin_vector = torch.matmul(kashin_matrix, client_vector)
+
+    # TODO adjust this for l-infinity clipping
+    scalar_radius = radius * K / math.sqrt(dim)
```
emersodb, line 103: super minor, but I think you mean `scaled_radius` here? The radius is already a "scalar" 🙂

fl4health/privacy_mechanisms/dpPoissonBinomial.py
```diff
@@ -0,0 +1,162 @@
...
+
+def PoissonBinomialMechanism_Server(
+    inverse_matrix: List[List[int]], radius: float, modulus: int, theta: float, *client_vectors: List[int]
+) -> List[int]:
+    """
+    Server side function that recovers the sum of vectors from clients.
+
+    Args:
+        inverse_matrix (List[List[int]]): The inverse of the random rotation generated by Kashin_alternative().
+        radius (float): Radius.
```
emersodb, line 120: A bit more detail would be useful here. It appears that this should be the scaled version of the radius if I'm reading the algorithm correctly. That is, radius should be c' = \frac{cK}{\sqrt{d}}?

fl4health/privacy_mechanisms/dpSkellam.py
```diff
@@ -0,0 +1,58 @@
...
+         skellam_variance (float): Variance associated with each component of the d-dimensional Skellam distribution.
+
+     Raises:
+         Exception: Empty query
+         Exception: Un-discretized query.
+
+     Returns:
+         list[int]: Perturbed vector.
+
+    Notes
```
emersodb, line 21: Move to the method description body

fl4health/privacy_mechanisms/dpSkellam.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def SkellamMechanism(query_vector: list[int], skellam_variance: float) -> list[int]:
+    """
+     An additive mechanism that adds to the given query vector a discrete noise vector
+     sampled from the centered Skellam distribution of given variance.
+
+
+     Args:
+         query_vector (list[int]): Discritized vector to which noise will be added.
```
emersodb, line 11: discretized*

fl4health/privacy_mechanisms/dpSkellam.py
```diff
@@ -0,0 +1,58 @@
...
+
+    if dim == 0:
+        raise Exception("query_vector cannot be empty")
+
+    if type(query_vector[0]) != int:
+        raise Exception("Did you forget to discretize the gradient vector? query_vector must be a list of integers")
+
+    # mean of Possion variables used to construct a
+    # centered Skellam variable S with Var(S) = skellam_variance
+    mean = skellam_variance / 2
```
emersodb, line 41: Maybe name this something like `poisson_mean` to make it explicit that this is the poisson expectation we're going to use.

It also might be worth noting that Var(S) = \mu_1 + \mu_2 when S ~ Poisson(mu_1) - Poisson(mu_2)?

fl4health/privacy_mechanisms/dpSkellam.py
```diff
@@ -0,0 +1,58 @@
...
+    mean_vector = torch.ones(dim) * mean
+
+    Skellam = torch.poisson(mean_vector) - torch.poisson(mean_vector)
```
emersodb, line 44: variables don't tend to have capital names in the PEP8 standard.

fl4health/privacy_mechanisms/dpSkellam.py
```diff
@@ -0,0 +1,58 @@
...
+    Skellam = Skellam.int()  # convert to 32-bit signed integer
+
+    perturbed = torch.tensor(query_vector, dtype=torch.int) + Skellam  # add noise
+
+    return perturbed.tolist()
+
+
+if __name__ == "__main__":
+    # test cases & type checking
```
emersodb, line 53: Suggest moving this into test to create formal unit tests and make sure everything is working as expected.

fl4health/privacy_mechanisms/dpBinomial.py
zxj-c: Good catch, the correct parameter should be N.

You are absolutely right about your first point, I was also talking with @masi-sh about this. In the formula both $p$ and $s$ would make the output not an integer.

fl4health/privacy_mechanisms/dpBinomial.py
zxj-c: I was hoping doing this emphasize to the code readers this particular `Binomial` is from torch and not one I defined in this file which is also the "binomial" mechanism. Actually I should do as you suggested but leave a comment to avoid ambiguity.

fl4health/privacy_mechanisms/dpBinomial.py
emersodb: I learned that github can render latex. Very cool.

PR #67, opened 2023-10-27
Add contrastive losses

sanaAyrml:
# PR Type
[Feature]

# Short Description
Added Cosine Similarity, Contrastive Loss and perFCL Loss for fender experiments.


examples/fenda_example/README.md
```diff
@@ -1,6 +1,6 @@
 # FENDA Federated Learning Example
 This example provides an example of training a FENDA type model on a non-IID subset of the MNIST data. The FL server
-expects two clients to be spun up (i.e. it will wait until two clients report in before starting training). Each client
+expects three clients to be spun up (i.e. it will wait until three clients report in before starting training). Each client
```
emersodb, line 3: Good catch. Thanks for the edit.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+    def __init__(
+        self,
+        data_path: Path,
+        metrics: Sequence[Metric],
+        device: torch.device,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        metric_meter_type: MetricMeterType = MetricMeterType.AVERAGE,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        temperature: float = 0.5,
+        perFCL_loss_weights: List[float] = [0.0, 0.0],
```
emersodb, line 27: I believe that there are only two of these weights, corresponding to mu and gamma from the paper. So we should make this of type Tuple[float, float] to constrain the potential values.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        cos_sim_loss_weight: float = 0.0,
+        contrastive_loss_weight: float = 0.0,
+    ) -> None:
```
emersodb, line 30: Some documentation in the Google format would be useful here.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+            loss_meter_type=loss_meter_type,
+            metric_meter_type=metric_meter_type,
+            checkpointer=checkpointer,
+        )
+        self.perFCL_loss_weights = perFCL_loss_weights
+        self.cos_sim_loss_weight = cos_sim_loss_weight
+        self.contrastive_loss_weight = contrastive_loss_weight
+        self.cos_sim = torch.nn.CosineSimilarity(dim=-1)
+        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
+        self.temperature: float = temperature
```
emersodb, line 44: I don't think you need the type annotation for the `self.temperature` attribute here?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+            device=device,
+            loss_meter_type=loss_meter_type,
+            metric_meter_type=metric_meter_type,
+            checkpointer=checkpointer,
+        )
+        self.perFCL_loss_weights = perFCL_loss_weights
+        self.cos_sim_loss_weight = cos_sim_loss_weight
+        self.contrastive_loss_weight = contrastive_loss_weight
+        self.cos_sim = torch.nn.CosineSimilarity(dim=-1)
+        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
```
emersodb, line 43: Do we need to move this criterion to device? 

On a related note, this does highlight something that might be worth folding into this PR, but in the `setup_client` of `BasicClient` when we call `self.get_criterion(config)` should we actually call `self.get_criterion(config).to(self.device)` by default? 
@jewelltaylor: What do you think?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        self.temperature: float = temperature
+
+        # Need to save previous local module, global module and aggregated global module at each communication round
+        # to compute contrastive loss.
```
emersodb, line 47: These should probably be optional depending on whether we're actually going to be using any auxiliary losses I think?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+
     def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         assert isinstance(self.model, FendaModel)
         return FixedLayerExchanger(self.model.layers_to_exchange())
+
+    def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
+
+        preds = self.model(input)
+
+        preds["old_local_features"] = self.old_local_module.forward(input).view(len(preds["local_features"]), -1)
```
emersodb, line 60: These predictions only need to be present when performing auxiliary loss calculations correct?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        preds["old_global_features"] = self.old_global_module.forward(input).view(len(preds["global_features"]), -1)
```
emersodb, line 61: I think, but I'm not sure, that `reshape` is a more general tensor transformation than `view`. That is, reshape will perform a view if possible and return a copy if not. I'm not 100% sure though.

I believe that we're making an assumption about how the features are structured here. That is, I think we're assuming they are "batch first"? If that's true we should at least leave a comment.

fl4health/clients/fenda_client.py
emersodb: Will len(tensor) always return the size of the first dimension? Just want to double check.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        self.old_global_module: torch.nn.Module
+        self.aggregated_global_module: torch.nn.Module
+
     def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         assert isinstance(self.model, FendaModel)
         return FixedLayerExchanger(self.model.layers_to_exchange())
+
+    def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
+
+        preds = self.model(input)
```
emersodb, line 58: It doesn't look like the FendaModel forward has been changed in this PR. So I don't think preds will contain the keys `'local_features'` and `'global_features'` but maybe I'm missing where that gets set.

fl4health/clients/fenda_client.py
emersodb: It also looks like we're already performing this operation in the forward of the `fenda_base`. Do we want to assume that it has already been done in the forward and eliminate the operation here, or not perform the flattening in the forward and perform it here?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+
+        preds["old_local_features"] = self.old_local_module.forward(input).view(len(preds["local_features"]), -1)
+        preds["old_global_features"] = self.old_global_module.forward(input).view(len(preds["global_features"]), -1)
+        if self.perFCL_loss_weights[0] != 0.0 or self.perFCL_loss_weights[1] != 0.0:
+            preds["aggregated_global_features"] = self.aggregated_global_module.forward(input).view(
+                len(preds["global_features"]), -1
+            )
+
+        if isinstance(preds, dict):
```
emersodb, line 67: We're already expecting preds to be a dictionary for the FENDA client, so I don't think we need this set of checks?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        elif isinstance(preds, torch.Tensor):
+            return {"prediction": preds}
+        else:
+            raise ValueError("Model forward did not return a tensor or dictionary or tensors")
+
+    def set_parameters(self, parameters: NDArrays, config: Config) -> None:
+
+        # Save the parameters of the old model
+        assert isinstance(self.model.local_module, torch.nn.Module)
+        assert isinstance(self.model.global_module, torch.nn.Module)
```
emersodb, line 78: Rather than doing an assert on these submodules, could you rather assert that self.model is a FENDA model, making these two submodules proper nn.Modules?

examples/fedprox_example/server.py
```diff
@@ -54,6 +62,10 @@ def main(config: Dict[str, Any], server_address: str) -> None:
         config["local_epochs"],
         config["batch_size"],
         config["n_server_rounds"],
+        config["adaptive_proximal_weight"],
+        config["proximal_weight"],
+        config["proximal_weight_delta"],
+        config["proximal_weight_patience"],
```
emersodb, line 68: I might be missing something, but I removed these from the config, because I didn't think they were used anywhere on the client-side. That is, they are set in the strategy class below in this server. So I didn't think they needed to be sent as part of the client config. I might be wrong though.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        self.old_local_module = copy.deepcopy(self.model.local_module)
+        self.old_global_module = copy.deepcopy(self.model.global_module)
```
emersodb, line 80: For the first round of federated learning, these weights are going to be random aren't they? If that's true I don't know if they will be useful for contrastive learning.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+        for param in self.old_local_module.parameters():
+            param.requires_grad = False
+        for param in self.old_global_module.parameters():
```
emersodb, line 83: Since we're doing this a few times maybe abstract the setting of all params.requires_grad = False and setting it to eval mode into a function?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+            param.requires_grad = False
+        self.old_local_module.eval()
+        self.old_global_module.eval()
+
+        # Set the parameters of the model
+        output = super().set_parameters(parameters, config)
```
emersodb, line 89: Output here is none (I think) so you wouldn't need to store it.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,220 @@
...
+
+        # Save the parameters of the global model
+        if self.perFCL_loss_weights[0] != 0.0 or self.perFCL_loss_weights[1] != 0.0:
+            assert isinstance(self.model.global_module, torch.nn.Module)
+            self.aggregated_global_module = copy.deepcopy(self.model.global_module)
+            for param in self.aggregated_global_module.parameters():
+                param.requires_grad = False
+            self.aggregated_global_module.eval()
+        return output
```
emersodb, line 98: Since the function return type is none, this return can be omitted.

fl4health/clients/fenda_client.py
sanaAyrml: Yes I change reshape the output in `fenda_base.py` .

fl4health/clients/fenda_client.py
sanaAyrml: Oh no for global module and local module it is not done yet. In fenda model I do the reshape (`.view`) in the head module, not in each local and global modules separately. That's why we are sure of preds["local_features"] and preds["global_features"] dimensionality and here we reshape the output of each module accordingly.

sanaAyrml: Yes that is true, I rechecked the difference between `reshape` and `view`. Although I don't think `view` would be problematic, I will change it to `reshape` to be more generalized. Thanks for mentioning.
Also yes we assume that batch should be the first dimension. I will add the comment.

fl4health/clients/fenda_client.py
sanaAyrml: True, I just got this from basic client in case if in the future the fenda_base get changed or different models get added for it. I can omit it.

fl4health/clients/fenda_client.py
sanaAyrml: These assertion were needed for mypy because of `copy.deepcopy` in the next lines. Also as we only need previous global module and local module I don't make a copy of whole module which also has head module.

fl4health/clients/fenda_client.py
sanaAyrml: Yeah that is true, I ran all of my experiments with fedavg warmup, maybe we can add a comment to point this out and refer them to use warm up module if they don't want them to be random?
Or one other option would be for the first round we shut down and possible contrastive loss and then activate it?

fl4health/clients/fenda_client.py
sanaAyrml: In my experience, I think generally it is good to bring all losses on device, however I think it would not particularly change anything from cross entropy loss here, so I deleted it for now.

fl4health/clients/basic_client.py
```diff
@@ -360,6 +361,18 @@ def set_optimizer(self, config: Config) -> None:
         assert not isinstance(optimizer, dict)
         self.optimizer = optimizer
 
+    def clone_model(self, model: nn.Module) -> nn.Module:
```
emersodb, line 364: This does a bit more than clone the model. It also makes it "untrainable." So I might suggest calling this `clone_and_freeze_model` or something similar.

For method comments, we're trying to use the Google Docstring format. VS Code will autogenerate the structure for you if you follow the autoDocstring installation instructions in the README.MD

fl4health/clients/fenda_client.py
emersodb: I think I was unclear as to what I meant could be optional. I was referring to the modules below. Something like: 
```
self.old_local_module: Optional[torch.nn.Module] = None
self.old_global_module: Optional[torch.nn.Module] = None
self.aggregated_global_module: Optional[torch.nn.Module] = None
```

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        self.old_local_module: torch.nn.Module
+        self.old_global_module: torch.nn.Module
+        self.aggregated_global_module: torch.nn.Module
+
     def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         assert isinstance(self.model, FendaModel)
         return FixedLayerExchanger(self.model.layers_to_exchange())
+
+    def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
+
```
emersodb, line 71: I think we had discussed putting a comment here about assuming that the input is "batch first." Maybe it was just missed in the refactor?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        preds = self.model(input)
+
+        if self.contrastive_loss_weight or self.perFCL_loss_weights:
+            preds["old_local_features"] = self.old_local_module.forward(input).view(len(preds["local_features"]), -1)
```
emersodb, line 75: We're still doing view here instead of reshape (as done in the two below). Is there a reason we stuck with view here?

fl4health/clients/fenda_client.py
emersodb: Yes, I can see why you need to do some kind of typing assertion for mypy and that you're not copying the full fenda model. However, I was suggesting we just assert that self.model is a FendaModel type and mypy should be able to confirm that the submodules have the proper type. I tested this 
```        
assert isinstance(self.model, FendaModel)
self.old_local_module = copy.deepcopy(self.model.local_module)
self.old_global_module = copy.deepcopy(self.model.global_module)
```
and it seems like mypy was happy with that. It's a very minor code simplification.

fl4health/clients/fenda_client.py
emersodb: I think that makes a lot of sense. With that in mind. I'm happy for us to just do that by default (i.e. self.cos_sim and self.ce_criterion to device) and also change the BasicClient setup to perform `self.get_criterion(config).to(self.device)` in the `setup_client`

fl4health/clients/fenda_client.py
emersodb: Did you get a sense for what they do in the paper? If they had an approach to this, we can go with theirs. Otherwise, I'd say the later makes, where we only use previous client model comparison losses after the first training pass. 

It's a little tricky because some clients may not participate in every round and so their first round of client training may not necessarily be the first `server_round`. However, I think this would easily be accomplished by just moving the storage of these modules to the `get_parameters` function and any time we go to apply the loss, we check if the `old_local_module` and `old_global_module` values have been set. If they haven't then the loss is skipped.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        if self.contrastive_loss_weight or self.perFCL_loss_weights:
+            self.old_local_module = self.clone_model(self.model.local_module)
+            self.old_global_module = self.clone_model(self.model.global_module)
+
+        # Set the parameters of the model
+        super().set_parameters(parameters, config)
+
+        # Save the parameters of the global model
+        if self.perFCL_loss_weights:
+            assert isinstance(self.model.global_module, torch.nn.Module)
```
emersodb, line 99: I think if we do the `assert isinstance(self.model, FendaModel)`, this assert could also be removed.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        features of fenda model.
+        """
+        assert len(local_features) == len(global_features)
+        return torch.abs(self.cos_sim(local_features, global_features)).mean()
+
+    def get_contrastive_loss(
+        self, local_features: torch.Tensor, old_local_features: torch.Tensor, global_features: torch.Tensor
+    ) -> torch.Tensor:
+        """
+        Contrastive loss airms to enhance the similarity between the current local features and old local feature
```
emersodb, line 116: "airms" -> "aims" and "local feature" to "local features"

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        """
+        assert isinstance(self.temperature, float)
+        assert len(local_features) == len(old_local_features)
+        posi = self.cos_sim(local_features, old_local_features)
+        logits = posi.reshape(-1, 1)
+        assert len(local_features) == len(global_features)
+        nega = self.cos_sim(local_features, global_features)
+        logits = torch.cat((logits, nega.reshape(-1, 1)), dim=1)
+        logits /= self.temperature
+        labels = torch.zeros(local_features.size(0)).to(self.device).long()
```
emersodb, line 128: Based on a bit of decomposition, I think the calculation you're doing here is correct, but it is a bit hard to read based on the manipulations and we perform pretty similar calculations in the get_perFCL_loss() below. So I think there is some merit to constructing a generic function that computes the `Q(., .)` between two vectors and the corresponding loss then calling them in these functions. Something like 
```
def contrastive_measure(z_1: torch.Tensor, z_2: torch.Tensor, temperature: float) -> torch.Tensor:
    similarity = self.cos_sim(z_1, z_2)
    scaled_sim = similarity/temperature
    return torch.exp(scaled_sim)

def constrastive_loss(reference: torch.Tensor, positive: torch.Tensor, negative: torch.Tensor, temperature: float) -> torch.Tensor:
    Q_num = contrastive_measure(reference, positive, temperature)
    Q_denom = contrastive_measure(reference, negative, temperature)

    loss = -torch.log(Q_num/(Q_num + Q_denom)).mean()
```
Then these can be called in various ways to form the losses?

emersodb: I would also recommend adding tests for each of the loss calculations to ensure they are being performed correctly.

fl4health/clients/fenda_client.py
sanaAyrml: That's true, though temperature can be optional too. should we leave it as optional?

fl4health/clients/fenda_client.py
sanaAyrml: I don't get why it would be different if we move the module storage to get parameters function. For me it is more straight forward to store them when we set new parameters to our model.

sanaAyrml: So the paper does not provide their code and they have not mentioned anything specific about this in the text.

fl4health/clients/fenda_client.py
sanaAyrml: I get this kind of contrastive loss computation from MOON code, and I am not sure about changing it to these losses, as I have always used this version of supervised contrastive loss computing. Also this version has an extra torch.exp and torch.log that with this code we escape extra overhead of them. 
Although I can make a separate function to calculate contrastive loss and call it in PerFCL_loss function though, rather than compute it right there if it is so confusing now.

fl4health/clients/fenda_client.py
emersodb: Yeah for sure. I'm good with the `__init__` arguments as you've edited them

fl4health/clients/fenda_client.py
emersodb: Of course they didn't 😂. Well let's go with your suggestion of no contrastive loss in the first round, as I think that makes the most sense.

fl4health/clients/fenda_client.py
emersodb: I think it would be good to encapsulate it into a callable function that can be used by both this method and `get_perFCL_loss`. I'm good if we leave the calculation as is. It's just takes a bit of thinking to understand how the cross entropy loss is helping you here. There are likely optimizations in the torch loss, but in that loss, you still need to do the exponentiation and log calculations, its just being done by the torch library rather than by us.

Let's leave it as is (other than the encapsulation) and just make sure we write a few tests to verify the calculations are correct.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+            loss_meter_type: Type of loss meter to be used.
+            metric_meter_type: Type of metric meter to be used.
+            checkpointer: Checkpointer to be used for checkpointing.
+            temperature: Temperature to be used for contrastive loss.
+            perFCL_loss_weights: Weights to be used for perFCL loss.
+            Each value associate with one of two contrastive losses in perFCL loss.
+            cos_sim_loss_weight: Weight to be used for cosine similarity loss.
+            contrastive_loss_weight: Weight to be used for contrastive loss.
+        """
+        self.perFCL_loss_weights = perFCL_loss_weights
```
emersodb, line 53: This is sort of ambiguous in the python standards, but we've unified on having acronyms be lower case when they are variables and only the first letter by capitalized if they are classes.

That is, `perFCL_loss_weights` -> `perfcl_loss_weights` or `per_fcl_loss_weights` as a variable and `PerFclClass` if it were a class object.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        logits = posi.reshape(-1, 1)
+        assert len(local_features) == len(global_features)
+        nega = self.cos_sim(local_features, global_features)
+        logits = torch.cat((logits, nega.reshape(-1, 1)), dim=1)
+        logits /= self.temperature
+        labels = torch.zeros(local_features.size(0)).to(self.device).long()
+
+        return self.ce_criterion(logits, labels)
+
+    def get_perFCL_loss(
```
emersodb, line 135: See comment about capitalization of variables and methods above.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        self,
+        local_features: torch.Tensor,
+        old_local_features: torch.Tensor,
+        global_features: torch.Tensor,
+        old_global_features: torch.Tensor,
+        aggregated_global_features: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        PerFCL loss consists of two contrastive losses.
+        First one airms to enhance the similarity between the current global features and aggregated global feature
```
emersodb, line 145: airms -> aims and global feature -> global features

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        as positive pairs while reducing the similarity between the current global features and old global
+        features as negative pairs.
+        Second one airms to enhance the similarity between the current local features and old local feature
```
emersodb, line 148: airms -> aims and local feature -> local features

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        as positive pairs while reducing the similarity between the current local features and aggregated global
+        features as negative pairs.
+        """
+
+        assert isinstance(self.temperature, float)
```
emersodb, line 153: I'm not 100% sure which is better, but it feels like rather than asserting this is an instance of float, we can more clearly do
`assert self.temperature is not None`

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        aggregated_global_features: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        PerFCL loss consists of two contrastive losses.
+        First one airms to enhance the similarity between the current global features and aggregated global feature
+        as positive pairs while reducing the similarity between the current global features and old global
+        features as negative pairs.
+        Second one airms to enhance the similarity between the current local features and old local feature
+        as positive pairs while reducing the similarity between the current local features and aggregated global
+        features as negative pairs.
```
emersodb, line 150: It might be helpful, for a reader of the code to connect each of the inputs to this loss to the variables in the paper. Something like 
```
local features (z_p)
old_local_features (\hat{z_p})
global_features (z_s)
old_global_features (\hat{z_s})
aggregated_global_features (z_g)
```

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        labels_min = torch.zeros(local_features.size(0)).to(self.device).long()
+
+        return self.ce_criterion(logits_min, labels_min), self.ce_criterion(logits_max, labels_max)
+
+    def compute_loss(self, preds: Dict[str, torch.Tensor], target: torch.Tensor) -> Losses:
+        if self.old_global_module is None:
+            return super().compute_loss(preds, target)
+        loss = self.criterion(preds["prediction"], target)
+
+        # Optimal cos_sim_loss_weight is 100.0
```
emersodb, line 179: Maybe mention where this optimal value is derived from? I imagine the optimal value will differ depending on the downstream application. Is this a good default?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+                local_features=preds["local_features"],
+                old_local_features=preds["old_local_features"],
+                global_features=preds["global_features"],
+            )
+            total_loss = loss + self.contrastive_loss_weight * contrastive_loss
+            losses = Losses(
+                checkpoint=loss, backward=total_loss, additional_losses={"contrastive_loss": contrastive_loss}
+            )
+
+        # Optimal perFCL_loss_weight is [10.0, 10.0]
```
emersodb, line 200: Maybe mention where this optimal value is derived from? I imagine the optimal value will differ depending on the downstream application. If these are the suggest "default" values from the paper, maybe state that.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        # Optimal cos_sim_loss_weight is 100.0
+        if self.cos_sim_loss_weight:
+            cos_loss = self.get_cosine_similarity_loss(
+                local_features=preds["local_features"],
+                global_features=preds["global_features"],
+            )
+            total_loss = loss + self.cos_sim_loss_weight * cos_loss
+            losses = Losses(checkpoint=loss, backward=total_loss, additional_losses={"cos_sim_loss": cos_loss})
+
+        # Optimal contrastive_loss_weight is 10.0
```
emersodb, line 188: Maybe mention where this optimal value is derived from? I imagine the optimal value will differ depending on the downstream application.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,229 @@
...
+        assert len(local_features) == len(aggregated_global_features)
+        nega = self.cos_sim(local_features, aggregated_global_features)
+        logits_min = torch.cat((logits_min, nega.reshape(-1, 1)), dim=1)
+        logits_min /= self.temperature
+        labels_min = torch.zeros(local_features.size(0)).to(self.device).long()
+
+        return self.ce_criterion(logits_min, labels_min), self.ce_criterion(logits_max, labels_max)
+
+    def compute_loss(self, preds: Dict[str, torch.Tensor], target: torch.Tensor) -> Losses:
+        if self.old_global_module is None:
```
emersodb, line 175: We have a "catch all" fallback in the form of the else statement at the end of this function. It seems like if this module is defined and we still get down to the else statement below, something has been "weirdly" configured and we should at least warn the user about it? Alternatively, we could just eliminate this else and let the catch all happen if all the weight values are `None`

fl4health/clients/basic_client.py
```diff
@@ -308,7 +309,7 @@ def setup_client(self, config: Config) -> None:
 
         self.set_optimizer(config)
         self.learning_rate = self.optimizer.defaults["lr"]
-        self.criterion = self.get_criterion(config)
+        self.criterion = self.get_criterion(config).to(self.device)
```
emersodb, line 312: 🎉

fl4health/clients/moon_client.py
```diff
@@ -89,23 +84,18 @@ def set_parameters(self, parameters: NDArrays, config: Config) -> None:
...
 
         # Save the parameters of the global model
-        self.global_model = copy.deepcopy(self.model)
-        for param in self.global_model.parameters():
-            param.requires_grad = False
-        self.global_model.eval()
-        return output
+        self.global_model = self.clone_and_freeze_model(self.model)
+
+        return
```
emersodb, line 98: return isn't necessarily here for a none function 🙂

fl4health/model_bases/moon_base.py
```diff
@@ -14,10 +14,11 @@ def __init__(
         self.head_module = head_module
 
     def forward(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
+        # input is expected to be of shape (batch_size, *)
         x = self.base_module.forward(input)
         if self.projection_module:
```
emersodb, line 19: I know you didn't change this code, but wanted to just suggest a "pythonic" collapsing of the if-else. That is
```
p = self.projection_module.forward(x) if self.projection_module else x
```

research/flamby/fed_heart_disease/fenda/server.py
```diff
@@ -20,7 +20,7 @@
 )
 
 
-def main(config: Dict[str, Any], server_address: str) -> None:
+def main(config: Dict[str, Any], server_address: str, run_name: str) -> None:
```
emersodb, line 23: It doesn't look like this run name is used anywhere? Other servers use run_name to save a server-side checkpoint, but FENDA doesn't have "server-side" models. So it hasn't needed this value in the past.

research/flamby/fed_isic2019/fedavg/server.py
```diff
@@ -36,6 +37,28 @@ def main(config: Dict[str, Any], server_address: str, checkpoint_stub: str, run_
     client_manager = SimpleClientManager()
     model = Baseline()
 
+    log(INFO, f"if pretrain: {pretrain}")
+    if pretrain:
```
emersodb, line 41: We want to avoid these hard-coded values in our code because someone running this outside of our cluster will have issues. My suggestion would be rather than having a `pretrain: bool` as an argument. You rather have a `pretrain_path: Optional[Path] = None`. If there is a provided path, attempt to load a checkpoint from the provided path.

research/flamby/fed_isic2019/fedavg/server.py
```diff
@@ -36,6 +37,28 @@ def main(config: Dict[str, Any], server_address: str, checkpoint_stub: str, run_
...
+        dir = (
+            "/ssd003/projects/aieng/public/FL_env/models/fed_isic2019/fedavg/hp_sweep_results/lr_0.001/"
+            + run_name
+            + "/server_best_model.pkl"
+        )
+        fedavg_model_state = torch.load(dir).state_dict()
```
emersodb, line 47: I would suggest encapsulating this load into a method.

research/flamby/fed_isic2019/fedavg/server.py
```diff
@@ -36,6 +37,28 @@ def main(config: Dict[str, Any], server_address: str, checkpoint_stub: str, run_
...
+        model_state = model.state_dict()
+        matching_state = {}
+        for k, v in fedavg_model_state.items():
+            if k in model_state:
+                if v.size() == model_state[k].size():
+                    matching_state[k] = v
+                elif model_state[k].size()[1:] == v.size()[1:]:
+                    repeat = model_state[k].size()[0] // v.size()[0]
+                    original_size = tuple([1] * (len(model_state[k].size()) - 1))
+                    matching_state[k] = v.repeat((repeat,) + original_size)
```
emersodb, line 57: I'm a little confused as to what is happening here? A comment or set of comments would be useful to help illustrate what these calculations are doing.

research/flamby/fed_isic2019/fenda/client.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
         metric_meter_type: MetricMeterType = MetricMeterType.ACCUMULATION,
         checkpointer: Optional[TorchCheckpointer] = None,
+        type_run: str = "cos_sim",
```
emersodb, line 36: I'd say that we don't want the default to be using a "cos_sim" loss, but rather the vanilla FENDA approach. That is, no auxiliary losses attached.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
```diff
@@ -117,6 +117,7 @@ do
     nohup python -m research.flamby.fed_isic2019.fenda.server \
         --config_path ${SERVER_CONFIG_PATH} \
         --server_address ${SERVER_ADDRESS} \
+        --run_name ${RUN_NAME} \
```
emersodb, line 120: I don't think these are needed for FENDA, but I may be wrong.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
```diff
@@ -138,6 +139,7 @@ do
             --client_number ${c} \
             --learning_rate ${CLIENT_LR} \
             --server_address ${SERVER_ADDRESS} \
+            --type_run cos_sim \
```
emersodb, line 142: Does batch properly interpret `cos_sim` as a string here without the quotes? I feel like maybe it doesn't?

research/flamby/fed_ixi/fenda/client.py
```diff
@@ -33,7 +33,9 @@ def __init__(
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
         metric_meter_type: MetricMeterType = MetricMeterType.ACCUMULATION,
         checkpointer: Optional[TorchCheckpointer] = None,
+        type_run: str = "cos_sim",
```
emersodb, line 36: Same comment about defaulting to vanilla fenda rather than an auxiliary loss one.

research/flamby/utils.py
```diff
@@ -99,6 +99,8 @@ def evaluate_model_on_dataset(
                 preds = model(input, personal=True)["personal"]
```
emersodb, line 99: I actually think this is going to throw an error now and needs to be updated, because apfl models don't take a `personal` argument in the model forward. I think this can just be
```
preds = model(input)["personal"]
```

research/flamby/utils.py
```diff
@@ -99,6 +99,8 @@ def evaluate_model_on_dataset(
...
             else:
                 preds = model(input)
+                if isinstance(preds, tuple):
```
emersodb, line 102: I don't think `preds` should be coming back as a tuple, but maybe I'm missing something. It should always be a dictionary now?

fl4health/clients/fenda_client.py
sanaAyrml: Yeah honestly I just put it here based on my observations in fedisic, but I am not sure about other datasets.

fl4health/clients/fenda_client.py
sanaAyrml: Again it is based on my own observation.

research/flamby/fed_isic2019/fedavg/server.py
sanaAyrml: oh these things needed to be deleted, I am shocked they are here

research/flamby/fed_isic2019/fedavg/server.py
sanaAyrml: I will delete this from this pr and put it up on separate pr. This was not supposed to be here sorry.

research/flamby/fed_isic2019/fedavg/server.py
sanaAyrml: Again I will delete them for now. But for clarification, basic idea was because we were loading fedavg and model was not fenda model I needed to do such a checking and duplications in weights to not get error in loading from state dict.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
sanaAyrml: If we want to run with auxiliary task, that's where we define it. But for vanilla one we can omit it.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
sanaAyrml: In my runs there were no issues, as I was checking whether they were running with true losses or not.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        assert len(features) == len(positive_pairs)
+        posi = self.cos_sim(features, positive_pairs)
+        logits = posi.reshape(-1, 1)
+        assert len(features) == len(negative_pairs)
+        nega = self.cos_sim(features, negative_pairs)
+        logits = torch.cat((logits, nega.reshape(-1, 1)), dim=1)
+        logits /= self.temperature
+        labels = torch.zeros(features.size(0)).to(self.device).long()
+
+        return self.ce_criterion(logits, labels)
```
sanaAyrml, line 135: I have put compute_contrastive_loss function in here, but we can also move it to any other file specific for new losses.

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+
+        preds = self.model(input)
+
+        if self.contrastive_loss_weight or self.perfcl_loss_weights:
+            if self.old_local_module is not None:
+                preds["old_local_features"] = self.old_local_module.forward(input).reshape(
+                    len(preds["local_features"]), -1
+                )
+                if self.perfcl_loss_weights:
+                    if self.old_global_module is not None:
```
emersodb, line 80: If these modules aren't defined, the PerFCL loss isn't going to be calculable, at least in a correct way right? Rather than doing an if here, where we'll fail somewhere else (or worse get through with a bad calculation), I'd suggest we do
```
assert self.old_global_module is not None and self.aggregated_global_module is not None
```

fl4health/clients/fenda_client.py
emersodb: Let's have it here for now. I'll put a ticket in the backlog for us to maybe have a "losses" folder or something for us to place our custom losses in!

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+    def get_perfcl_loss(
+        self,
+        local_features: torch.Tensor,
+        old_local_features: torch.Tensor,
+        global_features: torch.Tensor,
+        old_global_features: torch.Tensor,
+        aggregated_global_features: torch.Tensor,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Perfcl loss consists of two contrastive loss computation.
```
emersodb, line 159: Do you mind putting a link to the arXiv paper in this comment to draw a line from the variable notation to the paper itself?

fl4health/clients/fenda_client.py
```diff
@@ -1,12 +1,226 @@
...
+        additional_losses = {}
+
+        # Optimal cos_sim_loss_weight for FedIsic dataset is 100.0
+        if self.cos_sim_loss_weight:
+            cos_sim_loss = self.get_cosine_similarity_loss(
+                local_features=preds["local_features"],
+                global_features=preds["global_features"],
+            )
+            total_loss += self.cos_sim_loss_weight * cos_sim_loss
+            additional_losses["cos_sime_loss"] = cos_sim_loss
```
emersodb, line 196: I'm not 100% sure, but I imagine that the key `"cos_sime_loss"` might cause some issues? Either way I think the typo should probably be fixed.

fl4health/clients/moon_client.py
emersodb: This was marked resolved, but it looks like you didn't remove the return here.

research/flamby/fed_isic2019/fedavg/server.py
```diff
@@ -93,6 +93,11 @@ def main(config: Dict[str, Any], server_address: str, checkpoint_stub: str, run_
         help="Server Address to be used to communicate with the clients",
         default="0.0.0.0:8080",
     )
+    parser.add_argument(
```
emersodb, line 96: This doesn't appear to be used anywhere? If that's true, it should be removed.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
emersodb: Maybe I'm misunderstanding, but the fenda server doesn't appear to use this run name for anything. That is, it takes the argument, but doesn't use it.

research/flamby/fed_heart_disease/fenda/run_fold_experiment.slrm
```diff
@@ -117,6 +117,7 @@ do
     nohup python -m research.flamby.fed_heart_disease.fenda.server \
         --config_path ${SERVER_CONFIG_PATH} \
         --server_address ${SERVER_ADDRESS} \
+        --run_name ${RUN_NAME} \
```
emersodb, line 120: This doesn't appear to be used by the FENDA server in any way.

research/flamby/fed_isic2019/fenda/server.py
```diff
@@ -77,6 +77,12 @@ def main(config: Dict[str, Any], server_address: str) -> None:
         help="Server Address to be used to communicate with the clients",
         default="0.0.0.0:8080",
     )
+    parser.add_argument(
```
emersodb, line 80: This argument doesn't appear to be used by the FENDA server anywhere.

research/flamby/fed_heart_disease/fenda/run_fold_experiment.slrm
emersodb: It also looks like the fed_heart_disease FENDA server doesn't parse this argument at all, which might throw an error.

tests/clients/test_fenda_client.py
```diff
@@ -0,0 +1,57 @@
...
+    params_local = [
+        copy.deepcopy(val.cpu().numpy()) for _, val in fenda_client.model.local_module.state_dict().items()
+    ]
+    params_global = [
+        copy.deepcopy(val.cpu().numpy()) for _, val in fenda_client.model.global_module.state_dict().items()
+    ]
+    fenda_client.get_parameters(config)
+
+    assert isinstance(fenda_client.old_local_module, torch.nn.Module)
+    assert isinstance(fenda_client.old_global_module, torch.nn.Module)
```
emersodb, line 29: One other check that might be useful here is to assert that `requires_grad` is false for each of the parameter components of these modules, just to be sure no gradient is being calculated for them.

tests/clients/test_fenda_client.py
```diff
@@ -0,0 +1,57 @@
...
+        assert (params_local[i] == old_local_module_params[i]).all()
+    for i in range(len(params_global)):
+        assert (params_global[i] == old_global_module_params[i]).all()
+
+
+@pytest.mark.parametrize("local_module,global_module,head_module", [(FeatureCnn(), FeatureCnn(), FendaHeadCnn())])
+def test_computting_contrastive_loss(get_fenda_client: FendaClient) -> None:  # noqa
+    torch.manual_seed(42)
+    fenda_client = get_fenda_client
+    fenda_client.temperature = 1
```
emersodb, line 43: I would suggest testing the constrastive loss with a non-unit temperatures to ensure that the role that temperature plays happens properly. Using 1 might collapse parts of the calculation that we want to test.

emersodb (commented):
Overall, I think all of the new changes you made look really good. Just a few additional small comments and then I think we should be good to go

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
sanaAyrml: I think it does use it in checkpointing and saving models under that subfolder.

fl4health/clients/fenda_client.py
sanaAyrml: I don't think assertion here would be a wise choice as in the first round of training we won't have old module, but it is fine to compute one round without contrastive loss.

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
emersodb: Because FENDA is a personal model, it actually uses the `PersonalServer` class. These models don't save a "server-side" model. So only the client mains receive a `run_name` argument. Here you're passing it to the server main, which actually doesn't need it.

fl4health/clients/fenda_client.py
emersodb: Ah good point. Keep it as is then!

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
sanaAyrml: Yeah I figured it out later and omitted it. I have added these run_name for loading from fedavg because of experiments.

emersodb (approved):
Looks good to go. Great addition Sana.

PR #69, opened 2023-11-04
Metric refactor

jewelltaylor:
# PR Type
[**Feature** | Fix | Documentation | Other() ]

# Short Description
Clickup Story: [Refactor Client Metrics](https://app.clickup.com/t/86860zbdk)

Refactor client metrics so that they maintain state, eliminating the need for MetricMeter's which tend to just make things more complicated. Also add the option to return features from models through predict method to be used in loss computation. Then apply this to the MOON client to compute contrastive loss. Added some documentation with proper  formatting to Metric related code and a few client methods that I altered slightly in this PR. In a follow up PR, I will more extensively apply the proper formatting to client code and more broadly. 

**Note:** This should only be reviewed once the [Create Fixed Requirements File for FLamby, Update Dynamic Weight Exchanger and FedOpt Example](https://github.com/VectorInstitute/FL4Health/pull/68) PR is merged. I just wanted to base my PR on that branch to adapt David's CustomMetricMeter (and a few other relevant parts) to the simplified Metric tracking.

# Tests Added
- Adapted Metric related tests that were affected in the refactor 
- Add a test for the updated MetricMeter


jewelltaylor:
Now that PR #68 is approved and will presumably be merged into main with minimal changes, I think this PR is ready to review. I had to make some additional changes to update the parts of the code affected by Sana's recent PR #66. I would appreciate feedback that anyone has!

examples/fedopt_example/metrics.py
```diff
@@ -75,20 +75,26 @@ def compute_metrics(self) -> Metrics:
...
         """
+        super().__init__(name)
         self.true_preds = 0
         self.total_preds = 0
+        self.classes: List[str]
+        self.label_to_class: Dict[int, str]
+        self.n_classes: int
+        self.outcome_dict: Dict[str, Outcome]
+
+    def _setup(self, label_encoder: LabelEncoder) -> None:
```
emersodb, line 97: Typically the `_` prefix is reserved for protected or private methods. That is, methods that are exclusively called within the class itself, rather than externally. All that is to say, I would recommend dropping the _ based on the way this is being used 🙂

examples/fedopt_example/metrics.py
```diff
@@ -75,20 +75,26 @@ def compute_metrics(self) -> Metrics:
...
         """
         This class is used to compute metrics associated with the AG's News task. There are a number of classes and
         we want to accumulate a bunch of statistics all at once to facilitate the computation of a number of different
-        metrics for this problem. As such, we define our own MetricMeter and bypass the standard metric meter
-        implementations, which calculate separate metrics individually.
+        metrics for this problem. As such, we define our own Metric class and bypass the standard SimpleMetric class,
+        which calculate separate metrics individually.
 
         Args:
-            label_encoder (LabelEncoder): This class is used to determine the mapping of integers to label names for
```
emersodb, line 87: Maybe transfer this comment about `label_encoder` to the setup method below?

examples/fedopt_example/client.py
```diff
@@ -99,5 +86,5 @@ def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
     # Load model and data
     DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
     data_path = Path(args.dataset_path)
-    client = NewsClassifierClient(data_path, DEVICE)
+    client = NewsClassifierClient(data_path, [CompoundMetric("")], DEVICE)
```
emersodb, line 89: Should we give this metric a name?

fl4health/clients/basic_client.py
```diff
@@ -53,15 +50,8 @@ def __init__(
...
-        # Define mapping from prediction key to meter to pass to MetricMeterManager constructor for train and val
-        train_key_to_meter_map = {
-            "prediction": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "train_meter")
-        }
-        self.train_metric_meter_mngr = MetricMeterManager(train_key_to_meter_map)
-        val_key_to_meter_map = {
-            "prediction": MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "val_meter")
-        }
-        self.val_metric_meter_mngr = MetricMeterManager(val_key_to_meter_map)
+        self.train_metric_meter_mngr = MetricManager(metrics=self.metrics, metric_mngr_name="train")
```
emersodb, line 53: Since these are now "MetricManager" objects, should we change the names from `train_metric_meter_mngr` to `train_metric_manager ` and `val_metric_meter_mngr ` to `val_metric_manager` throughout?

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+    def update(self, preds: Dict[str, torch.Tensor], target: torch.Tensor) -> None:
+        """
+        Updates (or creates then updates) a list of metrics for each prediction type.
 
-class MetricMeterManager:
-    """
-    Class to manage one or metric meters.
-    """
+        Args:
+            preds (Dict[str, torch.Tensor]): A dictionairy of
```
emersodb, line 281: This comment is incomplete. Also, I think there are a few places throughout the code where "dictionary" is spelled incorrectly in this way 😂

fl4health/clients/basic_client.py
```diff
@@ -332,30 +322,40 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
-        User can override for more complex logic.
+        Computes the prediction(s) (and potentially features) of the model(s) given the input.
+
+        Args:
+            input (torch.Tensor): Inputs to be fed into the model.
+
+        Returns:
+            Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
+            contains predictions indexed by name and the second element contains intermediate activations
+            index by name. BY passing features, we can compute losses such as the model contrasting loss in MOON.
```
emersodb, line 335: I think you mean to capitalize BY, but just want to check

fl4health/clients/basic_client.py
```diff
@@ -332,30 +322,40 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
         return FullParameterExchanger()
 
-    def predict(self, input: torch.Tensor) -> Dict[str, torch.Tensor]:
+    def predict(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]:
         """
-        Return dict of str and torch.Tensor containing predictions when given input.
-        In the default case, the dict has a single item with key prediction.
-        In more complicated approaches such as APFL, the dict has as many items as prediction types
-        User can override for more complex logic.
+        Computes the prediction(s) (and potentially features) of the model(s) given the input.
```
emersodb, line 327: Super minor, but I think you can drop the parentheses here and just write 
```
Computes the prediction(s), and optionally features, of the model(s) given the input.
```

fl4health/clients/evaluate_client.py
```diff
@@ -42,9 +41,9 @@ def __init__(
         self.data_loader: DataLoader
         self.criterion: _Loss
         self.global_loss_meter = LossMeter.get_meter_by_type(loss_meter_type)
-        self.global_metric_meter = MetricMeter.get_meter_by_type(self.metrics, metric_meter_type, "global_eval_meter")
+        self.global_metric_meter = MetricManager(self.metrics, "global_eval_meter")
```
emersodb, line 44: Maybe change the names of these properties to be something like `global_metric_manager` and `local_metric_manager`?

fl4health/clients/fed_prox_client.py
```diff
@@ -96,7 +94,21 @@ def set_parameters(self, parameters: NDArrays, config: Config) -> None:
             initial_layer_weights.detach().clone() for initial_layer_weights in self.model.parameters()
         ]
 
-    def compute_loss(self, preds: Dict[str, torch.Tensor], target: torch.Tensor) -> Losses:
+    def compute_loss(
+        self, preds: Dict[str, torch.Tensor], features: Dict[str, torch.Tensor], target: torch.Tensor
+    ) -> Losses:
+        """
+        Computes loss given predictions of the model and ground truth data.
```
emersodb, line 101: Maybe add a description that we're also adding in the proximal loss, comparing the l2 norm between the initial and final weights of local training

fl4health/clients/fenda_client.py
```diff
@@ -180,7 +176,23 @@ def get_perfcl_loss(
...
+        set from server config.
+
+        Args:
+            preds (Dict[str, torch.Tensor]): Prediction(s) of the model(s) indexed by name.
+            features: (Dict[str, torch.Tensor]): Feature(s) of the model(s) indexed by name.
+            target: (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            Losses: Object containing checkpoint loss, backward loss and additional losses indexed by name.
+            Additional losses includes proximal loss.
```
emersodb, line 194: No proximal loss in these calculations 🙂

fl4health/clients/moon_client.py
```diff
@@ -48,16 +46,26 @@ def __init__(
...
+
+        Args:
+            input (torch.Tensor): Inputs to be fed into the model.
 
-        return preds
+        Returns:
+            Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
+            contains predictions indexed by name and the second element contains intermediate activations
+            index by name. Specificaly the features of the model, features of the global model and features of
+            the old model are passed.
```
emersodb, line 60: Super minor I would say are _returned_ rather than "passed"

fl4health/model_bases/fenda_base.py
```diff
@@ -43,12 +43,14 @@ def __init__(self, local_module: nn.Module, global_module: nn.Module, model_head
...
         local_output = self.local_module.forward(input)
         global_output = self.global_module.forward(input)
-        return {
-            "prediction": self.model_head.forward(local_output, global_output),
+        preds = {"prediction": self.model_head.forward(local_output, global_output)}
+        features = {
             "local_features": local_output.reshape(len(local_output), -1),
             "global_features": global_output.reshape(len(global_output), -1),
         }
+        # Return preds and features as seperate dictionairy as in moon base
```
emersodb, line 55: seperate -> separate 🙂

fl4health/utils/metrics.py
```diff
@@ -1,47 +1,158 @@
-from __future__ import annotations
```
emersodb, line 1: lol...where did this guy come from 😂

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        Updates (or creates then updates) a list of metrics for each prediction type.
 
-class MetricMeterManager:
-    """
-    Class to manage one or metric meters.
-    """
+        Args:
+            preds (Dict[str, torch.Tensor]): A dictionairy of
+        """
+        if len(self.metrics_per_prediction_type) == 0:
```
emersodb, line 283: super minor, but I think you can just do `if self.metrics_per_perdiction_type:`

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+            self.metrics_per_prediction_type = {key: copy.deepcopy(self.og_metrics) for key in preds.keys()}
 
-    def __init__(self, key_to_meter_map: Dict[str, MetricMeter]):
-        self.key_to_meter_map = key_to_meter_map
+        for pred, mtrcs in zip(preds.values(), self.metrics_per_prediction_type.values()):
```
emersodb, line 286: I'm a bit wary of doing the zip here, as it assumes that the keys of the dictionaries are ordered in the same way and are the same length and won't fail if that is not the case. Maybe we do something like
```
assert len(preds) == len(self.metrics_per_prediction_type)
for prediction_key, pred in prediction_keys.items():
    metrics_for_prediction_type = self.metrics_per_prediction_type[prediction_key]
    for metric_for_prediction_type in metrics_for_prediction_type:
        metric_for_prediction_type.update(pred, target)
```

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-    def compute(self) -> Dict[str, Scalar]:
+        Returns:
+            Metrics: Dictionairy containing computed metrics along with string identifiers
+                for each prediction type.
+        """
         all_results = {}
-        for meter in self.key_to_meter_map.values():
-            result = meter.compute()
-            all_results.update(result)
+        for metrics_key, mtrcs in self.metrics_per_prediction_type.items():
```
emersodb, line 299: Super minor, but let's avoid the abbreviation here, since `metrics` is pretty short anyway.

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-            f"{self.name}_{str(metric)}".lstrip("_"): avg for metric, avg in zip(self.metrics, metric_value_averages)
-        }
-
-        return results
-
-    def clear(self) -> None:
-        self.metric_values_history = [[] for _ in range(len(self.metrics))]
-        self.counts = []
+        self.og_metrics = metrics
+        self.metric_mngr_name = metric_mngr_name
```
emersodb, line 273: any objection to expanding these to `metric_manager_name ` and `metric_manager_name `, respectively since we're only saving three letters in the abbreviation anyway?

emersodb: There are a few other places in the code where we use this abbreviation that I'd suggest we expand as well, unless you really don't like it 🙂

emersodb (commented):
Overall, I think this is an awesome refactor. All the comments I left are quite minor and I didn't see anything major that was missed.

fl4health/utils/metrics.py
jewelltaylor: I believe this required for classes that have methods that return instances of class and you are using annotations. I think this was relevant when defining the metric meter based on meter type but no longer required.

SO post: https://stackoverflow.com/questions/61544854/from-future-import-annotations

fl4health/utils/metrics.py
```diff
@@ -133,140 +260,51 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-        results: Dict[str, Scalar] = {
-            f"{self.name}_{str(metric)}".lstrip("_"): avg for metric, avg in zip(self.metrics, metric_value_averages)
-        }
-
-        return results
-
-    def clear(self) -> None:
-        self.metric_values_history = [[] for _ in range(len(self.metrics))]
-        self.counts = []
+        self.og_metrics = metrics
```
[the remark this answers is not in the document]
jewelltaylor: hahah not quite, its just a short form I have used for original but I can change it to avoid that interpretation

fl4health/utils/metrics.py
emersodb: Ah I see. I completely forgot we did that

examples/fedopt_example/metrics.py
jewelltaylor: Yeah, good call! I initially was aiming for it to be internal but realized we have to call it externally so I forgot to change it back

fl4health/utils/metrics.py
jewelltaylor: I don't think this is the case for dictionaries, just verified in python interpreter

fl4health/utils/metrics.py
jewelltaylor: Yeah I agree this is a better way to go about it. Just one small issue, the length of the metric_per_prediction_type is going to as long as as the number of prediction types not the as the number of metrics. Thus I instead assert that the list of metrics at a given key is long as preds.

fl4health/utils/metrics.py
emersodb: huh...I thought I verified it in an interpreter too 😂 
```
>>> dict_empty = {}
>>> dict_filled = {"a": "b"}
>>> print("HI") if dict_empty else print("BYE")
BYE
>>> print("HI") if dict_filled else print("BYE")
HI
```

emersodb: I'm fine with leaving it though. It's very minor

jewelltaylor: Sorry its cause `if self. metrics_per_prediction_type:` should be `if not self. metrics_per_prediction_type:`. My bad hahaha

emersodb: Right...also my bad for not writing the correct condition 🤦

fl4health/clients/basic_client.py
```diff
@@ -332,30 +322,40 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
+
+        Args:
+            input (torch.Tensor): Inputs to be fed into the model.
+
+        Returns:
+            Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
+            contains predictions indexed by name and the second element contains intermediate activations
```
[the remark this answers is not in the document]
emersodb: I see you added a comment in the compute loss function about anything stored in preds being used to compute metrics, but I think we should also put that in the comment for the `predict` function since that's where people will define what's in the predictions dictionary. So they "know" this before potentially stuffing extra things into that dictionary.

fl4health/clients/fed_prox_client.py
```diff
@@ -98,7 +98,8 @@ def compute_loss(
         self, preds: Dict[str, torch.Tensor], features: Dict[str, torch.Tensor], target: torch.Tensor
     ) -> Losses:
         """
-        Computes loss given predictions of the model and ground truth data.
+        Computes loss given predictions of the model and ground truth data. Adds to objective by including
+        proximal loss which is the L2 norm between the initial and final weights of local training.
```
emersodb, line 102: I'm going to be a pedantic mathematician and have you lower-case l2 here. L^2/L_2 norms operate on functions not vectors lol

emersodb (approved):
Changes look good to me. Left two additional small comments. Feel free to take them or leave them

fl4health/clients/basic_client.py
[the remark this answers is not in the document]
jewelltaylor: Good call!

fl4health/clients/fed_prox_client.py
jewelltaylor: hahahah don't ever not be a pedantic mathematician, its good to know these notation conventions

PR #70, opened 2023-11-09
Add warmup module

sanaAyrml:
# PR Type
[Feature]

# Short Description

This is a preview PR that I just put up to get feedback on the way I added the warm-up module.
In this PR, I added a warm_up_base to models along with a maybe_init_model function. To enable FedAvg warm-up, model bases should inherit warm_up_base. Additionally, they can override the maybe_init_model function if they wish to modify the loading method. I have also updated apfl_base and fenda_base based on warm_up_base, modifying them to only train the global part during retraining.

Furthermore, I added fedavg_warm_up folders to SCAFFOLD, Fedprox, Fenda, Moon, and APFL experiments. To enable FedAvg warm-up, users should first run the server and client in the fedavg_warm_up folders and then in the original ones. I also set the seed on both the client and server sides to ensure consistent client data for these two runs.

# Tests Added
None`


examples/apfl_example/client.py
```diff
@@ -40,11 +42,29 @@ def get_criterion(self, config: Config) -> _Loss:
...
+        type=str,
+        help="Server Address for the clients to communicate with the server through",
+        default="0.0.0.0:8080",
+    )
+    parser.add_argument(
+        "--seed",
+        action="store",
+        type=int,
+        help="Seed for the random number generator",
+        required=True,
```
emersodb, line 57: Can we make this an optional parameter rather than required throughout? We have it as an optional argument for the clients so I think that's reasonable.

emersodb: I'm also good with using a default parameter if we would rather.

emersodb: I'm also good with using a default parameter if we would rather.

emersodb:
Apologies, because I imagine it was a lot of work, but I don't think we need every example to have a warmup option. The goal of the examples is to demonstrate different workflows that someone might want to do. So I think injecting warm up into 1-2 different examples makes sense, especially for more difficult settings (like APFL or Moon), but having it everywhere might be confusing for someone just trying to learn the basics. Do you think we could only have warm up for two of the example implementations?

examples/apfl_example/config.yaml
```diff
@@ -1,5 +1,6 @@
 # Parameters that describe server
 n_server_rounds: 25 # The number of rounds to run FL
+n_warm_up_rounds: 5 # The number of rounds to run before starting APFL
```
emersodb, line 3: I would say that we don't want to mix the configs here across separate runs. That is, the warm up training should have its own configuration rather than sharing one.

fl4health/model_bases/apfl_base.py
```diff
@@ -1,19 +1,23 @@
...
+from typing import Dict, List, Optional
 
 import torch
 import torch.nn as nn
 
+from fl4health.model_bases.warm_up_base import WarmUpModel
 
-class ApflModule(nn.Module):
+
+class ApflModule(WarmUpModel):
```
emersodb, line 10: The way you did this is quite interesting, but I think it might overloading the idea of an APFL module (and other modules like FENDA), especially from the perspective of adaptive alpha calculations etc. The way it is being "loaded" as a warmed up model in the server for the APFL example above is also a bit odd, in the sense that it's not really an APFL model being loaded but a part of one. 

What I was thinking is that we would have a "warm up" loading class that takes in a **source nn.Module**, a **checkpoint** corresponding to a trained version of that module, a **target nn.Module** and a **mapping** that maps the state_dict components of the source module into components of the target module. The class would then perform the transfer and return a "warmed up" model. We could have pre-defined defaults for the various models like APFL/FENDA making assumptions, but the vanilla version would essentially be capable of following any mapping.

For the APFL example, for instance, we'd have a module that simply corresponds to the single model that is duplicated in an APFL module for the source, the target would be an APFL model where we load parameters for `layer_name` to `global_model.layer_name` in the APFL target.

examples/apfl_example/client.py
sanaAyrml: Actually, I am not sure which way to go is better. So if they want to use warm up and first run them, it is good to have it as required so they wouldn't forget to fix it. On the other hand I agree that it is not necessary if they run a vanilla version so is should not be required. Maybe I can have assertion in beginning as if they provide warm_up_dir they should set a seed?

sanaAyrml:
Sure I also agree that it is a bit so much to have it every where, regardless I needed to run to make sure everything works well. In my opinion we can keep it for SCAFFOLD (to show that strategy should be different), and one of models that have base models and also it has personalized module, such as APFL or Fenda.

fl4health/model_bases/apfl_base.py
sanaAyrml: That is doable for APFL as they have separate models for personalized section and global one, but I am not sure of same thing in fenda, due to classifier layer. I mean if we train only global module during warmup the dimension of first layer classifier would be half of what it would be in original model and it is a bit tricky how we would load models. Also regardless of that, if we do like that a person needs to define a model for warmup and a model for original training which is a bit more confusing that what do we have now. 
Though I am open to both ways, I just initially thought maybe this way has better interfaca for user.

jewelltaylor: I agree with David that having a warm up loading class is probably the cleanest way to incorporate the warm up functionality. It only requires establishing the loading class and some minimal interventions in the constructor of the BasicClient. I do agree that in some cases it may require a bit more code from the user but in most cases we can use sensible defaults (ie Load all the parameters of the source model into the target model). 

For the case of FENDA, I think you raise a solid point so I talked with David. He mentioned you could just train a global model in the warmup phase and in the subsequent phase discard the classifier and set the global and feature extractor with the remaining parameters. 

I think David is going to speak with you about this further in your meeting, so he can further clarify if I am missing anything.

examples/apfl_example/client.py
sanaAyrml: With the new setting of experiments I think this can be the best approach, previously I wanted to keep the setting of experiment in the paper.

fl4health/model_bases/apfl_base.py
sanaAyrml: That was the original approach I had for my experimentations, thanks for clarifying. I can do that.

PR #72, opened 2023-11-22
FedPer Implementation

emersodb:
# PR Type
**Feature**

# Short Description

Clickup Ticket: https://app.clickup.com/t/8686ckn31

This PR adds in the FedPer method into the repository. The addition is fairly straightforward within our infrastructure. It's essentially just a globally trained feature extractor with a locally trained classification head in each client. I added an example using the infrastructure while also training it with MOON. So the example client **inherits** from the MOON client to apply the auxiliary losses to the model along with local personalization, which I thought was nice.

In the course of adding the method, I noticed a few small bugs in MOON and the auxiliary losses (like PerFCL) in the FENDA approaches. These have been fixed and tests have been added to ensure that they are indeed fixed.

# Tests Added

Added losses associated with MOON's contrastive loss calculations and the loss computations of MOON, FENDA, FedProx, and FedPer

### DIFF
--- a/examples/fedper_example/README.md
+++ b/examples/fedper_example/README.md
@@ -1,5 +1,5 @@
-# FENDA Federated Learning Example
-This example provides an example of training a FENDA type model on a non-IID subset of the MNIST data. The FL server
+# FedPer Federated Learning Example
+This example provides an example of training a FedPer type model on a non-IID subset of the MNIST data. The FL server
 expects three clients to be spun up (i.e. it will wait until three clients report in before starting training). Each client
 has a modified version of the MNIST dataset. This modification essentially subsamples a certain number from the original
 training and validation sets of MNIST in order to synthetically induce local variations in the statistical properties
@@ -21,7 +21,7 @@ to install all of the dependencies for this project.
 
 The next step is to start the server by running
 ```
-python -m examples.fenda_example.server  --config_path /path/to/config.yaml
+python -m examples.fedper_example.server  --config_path /path/to/config.yaml
 ```
 from the FL4Health directory. The following arguments must be present in the specified config file:
 * `n_clients`: number of clients the server waits for in order to run the FL training
@@ -35,7 +35,7 @@ from the FL4Health directory. The following arguments must be present in the spe
 Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the three
 clients. This is done by simply running (remembering to activate your environment)
 ```
-python -m examples.fenda_example.client --dataset_path /path/to/data --minority_numbers <sequence of numbers>
+python -m examples.fedper_example.client --dataset_path /path/to/data --minority_numbers <sequence of numbers>
 ```
 **NOTE**: The argument `dataset_path` has two functions, depending on whether the dataset exists locally or not. If
 the dataset already exists at the path specified, it will be loaded from there. Otherwise, the dataset will be
--- a/examples/fedper_example/README.md
+++ b/examples/fedper_example/README.md
@@ -1,0 +1,48 @@
+# FENDA Federated Learning Example
+This example provides an example of training a FENDA type model on a non-IID subset of the MNIST data. The FL server
+expects three clients to be spun up (i.e. it will wait until three clients report in before starting training). Each client
+has a modified version of the MNIST dataset. This modification essentially subsamples a certain number from the original
+training and validation sets of MNIST in order to synthetically induce local variations in the statistical properties
+of the clients training/validation data. In theory, the models should be able to perform well on their local data
+while learning from other clients data that has different statistical properties. The subsampling is specified by
+sending a list of integers between 0-9 to the clients when they are run with the argument `--minority_numbers`.
+
+The server has some custom metrics aggregation and uses Federated Averaging as its server-side optimization. The implementation uses a special type of weight exchange based on named-layer identification.
+
+## Running the Example
+In order to run the example, first ensure you have the virtual env of your choice activated and run
+```
+pip install --upgrade pip
+pip install -r requirements.txt
+```
+to install all of the dependencies for this project.
+
+## Starting Server
+
+The next step is to start the server by running
+```
+python -m examples.fenda_example.server  --config_path /path/to/config.yaml
+```
+from the FL4Health directory. The following arguments must be present in the specified config file:
+* `n_clients`: number of clients the server waits for in order to run the FL training
+* `local_epochs`: number of epochs each client will train for locally
+* `batch_size`: size of the batches each client will train on
+* `n_server_rounds`: The number of rounds to run FL
+* `downsampling_ratio`: The amount of downsampling to perform for minority digits
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the three
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.fenda_example.client --dataset_path /path/to/data --minority_numbers <sequence of numbers>
+```
+**NOTE**: The argument `dataset_path` has two functions, depending on whether the dataset exists locally or not. If
+the dataset already exists at the path specified, it will be loaded from there. Otherwise, the dataset will be
+automatically downloaded to the path specified and used in the run.
+
+The argument `minority_numbers` specifies which digits (0-9) in the MNIST dataset the client will subsample to
+simulate non-IID data between clients. For example `--minority_numbers 1 2 3 4 5` will ensure that the client
+downsamples these digits (using the `downsampling_ratio` specified to the config).
+
+After both clients have been started federated learning should commence.
--- a/examples/fedper_example/client.py
+++ b/examples/fedper_example/client.py
@@ -28,6 +28,8 @@ class MnistFedPerClient(MoonClient):
         device: torch.device,
         minority_numbers: Set[int],
     ) -> None:
+        # We inherit from a MOON client here intentionally to be able to use auxiliary losses associated with the
+        # global module's feature space in addition to the personalized architecture of FedPer.
         super().__init__(data_path=data_path, metrics=metrics, device=device)
         self.minority_numbers = minority_numbers
 
@@ -39,6 +41,8 @@ class MnistFedPerClient(MoonClient):
         return train_loader, val_loader
 
     def get_model(self, config: Config) -> nn.Module:
+        # NOTE: Flatten features is set to true to make the model compatible with the MOON contrastive loss function,
+        # which requires the intermediate feature representations to be flattened for similarity calculations.
         model: nn.Module = FedPerModel(
             global_feature_extractor=FedPerGloalFeatureExtractor(),
             local_prediction_head=FedPerLocalPredictionHead(),
--- a/examples/fedper_example/client.py
+++ b/examples/fedper_example/client.py
@@ -1,0 +1,73 @@
+import argparse
+from pathlib import Path
+from typing import Sequence, Set, Tuple
+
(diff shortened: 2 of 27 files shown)


examples/fedper_example/client.py
```diff
@@ -0,0 +1,73 @@
...
+from fl4health.clients.moon_client import MoonClient
+from fl4health.model_bases.fedper_base import FedPerModel
+from fl4health.parameter_exchange.layer_exchanger import FixedLayerExchanger
+from fl4health.parameter_exchange.parameter_exchanger_base import ParameterExchanger
+from fl4health.utils.load_data import load_mnist_data
+from fl4health.utils.metrics import Accuracy, Metric
+from fl4health.utils.sampler import MinorityLabelBasedSampler
+
+
+class MnistFedPerClient(MoonClient):
```
emersodb, line 23: **Note**: We inherit from a MOON client here intentionally to be able to use auxiliary losses associated with the global module's feature space in addition to the personalized architecture.

examples/fenda_example/client.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         device: torch.device,
         minority_numbers: Set[int],
     ) -> None:
-        super().__init__(data_path=data_path, metrics=metrics, device=device)
+        super().__init__(data_path=data_path, metrics=metrics, device=device, perfcl_loss_weights=(1.0, 1.0))
```
emersodb, line 29: Adding in the perfcl loss to the FENDA example. This is just for testing when running the example to make sure nothing is broken there.

fl4health/clients/fenda_client.py
```diff
@@ -207,7 +205,7 @@ def compute_loss(
         """
 
         loss = self.criterion(preds["prediction"], target)
-        total_loss = loss
+        total_loss = loss.clone()
```
emersodb, line 208: Without clone, `total_loss` and `loss` share memory. This means that anything that is added to `total_loss` below is also added to loss. This. means that checkpoint and backward in the loss object end up being identical, which we don't want. I added a unit test to make sure the clone here fixes the issue.

fl4health/clients/moon_client.py
```diff
@@ -60,11 +59,12 @@ def predict(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[st
...
-        old_features = torch.zeros(self.len_old_models_buffer, *features.size()).to(self.device)
+        old_features = torch.zeros(self.len_old_models_buffer, *features["features"].size()).to(self.device)
         for i, old_model in enumerate(self.old_models_list):
-            old_features[i] = old_model(input)[1]
-        global_features = self.global_model(input)[1]
-        features.update({"global_features": global_features, "old_features": old_features})
+            _, old_model_features = old_model(input)
+            old_features[i] = old_model_features["features"]
+        _, global_model_features = self.global_model(input)
+        features.update({"global_features": global_model_features["features"], "old_features": old_features})
```
emersodb, line 67: With the refactor of the use of a features dictionary, this predict function ended up being broken. I believe these changes fix the issue. However, correct me if I'm wrong.

fl4health/clients/moon_client.py
```diff
@@ -89,8 +89,6 @@ def get_contrastive_loss(
         return self.ce_criterion(logits, labels)
 
     def set_parameters(self, parameters: NDArrays, config: Config) -> None:
-        assert isinstance(self.model, MoonModel)
```
emersodb, line 92: This is unnecessary, unless I'm mistaken. Removing it also allows for FedPer models to be used with MOON Clients, which is nice...

fl4health/model_bases/apfl_base.py
```diff
@@ -4,8 +4,10 @@
 import torch
 import torch.nn as nn
 
+from fl4health.model_bases.partial_layer_exchange_model import PartialLayerExchangeModel
 
-class ApflModule(nn.Module):
+
+class ApflModule(PartialLayerExchangeModel):
```
emersodb, line 10: Having models that exchange a subset of their weights inherit from an abstract based class that forces the implementation of the `layers_to_exchange` function. This is just a formalism for now, but should be useful in future iterations of the parameter exchanger mechanisms.e

fl4health/model_bases/apfl_base.py
```diff
@@ -69,5 +71,7 @@ def update_alpha(self) -> None:
         self.alpha = alpha
 
     def layers_to_exchange(self) -> List[str]:
-        layers_to_exchange: List[str] = [layer for layer in self.state_dict().keys() if "global_model" in layer]
+        layers_to_exchange: List[str] = [
+            layer for layer in self.state_dict().keys() if layer.startswith("global_model.")
```
emersodb, line 75: Changing this function to mirror FENDA and FedPer. In particular, we only want the layer name to start with `global_model.` rather than have it appear anywhere else in a name.

fl4health/model_bases/fedper_base.py
```diff
@@ -0,0 +1,45 @@
...
+        Args:
+            global_feature_extractor (nn.Module): First set of layers. These are exchanged with the server.
+            local_prediction_head (nn.Module): Final set of layers. These are not aggregated by the server.
+            flatten_features (bool): Whether or not the forward should flatten the produced features across the batch.
+                Defaults to False.
+        """
+        super().__init__()
+        self.global_feature_extractor = global_feature_extractor
+        self.local_prediction_head = local_prediction_head
+        self.flatten_features = flatten_features
```
emersodb, line 30: Flatten features is used to make the model compatible with MOON which requires the intermediate feature representations to be flattened for similarity calculations.

examples/fedper_example/client.py
```diff
@@ -0,0 +1,73 @@
...
+        downsample_percentage = self.narrow_config_type(config, "downsampling_ratio", float)
+        sampler = MinorityLabelBasedSampler(list(range(10)), downsample_percentage, self.minority_numbers)
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler)
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        model: nn.Module = FedPerModel(
+            global_feature_extractor=FedPerGloalFeatureExtractor(),
+            local_prediction_head=FedPerLocalPredictionHead(),
+            flatten_features=True,
```
emersodb, line 45: Flatten features is used to make the model compatible with MOON which requires the intermediate feature representations to be flattened for similarity calculations.

fl4health/model_bases/moon_base.py
```diff
@@ -18,5 +18,5 @@ def forward(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[st
         x = self.base_module.forward(input)
         features = self.projection_module.forward(x) if self.projection_module else x
         preds = self.head_module.forward(features)
-        # Return preds and features as seperate dictionairy as in fenda base
-        return {"prediction": preds}, {"features": features.view(len(features), -1)}
+        # Return preds and features as seperate dictionary as in fenda base
+        return {"prediction": preds}, {"features": features.reshape(len(features), -1)}
```
emersodb, line 22: reshape is slightly more general than view

fl4health/clients/moon_client.py
emersodb: That's a fair question. The main reason I didn't do that was because FedPer model's exchange a partial subset of weights and don't, at least by default, admit projection modules for their features. They are very related. So it's possible that unifying them is a good idea. What do you think?

fl4health/model_bases/fedper_base.py
emersodb: The reason I don't flatten by default is that it changes the shape of the features tensor. If a user was going to do something downstream with the features (other than the MOON calculations), I think they would be surprised if they weren't in the expected shape.

examples/fedper_example/client.py
emersodb: Yes, definitely. Thanks for pointing that out.

examples/fedper_example/README.md
```diff
@@ -0,0 +1,48 @@
+# FedPer Federated Learning Example
+This example provides an example of training a FedPer type model on a non-IID subset of the MNIST data. The FL server
+expects three clients to be spun up (i.e. it will wait until three clients report in before starting training). Each client
+has a modified version of the MNIST dataset. This modification essentially subsamples a certain number from the original
+training and validation sets of MNIST in order to synthetically induce local variations in the statistical properties
+of the clients training/validation data. In theory, the models should be able to perform well on their local data
+while learning from other clients data that has different statistical properties. The subsampling is specified by
+sending a list of integers between 0-9 to the clients when they are run with the argument `--minority_numbers`.
```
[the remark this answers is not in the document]
emersodb: The client here: `examples/fedper_example/client.py` does take the `--minority_numbers` and uses `MinorityLabelBasedSampler` in the `get_data_loaders` function. Please correct me if I'm wrong though

examples/fedper_example/README.md
```diff
@@ -0,0 +1,48 @@
...
+The next step is to start the server by running
+```
+python -m examples.fedper_example.server  --config_path /path/to/config.yaml
+```
+from the FL4Health directory. The following arguments must be present in the specified config file:
+* `n_clients`: number of clients the server waits for in order to run the FL training
+* `local_epochs`: number of epochs each client will train for locally
+* `batch_size`: size of the batches each client will train on
+* `n_server_rounds`: The number of rounds to run FL
+* `downsampling_ratio`: The amount of downsampling to perform for minority digits
```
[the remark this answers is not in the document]
emersodb: So the config does have this argument, but based on the `source_specified` comment, I think you might be looking at the wrong config, as that sounds like the config for the `feature_alignment_example`?

examples/fedper_example/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the three
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.fedper_example.client --dataset_path /path/to/data --minority_numbers <sequence of numbers>
```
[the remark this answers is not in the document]
emersodb: See the comments above. I think we're okay here. I did run the example with these args and it worked appropriately.

examples/fedper_example/README.md
```diff
@@ -0,0 +1,48 @@
+# FedPer Federated Learning Example
+This example provides an example of training a FedPer type model on a non-IID subset of the MNIST data. The FL server
+expects three clients to be spun up (i.e. it will wait until three clients report in before starting training). Each client
```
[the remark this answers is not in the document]
sanaAyrml: Oh true I somehow got confused with feature alignment example and thought this is added readme for it.

fl4health/clients/moon_client.py
sanaAyrml: projection_module is also optional in moon_base, so you can easily pass by None for it in inheritance, and everything should work well. I kinda prefer fedper_base to inherit from both moon and partial_layer_exchange_model so users can get the relationship between all of them and the added functionality of fed_per based on both.

emersodb: Yeah I think that makes sense. I'll do that right now.

emersodb: I added the assert back in now that FedPer inherits from MOON

sanaAyrml (approved):

fatemetkl (approved):
Everything looks good to me as well! No other comments, thanks for adding the tests!

PR #73, opened 2023-11-23
Build a first pass, basic smoke test associated with one of the examples

lotif:
# PR Type

# Short Description

# Tests Added

### DIFF
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -1,0 +1,32 @@
+# only has to pass for python 3.9
+name: Smoke Tests
+
+on:
+  push:
+    branches:
+      main
+  pull_request:
+    branches:
+      main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.9
+          cache: 'pip' # Display the Python version being used
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Install Requirements
+        run: |
+          pip install --upgrade pip
+          pip install $(grep -v '^torchdata\|^torchtext\|^torcheval' requirements.txt)
+      - name: Run Script
+        run: sh examples/utils/run_fl_local.sh # TODO create new script to check for logs, move it to the tests folder
+      - name: Clearing pip for space
+        run: pip uninstall -y -r requirements.txt


tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+
+    logger.info("All clients finished execution")
+
+    # Collecting the server output when its process finish
+    full_server_output = await _wait_for_process_to_finish_and_retrieve_logs(server_process, "Server")
+
+    logger.info("Server has finished execution")
+
+    # server assertions
```
lotif, line 93: Please check the assertions below and let me know if they look correct, if there is anything I should remove or anything else I should add.

.github/workflows/smoke_tests.yaml
```diff
@@ -0,0 +1,33 @@
+# only has to pass for python 3.9
+name: Smoke Tests
+
+on:
+  push:
+    branches:
+      main
```
emersodb, line 7: Just want to verify my understanding, does this mean that these tests would run on any forced pushes to main? Just want to make sure we don't run them each time we push to origin on development branches

.github/workflows/smoke_tests.yaml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          python-version: 3.9
+          cache: 'pip'
+      # Display the Python version being used
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Install Requirements
+        run: |
+          pip install --upgrade pip
+          pip install $(grep -v '^torchdata\|^torchtext\|^torcheval' requirements.txt)
```
emersodb, line 29: This is a bit of a hack that I put in to forgo installing these packages in the test environment, as they are not necessary, but I don't love how I accomplished it. Very open to ideas if you have any.

mypy.ini
```diff
@@ -8,3 +8,4 @@ non_interactive = True
 disallow_untyped_defs = True
 no_implicit_optional = True
 check_untyped_defs = True
+exclude = venv
```
emersodb, line 11: Just out of curiosity, what does this option do?

tests/smoke_tests/config.yaml
```diff
@@ -0,0 +1,25 @@
+# Parameters that describe server
```
emersodb, line 1: Since this is just going to be a smoke test, I'll suggest some config modifications here for it to run a bit faster.

`n_server_rounds` can be set to 2-3. This is the number of exchanges/aggregations done between server and clients. We want to test more than one exchange, but don't need to run 15.

`proximal_weight_patience` can be set to 1. If we're going to only be doing a few rounds, we want to make sure the `proximal_weight` is adjusted. Patience means that the algorithm waits a certain number of server rounds before making any changes.

Rather than using `local_epochs` we can use `local_steps` and set it to something small, maybe 5? Local epochs forces each client to run through all batches in the train set. If we use `local_steps` we can limit the number of train batches each client does.

I think you can safely remove the `reporting_config` section here. These are for W and B experiment tracking, which we don't need if we're just running a smoke test.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+) -> None:
+    with open(config_path, "r") as file:
+        config = yaml.safe_load(file)
+
+    _preload_dataset(dataset_path, config)
+
+    # Start the server and capture its process object
+    logger.info("Starting server...")
+    server_process = await asyncio.create_subprocess_exec(
+        "nohup",
```
emersodb, line 29: Dumb question, but I'm not familiar with the asyncio library. If we're performing an async execution with the library, do we need the nohup part of this call? I normally use nohup to "force" an async execution on the terminal but it feels like maybe the asyncio library would already do that?

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+        "-m",
+        "examples.fedprox_example.server",
+        "--config_path",
+        config_path,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.STDOUT,
+    )
+
+    # reads lines from the server output in search of the startup log message
+    # times out after 20s of inactivity if it doesn't find the log message
```
emersodb, line 40: Nice!

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+    full_server_output = ""
+    startup_message = "FL starting"
+    output_found = False
+    while True:
+        try:
+            assert server_process.stdout is not None, "Server's process stdout is None"
+            server_output_in_bytes = await asyncio.wait_for(server_process.stdout.readline(), 20)
```
emersodb, line 47: Out of curiosity, does this `wait_for` "refresh" each time the readline() here executes properly? That is, the process will wait 20 seconds from the time a line was last written?

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.STDOUT,
+    )
+
+    # reads lines from the server output in search of the startup log message
+    # times out after 20s of inactivity if it doesn't find the log message
+    full_server_output = ""
+    startup_message = "FL starting"
+    output_found = False
+    while True:
```
emersodb, line 44: Super minor, but it might be slightly cleaner to do
```
while not output_found:
    ...
    if startup_message in server_output:
        output_found = True
```
avoiding the "break"

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+            full_server_output += server_output
+        except asyncio.TimeoutError:
+            logger.error(f"Timeout waiting for server startup message '{startup_message}'")
+            break
+        logger.debug(f"Server output: {server_output}")
+        if startup_message in server_output:
+            output_found = True
+            break
+
+    assert output_found, (
```
emersodb, line 58: Also super minor, but I don't think you need the + here. That is python will automatically concatenate the f-strings if you do 
```
f"Full output:\n{full_server_output}\n"
f"[ASSERT_ERROR] Startup log message '{startup_message}' not found in server output."
```
I might be wrong though.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+    logger.info("All clients finished execution")
+
+    # Collecting the server output when its process finish
+    full_server_output = await _wait_for_process_to_finish_and_retrieve_logs(server_process, "Server")
+
+    logger.info("Server has finished execution")
+
+    # server assertions
+    assert "error" not in full_server_output.lower(), (
+        f"Full output:\n{full_server_output}\n" + "[ASSERT ERROR] Error message found for server."
```
emersodb, line 97: Similar to the f-string above, I don't think you need to + part in these. They can either be part of the f-string proper or on a separate line and they'll work nicely 🙂

tests/smoke_tests/run_smoke_test.py
emersodb: I think these look good to me. For this first pass I think these are awesome. 

For a follow up ticket, I think one thing that would be great, is if we set random seeds in the various places (potentially using Sana's module she has a PR up for right now) to make training deterministic. Then we can also check that the output metrics are as expected. This would catch issues that don't break the functionality, but might have altered (in an unexpected way) the way training happens.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Creating a client and getting the data loaders will trigger
+        # the dataset's download
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        client = MnistFedProxClient(Path(dataset_path), [Accuracy()], device)
+        client.get_data_loaders(config)
+
+        logger.info("Finished preloading MNIST dataset")
+    else:
+        logger.info("Preload not supported for specified dataset. Skipping.")
```
emersodb, line 151: Should we just throw an error here rather than proceeding? I imagine we'd blow up somewhere else, so might as well do it here a bit earlier? 😂

emersodb:
Overall, I think this looks really good! I think we can change a few args in the config to speed up the test a fair bit. Other than that just a few small comments/questions.

tests/smoke_tests/run_smoke_test.py
lotif: That's a great idea. I'll keep an eye on that PR and create a ticket.

tests/smoke_tests/run_smoke_test.py
lotif: No, it will return right away after each line and only waits if there are no more new lines written in the output. In other words, simplistically it will immediately return the next line in the buffer if there is any, if not wait 20s for one to be written.

tests/smoke_tests/run_smoke_test.py
lotif: You're right, it works. I personally don't like that syntax I came up with, but that's what works best or `black` would come up with some even uglier formatting.

tests/smoke_tests/run_smoke_test.py
lotif: If I break it in two different lines `black` will put it back into the same line because it's not over the line limit, and if I put it in a single string, `black` will format it like this:
```
assert (
    "error" not in full_server_output.lower()
), f"Full output:\n{full_server_output}\n[ASSERT ERROR] Error message found for server."
```
I honestly don't know what's worse. I have a slight preference for breaking the string in two. `black` is a bit too annoying on those edge cases.

tests/smoke_tests/run_smoke_test.py
lotif: I like the thinking, but the MNIST error seems to be specific to the server hosting it (details in the issue I linked in the code comment above). It's possible we don't need to preload other datasets, I'll find out about that when I add the other tests.

.github/workflows/smoke_tests.yaml
lotif: One thing I did in the past to solve that is to have two separate requirements files, one for test and one for prod. Also solves the issue of having test packages being unnecessarily added as dependencies for prod.

examples/fedprox_example/server.py
```diff
@@ -51,7 +51,7 @@ def main(config: Dict[str, Any], server_address: str) -> None:
     # This function will be used to produce a config that is sent to each client to initialize their own environment
     fit_config_fn = partial(
         fit_config,
-        config["local_epochs"],
+        config.get("local_epochs", config["local_steps"]),
```
emersodb, line 54: Note that the `fit_config` still function still packages this value as `"local_epochs": local_epochs` with this change, which isn't quite what we want. A possible way to do this would be 
```
def fit_config(
    local_epochs: Optional[int],
    local_steps: Optional[int],
    batch_size: int,
    n_server_rounds: int,
    reporting_enabled: bool,
    project_name: str,
    group_name: str,
    entity: str,
    current_round: int,
) -> Config:
    config = {"local_epochs": local_epochs} if local_epochs else {"local_steps": local_steps}
    return config.update(
        {
            "batch_size": batch_size,
            "n_server_rounds": n_server_rounds,
            "current_server_round": current_round,
            "reporting_enabled": reporting_enabled,
            "project_name": project_name,
            "group_name": group_name,
            "entity": entity,
        }
    )
```
and
```
    fit_config_fn = partial(
        fit_config,
        config.get("local_epochs", None),
        config.get("local_steps", None),
        config["batch_size"],
        config["n_server_rounds"],
        config["reporting_config"].get("enabled", False),
        # Note that run name is not included, it will be set in the clients
        config["reporting_config"].get("project_name", ""),
        config["reporting_config"].get("group_name", ""),
        config["reporting_config"].get("entity", ""),
    )
```

.github/workflows/smoke_tests.yaml
emersodb: For that, the only thing I don't love is that then we have to maintain two sets of requirement files. We can leave it for now and maybe put a ticket in the backlog to consider it more deeply later.

mypy.ini
emersodb: Gotcha! That's fine with me.

tests/smoke_tests/run_smoke_test.py
emersodb: Ah I see. That makes sense.

examples/fedprox_example/server.py
lotif: ok! Got it now. It took me a few attempts but I managed to make it work now with a slightly different code. Execution is down to 3 minutes now :)

.github/workflows/smoke_tests.yaml
lotif: Not sure if I understand your concern. Any pushes to main will trigger the workflow, both forced or via pull request merge. Let me know if you see any problems with that.
I assume by development branches you mean branches like this one, right? On the config below the line you commented, it establishes that it will also run when a pull request targeting main is updated (like this one) from any branch. It will not run if there is an update to a branch that doesn't have a pull request into main opened for it.

I copied this from the `tests.yaml` file, so whenever the tests run this will also run.

emersodb (approved):
Looks good to me!

PR #74, opened 2023-11-28
Consolidate BasicClient and NumpyFlClient

jewelltaylor:
# PR Type

# Short Description
A small PR aimed at consolidating `NumpyFlClient` with `BasicClient`. I had to also make some small changes to `EvaluateClient` which was the last client to extend `NumpyFlClient`.  I also documented the BasicClient fully trying to stick the google dosctring format we have settled on. 

Clickup Ticket(s): [Consolidate BasicClient and NumpyClient](https://app.clickup.com/t/86860zcj4)

# Tests Added

- Modified the tests on `NumpyFlClient` to be on `BasicClient`.


fl4health/clients/basic_client.py
```diff
@@ -44,30 +48,150 @@ def __init__(
...
-
         self.train_loader: DataLoader
         self.val_loader: DataLoader
         self.num_train_samples: int
         self.num_val_samples: int
         self.learning_rate: float
 
-        # Need to track total_steps across rounds for WANDB reporting
-        self.total_steps: int = 0
+    def _maybe_checkpoint(self, comparison_metric: float) -> None:
```
lotif, line 82: I think the `comparison_metric` name is a bit misleading, it's a value not a metric. Maybe `comparison_threshold` would be a better name?

emersodb: I would also argue that this isn't a threshold per se. The threshold is actually held by the checkpointing. For example, the `BestMetricTorchCheckpointer` stores the best "metric" that it's seen thus far. Maybe we call it `current_metric`?

fl4health/clients/basic_client.py
```diff
@@ -44,30 +48,150 @@ def __init__(
...
+           length (int): The length of the hash.
+
+        Returns:
+            str: client id
+        """
+        return "".join(random.choice(string.ascii_lowercase) for i in range(length))
+
+    def get_parameters(self, config: Config) -> NDArrays:
+        """
+        Determines which weights are sent back to the server for aggregation. This uses a parameter exchanger to
```
emersodb, line 107: Super minor: maybe we can say "parameters" instead of "weights" in the "determines which weights..." since weights sort of implies we only exchange model weights?

fl4health/clients/basic_client.py
```diff
@@ -44,30 +48,150 @@ def __init__(
...
+        Args:
+            config (Config): The config object from the server.
+            config_key (str): The key to check config for.
+            narrow_type_to (Type[T]): The expected type of config[config_key]
+
+        Returns:
+            T: The type-checked value at config[config_key]
+
+        Raises:
+            ValueError: If config[config_key] is not of type narrow_type_to.
```
emersodb, line 162: Super minor, but the value error could also be that the key doesn't exist.

fl4health/clients/basic_client.py
```diff
@@ -44,30 +48,150 @@ def __init__(
...
+
+        config_value = config[config_key]
+        if isinstance(config_value, narrow_type_to):
+            return config_value
+        else:
+            raise ValueError(f"Provided configuration key ({config_key}) value does not have correct type")
+
+    def shutdown(self) -> None:
+        """
+        Shuts down the client. Inolves shutting down W&B reporter if one exists.
```
emersodb, line 175: Inolves -> Involves 🙂

fl4health/clients/basic_client.py
```diff
@@ -44,30 +48,150 @@ def __init__(
...
+            config (Config): The config from the server.
+
+        Returns:
+            Tuple[Union[int, None], Union[int, None], int]: Returns the local_epochs, local_steps and
+                current_server_round. Ensures only one of local_epochs and local_steps is defined in the config
+                and sets the one that is not to None.
+
+        Raises:
+            ValueError: If the config contains both local_steps and local epochs or if local_steps, local_epochs or
+                current_server_round is of thje wrong type (int).
```
emersodb, line 194: "the wrong type"

fl4health/clients/basic_client.py
```diff
@@ -86,6 +210,20 @@ def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, N
...
+        Args:
+            parameters (NDArrays): The parameters of the model to be used in fit.
+            config (NDArrays): The config from the server.
+
+        Returns:
+            Tuple[NDArrays, int, Dict[str, Scalar]]: The parameters following the local training along with the
+            number of samples in the local training dataset and the computed metrics throughout the fit.
+
+        Raises:
+            ValueError: If local_steps or local_epochs is not speccified in config.
```
emersodb, line 225: specified 🙂

fl4health/clients/basic_client.py
```diff
@@ -113,6 +251,17 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
...
+        """
+        Evaluates the model on the validation set.
+
+        Args:
+            parameters (NDArrays): The parameters of the model to be evaluated.
+            config (NDArrays): The config object from the server.
+
+        Returns:
+            Tuple[float, int, Dict[str, Scalar]]: A loss associated with the evaluation, the number of samples in the
+                validation set and the metric_vaues associated with evaluation.
```
emersodb, line 263: metric_values?

fl4health/clients/basic_client.py
```diff
@@ -201,6 +384,16 @@ def val_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Losses, D
...
     ) -> Tuple[Dict[str, float], Dict[str, Scalar]]:
+        """
+        Train locally for the specified number of epochs.
+
+        Args:
+            epochs (int): The number of epochs for local training.
+            current_round (Optional[int]): The current FL round.
+
+        Returns:
+            Tuple[Dict[str, float], Dict[str, Scalar]]: The loss and metrics dictionary from the local training.
```
emersodb, line 395: Maybe mention that the loss here is a dictionary potentially corresponding to different components of the loss?

fl4health/clients/basic_client.py
```diff
@@ -187,8 +362,16 @@ def train_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Losses,
 
     def val_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Losses, Dict[str, torch.Tensor]]:
         """
-        Given input and target, compute loss, update loss and metrics
+        Given input and target, compute loss, update loss and metrics.
```
emersodb, line 365: In the doc string, perhaps mention that this function is meant to execute evaluation on a single batch?

fl4health/clients/basic_client.py
emersodb: Similarly for `train_by_steps` maybe

fl4health/clients/basic_client.py
```diff
@@ -314,11 +533,17 @@ def setup_client(self, config: Config) -> None:
...
 
     def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         """
         Returns Full Parameter Exchangers. Subclasses that require custom Parameter Exchangers can override this.
+
+        Args:
+            config (Config): The config from server.
+
+        Returns:
+            ParameterExchanger: Used to exchange weights between server and client.
```
emersodb, line 546: weights -> parameters, just to be a bit more general with what we can exchange?

fl4health/clients/basic_client.py
```diff
@@ -394,38 +626,79 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
+
+        Raises:
+            NotImplementedError: To be defined in child class.
         """
         raise NotImplementedError
 
     def get_optimizer(self, config: Config) -> Union[Optimizer, Dict[str, Optimizer]]:
         """
         Method to be defined by user that returns the PyTorch optimizer used to train models locally
         Return value can be a single torch optimizer or a dictionary of string and torch optimizer.
```
emersodb, line 656: Maybe we can mention a situations where we might have multiple optimizers. I.e. APFL?

fl4health/clients/basic_client.py
```diff
@@ -394,38 +626,79 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
+
+        Raises:
+            NotImplementedError: To be defined in child class.
         """
         raise NotImplementedError
 
     def get_model(self, config: Config) -> nn.Module:
         """
-        User defined method that Returns PyTorch model
+        User defined method that Returns PyTorch model.
```
emersodb, line 672: Returns to returns 🙂

fl4health/clients/basic_client.py
```diff
@@ -394,38 +626,79 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
+
+        Raises:
+            NotImplementedError: To be defined in child class.
         """
         raise NotImplementedError
 
     def update_after_train(self, local_steps: int, loss_dict: Dict[str, float]) -> None:
         """
-        Called after training with the number of local_steps performed over the FL round and
+        Hook method Called after training with the number of local_steps performed over the FL round and
```
emersodb, line 687: lowercase called 🙂

emersodb: It might be useful to provide an example here of where this might be used. I believe this is used by SCAFFOLD and FedProx

fl4health/clients/basic_client.py
```diff
@@ -394,38 +626,79 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
+        Args:
+            local_steps (int): The number of steps in the local training.
+            loss_dict (Dict[str, float]): A dictionary of losses from local training.
         """
         pass
 
     def update_after_step(self, step: int) -> None:
         """
-        Called after local train step on client. step is an integer that represents
+        Hook method Called after local train step on client. step is an integer that represents
```
emersodb, line 698: lowercase called

fl4health/clients/basic_client.py
```diff
@@ -394,38 +626,79 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
         the local training step that was most recently completed.
+
```
emersodb, line 700: Maybe mention an example of when this is used (I think APFL?)

fl4health/clients/evaluate_client.py
```diff
@@ -118,10 +130,10 @@ def validate_on_model(
         with torch.no_grad():
             for inputs, targets in self.data_loader:
                 inputs, targets = inputs.to(self.device), targets.to(self.device)
-                preds = model(inputs)
-                losses = self.compute_loss(preds, targets)
+                preds = {"prediction": model(inputs)}
```
emersodb, line 133: Good catch here.

fl4health/clients/evaluate_client.py
```diff
@@ -98,7 +107,10 @@ def evaluate(self, parameters: NDArrays, config: Config) -> Tuple[float, int, Di
             metric_values,
         )
 
-    def _handle_logging(self, losses: Losses, metrics_dict: Dict[str, Scalar], is_global: bool) -> None:
+    def _handle_logging(  # type: ignore
```
emersodb, line 110: Is this type ignore to handle the fact that the signatures don't match here?

fl4health/clients/evaluate_client.py
```diff
@@ -190,33 +202,12 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
-        """
-        Computes loss given preds and torch and the user defined criterion. Optionally includes dictionary of
-        loss components if you wish to train the total loss as well as sub losses if they exist.
-        """
-        loss = self.criterion(preds, target)
-        losses = Losses(checkpoint=loss, backward=loss)
-        return losses
-
-    def get_data_loader(self, config: Config) -> DataLoader:
+    def get_data_loader(self, config: Config) -> DataLoader:  # type: ignore
```
emersodb, line 205: Why is the type: ignore required here?

jewelltaylor: Wasn't a problem when we extended NumpyFlClient but now it is

fl4health/clients/evaluate_client.py
jewelltaylor: Yeah the signatures don't match so it won't pass unless we type ignore

fl4health/clients/evaluate_client.py
emersodb: Yeah I see now. Maybe this is a dumb idea but we could have `get_data_loader` return a type `Tuple[DataLoader, ...]` This allows for a tuple of unspecified length? I'm not sure if it would completely break everything though 😂

fl4health/clients/basic_client.py
jewelltaylor: Good call!

emersodb (approved):
Looks good to me. Very small comment on the docstring for `get_data_loaders` after updating the return type.

fl4health/clients/basic_client.py
```diff
@@ -622,7 +626,7 @@ def clone_and_freeze_model(self, model: nn.Module) -> nn.Module:
 
         return cloned_model
 
-    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, ...]:
```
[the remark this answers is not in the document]
jewelltaylor: Good catch! My bad

PR #75, opened 2023-12-01
Set seed

sanaAyrml:
# PR Type
Feature

# Short Description

This pull request implements seed setting for both clients and servers.

- On the client side, it is configured in the `__init__` function of the `BasicClient` class located at `fl4health/clients/basic_client.py` as `dumpy_fl_client.py` is going to be omitted  based on https://github.com/VectorInstitute/FL4Health/pull/74#issue-2015469124.

- On the server side, it is configured in the `__init__` function of the `FLServer` class at `fl4health/server/base_server.py`. Consequently, all `server.py` files in the `examples` dir have been updated to use the our implemented `FLServer` or its inheritances instead of flower's default server.

- Additionally, in `server.py` and `client.py` of `examples/fedprox_example`, I have implemented a sample of setting the seed by the user rather than using the default one through command line arguments.

Clickup Ticket(s): https://app.clickup.com/t/8686kfkh2

# Tests Added

Couple of tests added to `tests/clients` and `tests/server` to check whether seed is set.


fl4health/clients/basic_client.py
```diff
@@ -44,7 +45,7 @@ def __init__(
             checkpointer (Optional[TorchCheckpointer], optional): Checkpointer to be used for client-side
                 checkpointing. Defaults to None.
         """
-        super().__init__(data_path, device)
+        super().__init__(data_path, device, seed=seed)
```
emersodb, line 48: It looks like the seed setting still resides in the NumpyFLClient. As noted in the PR description, since this is going to be merged with BasicClient, let's move the seed logic into this class?

fl4health/clients/numpy_fl_client.py
```diff
@@ -17,7 +20,16 @@
 
 
 class NumpyFlClient(NumPyClient):
-    def __init__(self, data_path: Path, device: torch.device) -> None:
+    def __init__(self, data_path: Path, device: torch.device, seed: Optional[int] = None) -> None:
+        if seed is None:
```
emersodb, line 24: I think we can encapsulate this into a function called something like `fix_random_seeds` that takes the optional seed and performs this logic and logging?

fl4health/server/base_server.py
```diff
@@ -41,6 +45,16 @@ def __init__(
                 server side checkpointing based on some criteria. If none, then no server-side checkpointing is
                 performed. Defaults to None.
         """
+        if seed is None:
```
emersodb, line 48: Some comment as in the client code, we can encapsulate this logic into a function and I would suggest when seed is none then we don't set anything?

fl4health/clients/basic_client.py
emersodb: To incorporate his changes you can simple merge in his branch and set the target of this PR to his branch until it is merged.

emersodb:
Overall, looks good. Just a few small behavior suggestions and code encapsulation, along with an expansion of the tests to make sure the seeds are doing what we expect them to do.

fl4health/clients/basic_client.py
sanaAyrml: Yes in current version I moved it into the basic client.

fl4health/clients/numpy_fl_client.py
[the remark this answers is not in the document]
sanaAyrml: Oh sure, I thought it would be safer to have a set seed. But I will omit the default setting based on your suggestion.

tests/server/test_base_server.py
```diff
@@ -19,6 +19,11 @@ def _hydrate_model_for_checkpointing(self) -> nn.Module:
         return model
 
 
+def test_seed_setting() -> None:
```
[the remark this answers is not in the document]
sanaAyrml: That is good idea, the test right now is just for testing whether default seeds are set or not. Which I eliminated it. I will update tests.

sanaAyrml:
> Overall, looks good. Just a few small behavior suggestions and code encapsulation, along with an expansion of the tests to make sure the seeds are doing what we expect them to do.

sanaAyrml:
Based on the recommendations, I have encapsulated seed settings within a function named `_maybe_fix_random_seeds` for both the server and the client. Additionally, according to my tests and (Link)[https://discuss.pytorch.org/t/difference-between-torch-manual-seed-and-torch-cuda-manual-seed/13848/5], it seems that in newer versions of PyTorch, there's no need to set the GPU seed separately using `torch.cuda.manual_seed`. Therefore, I have omitted this step.

Finally, I have included a test to verify whether the seed is set correctly for both the server and the client based on David's suggestion. I have also removed previous tests as they were designed to check whether the seed is set by default, which is not the case in the current version. Users are now expected to set the seed explicitly.

fl4health/clients/evaluate_client.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         device: torch.device,
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
         model_checkpoint_path: Optional[Path] = None,
+        seed: Optional[int] = None,
```
emersodb, line 34: since evaluate_client doesn't call super, you just need to set the seed manually here. I.e. just
```
self._maybe_fix_random_seeds(seed)
```
in the at the beginning of the __init__

emersodb (approved):
This looks good to go to me, other than the minor comment about calling the seed setting function in the evaluation client explicitly.

PR #77, opened 2023-12-05
Federated ensemble

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Federated training ensemble of models](https://app.clickup.com/t/8686ja6he)

Small PR that adds some initial functionality to support training ensembles of models in a federated manner. To this end, I added two classes: 
- `EnsembleModel`: A nn.Module that acts as a wrapper for the ensemble of models with support to produce predictions for individual models as well as an ensemble prediction via voting or averaging.
- `EnsembleClient`: A custom client to handle specific details of training a federated ensemble of models such as setting up the client, computing losses and performing a train step. 

In the process I had to make some minor changes to the `BasicClient` such as store client optimizers in a dictionary  and making the backward attribute of `Losses` a union of Tensor and a dictionary of Tensors. 

It is worth mentioning that I plan on doing a follow up PR that makes it so that we can use Ensembling in conjunction with other client types (ie Scaffold). This seems like a good first start and it would be good to get some eyes on it before I go further. 

# Tests Added
- Add tests to ensure the `EnsembleModel` produces predictions of the current dimension
- Add tests to ensure the "ensemble prediction methods" (ie voting and averaging) work correctly
- A small addition to the tests on `Losses` focused on the new functionality I mentioned

### DIFF
--- a/examples/ensemble_example/client.py
+++ b/examples/ensemble_example/client.py
@@ -22,22 +22,22 @@ class MnistEnsembleClient(EnsembleClient):
     def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
         batch_size = self.narrow_config_type(config, "batch_size", int)
         sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.75)
-        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler)
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler=sampler)
         return train_loader, val_loader
 
     def get_model(self, config: Config) -> nn.Module:
         ensemble_models: Dict[str, nn.Module] = {
-            "ensemble-model-0": ConfigurableMnistNet(out_channel_mult=1).to(self.device),
-            "ensemble-model-1": ConfigurableMnistNet(out_channel_mult=2).to(self.device),
-            "ensemble-model-2": ConfigurableMnistNet(out_channel_mult=3).to(self.device),
+            "model_0": ConfigurableMnistNet(out_channel_mult=1).to(self.device),
+            "model_1": ConfigurableMnistNet(out_channel_mult=2).to(self.device),
+            "model_2": ConfigurableMnistNet(out_channel_mult=3).to(self.device),
         }
         return EnsembleModel(ensemble_models)
 
     def get_optimizer(self, config: Config) -> Dict[str, Optimizer]:
         ensemble_optimizers: Dict[str, torch.optim.Optimizer] = {
-            "ensemble-model-0": torch.optim.AdamW(self.model.models["ensemble-model-0"].parameters(), lr=0.01),
-            "ensemble-model-1": torch.optim.AdamW(self.model.models["ensemble-model-1"].parameters(), lr=0.01),
-            "ensemble-model-2": torch.optim.AdamW(self.model.models["ensemble-model-2"].parameters(), lr=0.01),
+            "model_0": torch.optim.AdamW(self.model.model_0.parameters(), lr=0.01),  # type: ignore
+            "model_1": torch.optim.AdamW(self.model.model_1.parameters(), lr=0.01),  # type: ignore
+            "model_2": torch.optim.AdamW(self.model.model_2.parameters(), lr=0.01),  # type: ignore
         }
         return ensemble_optimizers
 
--- a/examples/ensemble_example/client.py
+++ b/examples/ensemble_example/client.py
@@ -21,7 +21,7 @@ from fl4health.utils.sampler import DirichletLabelBasedSampler
 class MnistEnsembleClient(EnsembleClient):
     def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
         batch_size = self.narrow_config_type(config, "batch_size", int)
-        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.75)
+        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=float(config["sample_percentage"]))
         train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler=sampler)
         return train_loader, val_loader
 
--- a/examples/ensemble_example/client.py
+++ b/examples/ensemble_example/client.py
@@ -1,0 +1,58 @@
+import argparse
+from pathlib import Path
+from typing import Dict, Tuple
+
+import flwr as fl
+import torch
+import torch.nn as nn
+from flwr.common.typing import Config
+from torch.nn.modules.loss import _Loss
+from torch.optim import Optimizer
+from torch.utils.data import DataLoader
+
+from examples.models.ensemble_cnn import ConfigurableMnistNet
+from fl4health.clients.ensemble_client import EnsembleClient
+from fl4health.model_bases.ensemble_base import EnsembleModel
+from fl4health.utils.load_data import load_mnist_data
+from fl4health.utils.metrics import Accuracy
+from fl4health.utils.sampler import DirichletLabelBasedSampler
+
+
+class MnistEnsembleClient(EnsembleClient):
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        batch_size = self.narrow_config_type(config, "batch_size", int)
+        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.75)
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler)
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        ensemble_models: Dict[str, nn.Module] = {
+            "ensemble-model-0": ConfigurableMnistNet(out_channel_mult=1).to(self.device),
+            "ensemble-model-1": ConfigurableMnistNet(out_channel_mult=2).to(self.device),
+            "ensemble-model-2": ConfigurableMnistNet(out_channel_mult=3).to(self.device),
+        }
+        return EnsembleModel(ensemble_models)
+
+    def get_optimizer(self, config: Config) -> Dict[str, Optimizer]:
+        ensemble_optimizers: Dict[str, torch.optim.Optimizer] = {
+            "ensemble-model-0": torch.optim.AdamW(self.model.models["ensemble-model-0"].parameters(), lr=0.01),
+            "ensemble-model-1": torch.optim.AdamW(self.model.models["ensemble-model-1"].parameters(), lr=0.01),
+            "ensemble-model-2": torch.optim.AdamW(self.model.models["ensemble-model-2"].parameters(), lr=0.01),
+        }
+        return ensemble_optimizers
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return torch.nn.CrossEntropyLoss()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="FL Client Main")
+    parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
+
+    args = parser.parse_args()
+
+    DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    data_path = Path(args.dataset_path)
+
+    client = MnistEnsembleClient(data_path, [Accuracy()], DEVICE)
+    fl.client.start_numpy_client(server_address="0.0.0.0:8080", client=client)
--- a/examples/ensemble_example/client.py
+++ b/examples/ensemble_example/client.py
@@ -25,7 +25,7 @@ class MnistEnsembleClient(EnsembleClient):
         train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler=sampler)
         return train_loader, val_loader
 
-    def get_model(self, config: Config) -> nn.Module:
+    def get_model(self, config: Config) -> EnsembleModel:
         ensemble_models: Dict[str, nn.Module] = {
             "model_0": ConfigurableMnistNet(out_channel_mult=1).to(self.device),
             "model_1": ConfigurableMnistNet(out_channel_mult=2).to(self.device),
--- a/examples/ensemble_example/client.py
+++ b/examples/ensemble_example/client.py
@@ -35,9 +35,9 @@ class MnistEnsembleClient(EnsembleClient):
 
     def get_optimizer(self, config: Config) -> Dict[str, Optimizer]:
         ensemble_optimizers: Dict[str, torch.optim.Optimizer] = {
-            "model_0": torch.optim.AdamW(self.model.model_0.parameters(), lr=0.01),  # type: ignore
-            "model_1": torch.optim.AdamW(self.model.model_1.parameters(), lr=0.01),  # type: ignore
-            "model_2": torch.optim.AdamW(self.model.model_2.parameters(), lr=0.01),  # type: ignore
+            "model_0": torch.optim.AdamW(self.model.ensemble_models["model_0"].parameters(), lr=0.01),
+            "model_1": torch.optim.AdamW(self.model.ensemble_models["model_1"].parameters(), lr=0.01),
+            "model_2": torch.optim.AdamW(self.model.ensemble_models["model_2"].parameters(), lr=0.01),
         }
         return ensemble_optimizers
 
--- a/fl4health/clients/apfl_client.py
+++ b/fl4health/clients/apfl_client.py
@@ -102,7 +102,7 @@ class ApflClient(BasicClient):
 
     def set_optimizer(self, config: Config) -> None:
         optimizers = self.get_optimizer(config)
-        assert isinstance(optimizers, dict)
+        assert isinstance(optimizers, dict) and set(("global", "local")) == set(optimizers.keys())
         self.optimizers = optimizers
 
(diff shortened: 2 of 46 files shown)


fl4health/clients/apfl_client.py
```diff
@@ -26,8 +26,7 @@ def __init__(
 
         self.model: ApflModule
         self.learning_rate: float
-        self.optimizer: torch.optim.Optimizer
-        self.local_optimizer: torch.optim.Optimizer
+        self.optimizers: Dict[str, torch.optim.Optimizer]
```
jewelltaylor, line 29: Update optimizers attribute to be dict instead of a single optimizer which was necessary for the addition of ensemble models.

fl4health/clients/basic_client.py
```diff
@@ -72,7 +72,7 @@ def __init__(
         # Attributes to be initialized in setup_client
         self.parameter_exchanger: ParameterExchanger
         self.model: nn.Module
-        self.optimizer: torch.optim.Optimizer
+        self.optimizers: Dict[str, torch.optim.Optimizer]
```
jewelltaylor, line 75: Update optimizers attribute to be dict instead of a single optimizer which was necessary for the addition of ensemble models.

fl4health/clients/basic_client.py
```diff
@@ -607,7 +612,7 @@ def set_optimizer(self, config: Config) -> None:
         """
         optimizer = self.get_optimizer(config)
         assert not isinstance(optimizer, dict)
-        self.optimizer = optimizer
+        self.optimizers = {"global": optimizer}
```
jewelltaylor, line 615: If we just have a single optimizer, I call it the global one. Maybe this isn't the best name, feel free to suggest something else.

fl4health/model_bases/ensemble_base.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # If larger than two dimensions, we map to 2D to perfom voting operation (and reshape later)
+        if len(preds_dimension) > 2:
+            preds_list = [preds.reshape(-1, preds_dimension[-1]) for preds in preds_list]
+
+        # For each model prediction, compute the argmax of the model over the classes and stack column-wise into matrix
+        # Each row of matrix represents the argmax of each model for a given sample
+        argmax_per_model = torch.hstack([torch.argmax(preds, dim=1, keepdim=True) for preds in preds_list])
+        # For each row (sample), compute the unique class predictions and their respective counts
+        index_count_list = map(lambda x: torch.unique(x, return_counts=True), argmax_per_model.unbind())
```
jewelltaylor, line 86: Wasn't sure how to do this entirely in torch because unique returns a tuple. Feel free to suggest something more efficient.

fl4health/utils/losses.py
```diff
@@ -11,7 +11,7 @@ class Losses:
     def __init__(
         self,
         checkpoint: torch.Tensor,
-        backward: torch.Tensor,
+        backward: Union[torch.Tensor, Dict[str, torch.Tensor]],
```
jewelltaylor, line 14: Backward loss is a dictionary in the case of ensembles.

fl4health/utils/losses.py
```diff
@@ -74,7 +80,20 @@ def compute(self) -> Losses:
         num_losses = len(self.losses_list)
         # Compute average checkpoint and backward losses across list
         checkpoint_loss = torch.sum(torch.FloatTensor([losses.checkpoint for losses in self.losses_list])) / num_losses
-        backward_loss = torch.sum(torch.FloatTensor([losses.backward for losses in self.losses_list])) / num_losses
+        backward_loss: Union[torch.Tensor, Dict[str, torch.Tensor]]
```
jewelltaylor, line 83: my py was giving me some trouble unless I had these somewhat redundant asserts.

examples/ensemble_example/client.py
```diff
@@ -0,0 +1,58 @@
...
+
+class MnistEnsembleClient(EnsembleClient):
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        batch_size = self.narrow_config_type(config, "batch_size", int)
+        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.75)
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler=sampler)
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        ensemble_models: Dict[str, nn.Module] = {
```
emersodb, line 29: Super unimportant, but do you need the type annotation for this dictionary?

fl4health/clients/basic_client.py
```diff
@@ -531,7 +532,11 @@ def setup_client(self, config: Config) -> None:
         self.num_val_samples = len(self.val_loader.dataset)  # type: ignore
 
         self.set_optimizer(config)
-        self.learning_rate = self.optimizer.defaults["lr"]
+        if "global" in self.optimizers:
```
emersodb, line 535: Should we make `learning_rate` and optional parameter and have clients only set it when it makes sense to use? That is, the only default client that needs it seems to be SCAFFOLD and we can set it by overriding `setup_client` and then doing `self.optimizers["global"].defaults["lr"]`?

fl4health/clients/apfl_client.py
```diff
@@ -101,10 +101,9 @@ def compute_loss(
         return losses
 
     def set_optimizer(self, config: Config) -> None:
-        optimizer_dict = self.get_optimizer(config)
-        assert isinstance(optimizer_dict, dict)
-        self.optimizer = optimizer_dict["global"]
-        self.local_optimizer = optimizer_dict["local"]
+        optimizers = self.get_optimizer(config)
+        assert isinstance(optimizers, dict)
```
emersodb, line 105: We should probably assert that the dictionary is of length two and contains the proper keys here?

fl4health/clients/basic_client.py
emersodb: This would also avoid the "awkward" setting of the learning rate here to be the "first" optimizer in the dictionary, which I think is a bit "confusing" and sort of implicitly assumes that all optimizer learning rates are equally valid for this value?

fl4health/clients/basic_client.py
emersodb: I don't have a good alternative. Global is good for now.

fl4health/clients/basic_client.py
jewelltaylor: Yeah this is a better approach for sure

fl4health/clients/ensemble_client.py
```diff
@@ -0,0 +1,149 @@
...
+        optimizers = self.get_optimizer(config)
+        assert isinstance(optimizers, dict)
+        self.optimizers = optimizers
+
+    def train_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Losses, Dict[str, torch.Tensor]]:
+        """
+        Given a single batch of input and target data, generate predictions
+        (both individual models and ensemble prediction), compute loss, update parameters and
+        optionally update metrics if they exist. (ie backprop on a single batch of data).
+        Assumes self.model is in train model already. Differs from parent method in that, there are multiple losses
```
emersodb, line 81: "Assumes self.model is in train model already" to "Assumes self.model is in train mode already"

fl4health/utils/losses.py
```diff
@@ -21,7 +21,13 @@ def __init__(
     def as_dict(self) -> Dict[str, float]:
         loss_dict: Dict[str, float] = {}
         loss_dict["checkpoint"] = float(self.checkpoint.item())
-        loss_dict["backward"] = float(self.backward.item())
+
```
emersodb, line 24: In the previous iteration, we had to do a float(...) conversion. Is that no longer necessary?

fl4health/utils/losses.py
emersodb: This is pretty awkward. Maybe, we make the backward value a dictionary as well and have a default key when there is only "one" backward loss. Something like "backward" maybe? That way basic client and any others can just rely on that being there. Then we don't have to do as much here and the types are "unified"?

emersodb: If we do this, the averaging procedure would be the same as `additional_losses`. We should encapsulate the averaging functionality in that case into a class function.

fl4health/model_bases/ensemble_base.py
```diff
@@ -0,0 +1,111 @@
...
+        aggregation_mode: Optional[EnsembleAggregationMode] = EnsembleAggregationMode.AVERAGE,
+    ) -> None:
+        """
+        Class that acts a wrapper to an ensemble of models to be trained in federated manner with support
+        for both voting and averaging prediction of individual models.
+
+        Args:
+            ensemble_models (Dict[str, nn.Module]): A dictionary of models that make up the ensemble.
+            aggregation_mode (Optional[EnsembleAggregationMode]): The mode in which to aggregate the
+                predictions of indivdual models.
```
emersodb, line 26: indivdual -> individual 🙂

fl4health/utils/losses.py
lotif: This could be added as a code comment to explain the `Union`.

fl4health/utils/losses.py
```diff
@@ -74,7 +80,20 @@ def compute(self) -> Losses:
...
+            assert isinstance(self.losses_list[0].backward, dict)
+            backward_loss = {
+                key: torch.sum(
+                    torch.FloatTensor(
+                        [losses.backward[key] for losses in self.losses_list if isinstance(losses.backward, dict)]
+                    )
+                )
+                / num_losses
+                for key in self.losses_list[0].backward.keys()
+            }
```
lotif, line 94: IMHO this would be more easily readable as a regular for loop than a dictionary comprehension.

fl4health/utils/losses.py
```diff
@@ -74,7 +80,20 @@ def compute(self) -> Losses:
...
         # Compute average checkpoint and backward losses across list
         checkpoint_loss = torch.sum(torch.FloatTensor([losses.checkpoint for losses in self.losses_list])) / num_losses
-        backward_loss = torch.sum(torch.FloatTensor([losses.backward for losses in self.losses_list])) / num_losses
+        backward_loss: Union[torch.Tensor, Dict[str, torch.Tensor]]
+        if all(isinstance(losses.backward, dict) for losses in self.losses_list):
+            assert isinstance(self.losses_list[0].backward, dict)
+            backward_loss = {
+                key: torch.sum(
+                    torch.FloatTensor(
+                        [losses.backward[key] for losses in self.losses_list if isinstance(losses.backward, dict)]
```
lotif, line 89: What happens if this `if` condition is not satisfied?

lotif (commented):
You might as well add this to the smoke tests as well now that it has been merged 😬

fl4health/model_bases/ensemble_base.py
emersodb: I think this looks pretty good

tests/models/test_ensemble_base.py
```diff
@@ -0,0 +1,67 @@
...
+import torch
+import torch.nn as nn
+
+from fl4health.model_bases.ensemble_base import EnsembleAggregationMode, EnsembleModel
+from tests.test_utils.models_for_test import SmallCnn
+
+
+def test_forward_average_mode() -> None:
+    models: Dict[str, nn.Module] = {"model_0": SmallCnn(), "model_1": SmallCnn()}
+    em = EnsembleModel(models, EnsembleAggregationMode.AVERAGE)
```
emersodb, line 13: Maybe expand these variable names. em -> ensemble_model and ep -> ensemble_predictions

tests/models/test_ensemble_base.py
```diff
@@ -0,0 +1,67 @@
...
+    data = torch.rand((64, 1, 28, 28))
+    ep = em(data)
+
+    assert len(ep) == 3
+    assert "model_0" in ep and ep["model_1"].shape == torch.Size([64, 32])
```
emersodb, line 18: I think you want this key to be `"model_0"`?

emersodb: Similarly below, I think you want to test that `"model_0" in ep`?

tests/models/test_ensemble_base.py
```diff
@@ -0,0 +1,67 @@
...
+
+def test_forward_vote_mode() -> None:
+    models: Dict[str, nn.Module] = {"model_0": SmallCnn(), "model_1": SmallCnn()}
+    em = EnsembleModel(models, EnsembleAggregationMode.VOTE)
+    data = torch.rand((64, 1, 28, 28))
+    ep = em(data)
+
+    assert len(ep) == 3
+    assert "model_0" in ep and ep["model_0"].shape == torch.Size([64, 32])
+    assert "model_1" in ep and ep["model_0"].shape == torch.Size([64, 32])
```
emersodb, line 31: key probably should be model_1 here?

tests/models/test_ensemble_base.py
```diff
@@ -0,0 +1,67 @@
...
+
+    assert len(ep) == 3
+    assert "model_0" in ep and ep["model_1"].shape == torch.Size([64, 32])
+    assert "model_0" in ep and ep["model_1"].shape == torch.Size([64, 32])
+    assert "ensemble-pred" in ep and ep["ensemble-pred"].shape == torch.Size([64, 32])
+
+
+def test_forward_vote_mode() -> None:
+    models: Dict[str, nn.Module] = {"model_0": SmallCnn(), "model_1": SmallCnn()}
+    em = EnsembleModel(models, EnsembleAggregationMode.VOTE)
```
emersodb, line 25: expand these names for readability as well 🙂

fl4health/utils/losses.py
```diff
@@ -58,97 +55,183 @@ class LossMeterType(Enum):
...
+        """
         raise NotImplementedError
 
     @abstractmethod
     def compute(self) -> Losses:
+        """
+        Compute aggregation of current list of losses if non-empty.
+
+        Returns:
+            Losses: New Losses object with average of losses in losses_list.
```
emersodb, line 93: This doesn't necessarily have to be the "average" of the losses. Could just say aggregation again?

emersodb (approved):
I think all of my comments look good to go, so I'll approve. I think it's worth addressing one or two of Marcelo's comments that are outstanding and having him review again.

examples/ensemble_example/client.py
jewelltaylor: yeah my bad, I meant to mention that its throwing a mypy error if I don't do the type cast with the Dict[str, nn.Module]

fl4health/utils/losses.py
jewelltaylor: Yup, good catch!

examples/ensemble_example/client.py
emersodb: Interesting. Just wanted to check 🙂

examples/ensemble_example/client.py
```diff
@@ -0,0 +1,58 @@
...
+from fl4health.model_bases.ensemble_base import EnsembleModel
+from fl4health.utils.load_data import load_mnist_data
+from fl4health.utils.metrics import Accuracy
+from fl4health.utils.sampler import DirichletLabelBasedSampler
+
+
+class MnistEnsembleClient(EnsembleClient):
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        batch_size = self.narrow_config_type(config, "batch_size", int)
+        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.75)
```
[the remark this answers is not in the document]
jewelltaylor: Added to the config!

examples/ensemble_example/client.py
```diff
@@ -0,0 +1,58 @@
...
+        train_loader, val_loader, _ = load_mnist_data(self.data_path, batch_size, sampler=sampler)
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        ensemble_models: Dict[str, nn.Module] = {
+            "model_0": ConfigurableMnistNet(out_channel_mult=1).to(self.device),
+            "model_1": ConfigurableMnistNet(out_channel_mult=2).to(self.device),
+            "model_2": ConfigurableMnistNet(out_channel_mult=3).to(self.device),
```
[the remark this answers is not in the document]
jewelltaylor: Not a dumb question at all, its good to have an outside perspective asking these questions :) We can definitely add the number of models as a param in the config. The reason why I chose not too is because people may often want to have non uniform architectures across the models, so I think in most cases they wont be just duplicating the same model. And if its not relevant in all cases, maybe its just best to leave up to the user rather than bake in some convenience functionality that may complicate things. Let me know if you think this is fair or I should add it in!

fl4health/utils/losses.py
jewelltaylor: Added a comment and better overall documentation!

fl4health/clients/scaffold_client.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         self.client_control_variates: Optional[NDArrays] = None  # c_i in paper
         self.client_control_variates_updates: Optional[NDArrays] = None  # delta_c_i in paper
         self.server_control_variates: Optional[NDArrays] = None  # c in paper
-        self.optimizer: torch.optim.SGD  # Scaffold require vanilla SGD as optimizer
+        self.optimizers: Dict[str, torch.optim.SGD]  # type: ignore
```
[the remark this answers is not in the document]
jewelltaylor: Oops, added back in :)

PR #78, opened 2023-12-05
Fed pca

yc7z:
# PR Type
[Feature]

Clickup Ticket(s): https://app.clickup.com/t/8686dx41x

This is a draft PR for the federated PCA implementation mainly because I want to see what others think about the current design and which places could be improved.

# Tests Added

No tests for now, will add some later


examples/fedpca_example/plot_digits.py
yc7z, line 1: This file is used to plot the merged principal components as digits (since the example uses MNIST) so that we can visually see the effect of merging.

fl4health/PCA/pca.py
```diff
@@ -0,0 +1,62 @@
...
+    def pc_projection(self, X: NDArray) -> NDArray:
+        return self.principal_components @ X
+
+    def save_pcs(self, pc_path: Path) -> None:
+        pcs, eigenvalues = self.get_pcs()
+        log(INFO, f"principal components shape: {pcs.shape}")
+        log(INFO, f"eigenvalues shape: {eigenvalues.shape}")
+        np.save(pc_path, pcs)
+
+
```
yc7z, line 38: This class is not currently used since I changed design and implemented the merging in a strategy class instead. The alternative is to create a separate `FedPCAServer` class that leverages this to perform merging.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,150 @@
...
+from flwr.server.client_manager import ClientManager
+from flwr.server.client_proxy import ClientProxy
+from flwr.server.strategy import FedAvg
+
+from fl4health.client_managers.base_sampling_manager import BaseFractionSamplingManager
+from fl4health.parameter_exchange.parameter_packer import PrincipalComponentsPacker
+from fl4health.strategies.strategy_with_poll import StrategyWithPolling
+
+
+class FedPCA(FedAvg, StrategyWithPolling):
```
yc7z, line 26: Conceptually it doesn't really make sense to inherit from `FedAvg`, but this way reduces a lot of code duplication if we wish to adhere to making FedPCA a strategy.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def initialize_parameters(self, client_manager: ClientManager) -> Optional[Parameters]:
+        """Initialize the (global) model parameters.
+
+        FedPCA does not need initial global model parameters, so
+        an empty array is used here.
+        """
+        return Parameters(tensors=[], tensor_type="numpy.ndarray")
+
+    def aggregate_fit(
```
yc7z, line 73: After testing a few merging algorithms I opted for simply performing SVD on all the clients principal components on the server side.

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,125 @@
...
+        config_value = config[config_key]
+        if isinstance(config_value, narrow_type_to):
+            return config_value
+        else:
+            raise ValueError(f"Provided configuration key ({config_key}) value does not have correct type")
+
+    def setup_client(self, config: Config) -> None:
+        self.initialized = True
+
+    def evaluate(self, parameters: NDArrays, config: Dict[str, Scalar]) -> Tuple[float, int, Dict[str, Scalar]]:
```
yc7z, line 104: One thing I'm not sure about is what the `evaluate` step should be for federated PCA... Since PCA does not fit into the normal training paradigm, and from what I've read the quality of PCA is usually evaluated through the downstream tasks.

fl4health/PCA/pca.py
```diff
@@ -0,0 +1,62 @@
...
+import numpy as np
+from flwr.common.logger import log
+from flwr.common.typing import NDArray, NDArrays
+from sklearn.decomposition import PCA
+
+
+class ClientSidePCA:
+    def __init__(self, n_components: int) -> None:
+        self._pca = PCA(n_components=n_components, random_state=42)
```
yc7z, line 13: Since we are not assuming a streaming model as the paper below does, I think it is fine to borrow Sklearn's implementation for computing PCA. Let me know if this seems to be a good idea or if it would be preferable to have our own implementation.

https://arxiv.org/pdf/1907.08059.pdf

lotif (changes requested):

fl4health/PCA/pca.py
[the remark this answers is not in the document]
yc7z: I think you're right. I'll change the class to take in a random seed instead.

fl4health/PCA/pca.py
```diff
@@ -0,0 +1,62 @@
...
+        self.eigenvalues: NDArray
+
+    def compute_pc(self, X: NDArray) -> Tuple[NDArray, NDArray]:
+        self._pca.fit(X)
+        components, eigen_vals = self._pca.components_, self._pca.explained_variance_
+        return components, eigen_vals
+
+    def update_pcs(self, pcs: NDArray, eigenvalues: NDArray) -> None:
+        self.principal_components = pcs
+        self.eigenvalues = eigenvalues
```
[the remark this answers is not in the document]
yc7z: Yes that makes sense.

yc7z:
I'm going to close this for now since I have some changes in mind after a discussion with Fatemeh about the code design. Thanks Marcello for the comments!

PR #79, opened 2023-12-05
Add smoke tests for SCAFFOLD, APFL and others, Part 2

lotif:
# PR Type
Other

# Short Description

Clickup Ticket(s): [Link(s) if applicable.](https://app.clickup.com/t/8686m5mqp)

Adding smoke tests for all the other examples except feature alignment, fedopt and partial weight exchange.

# Tests Added

The smoke tests themselves

### DIFF
--- a/examples/basic_example/server.py
+++ b/examples/basic_example/server.py
@@ -1,6 +1,6 @@
 import argparse
 from functools import partial
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import flwr as fl
 import torch.nn as nn
@@ -23,19 +23,31 @@ def get_initial_model_parameters(model: nn.Module) -> Parameters:
 
 
 def fit_config(
-    local_epochs: int,
     batch_size: int,
     current_server_round: int,
+    local_epochs: Optional[int] = None,
+    local_steps: Optional[int] = None,
 ) -> Config:
-    return {"local_epochs": local_epochs, "batch_size": batch_size, "current_server_round": current_server_round}
+    if local_epochs is not None:
+        epochs_or_steps = {"local_epochs": local_epochs}
+    elif local_steps is not None:
+        epochs_or_steps = {"local_steps": local_steps}
+    else:
+        epochs_or_steps = {}
+    return {
+        **epochs_or_steps,
+        "batch_size": batch_size,
+        "current_server_round": current_server_round,
+    }
 
 
 def main(config: Dict[str, Any]) -> None:
     # This function will be used to produce a config that is sent to each client to initialize their own environment
     fit_config_fn = partial(
         fit_config,
-        config["local_epochs"],
         config["batch_size"],
+        local_epochs=config.get("local_epochs"),
+        local_steps=config.get("local_steps"),
     )
 
     # Initializing the model on the server side
--- a/examples/federated_eval_example/client.py
+++ b/examples/federated_eval_example/client.py
@@ -44,13 +44,7 @@ class CifarClient(EvaluateClient):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="FL Client Main")
     parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
-    parser.add_argument(
-        "--checkpoint_path",
-        action="store",
-        type=str,
-        help="Path to client model checkpoint.",
-        required=False,
-    )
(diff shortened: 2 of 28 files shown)


examples/federated_eval_example/client.py
```diff
@@ -44,13 +44,7 @@ def get_criterion(self, config: Config) -> _Loss:
...
     parser = argparse.ArgumentParser(description="FL Client Main")
     parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
-    parser.add_argument(
-        "--checkpoint_path",
-        action="store",
-        type=str,
-        help="Path to client model checkpoint.",
-        required=False,
-    )
+    parser.add_argument("--checkpoint_path", action="store", type=str, help="Path to client model checkpoint.")
```
lotif, line 47: This is not really optional, without the checkpoint the client fails, so I removed the `required=False` from it.

fl4health/clients/evaluate_client.py
```diff
@@ -53,11 +53,12 @@ def __init__(
 
         # The attributes to be set in setup_client
         # Models corresponding to client-side and server-side checkpoints,
-        # if they exist, to be evaluated on the clients dataset.
+        # if they exist, to be evaluated on the client's dataset.
         self.data_loader: DataLoader
         self.criterion: _Loss
         self.local_model: Optional[nn.Module] = None
         self.global_model: Optional[nn.Module] = None
+        self.wandb_reporter = None
```
lotif, line 61: Fixing a bug here that is surfaced when running `shutdown`.

examples/federated_eval_example/client.py
emersodb: So this should work gracefully, as long as either the server or the client provides a model checkpoint path. That is, at least one of them **needs** to be defined, but both do not. I think the reason you're seeing a failure on the client is that both the client and the server are not providing a path. 

I'm not sure what the right thing to do is in the sense that it is _technically_ optional but conditioned on the presence of a server checkpoint. My inclination is too keep this option, as otherwise it signals to the user that they must have a client side checkpoint. Very minor though.

examples/basic_example/server.py
```diff
@@ -23,19 +23,31 @@ def get_initial_model_parameters(model: nn.Module) -> Parameters:
...
 
 def fit_config(
-    local_epochs: int,
     batch_size: int,
     current_server_round: int,
+    local_epochs: Optional[int] = None,
+    local_steps: Optional[int] = None,
 ) -> Config:
-    return {"local_epochs": local_epochs, "batch_size": batch_size, "current_server_round": current_server_round}
+    if local_epochs is not None:
```
emersodb, line 31: We're repeating this code in a lot of places (which is perfectly fine), but maybe we can pull it into a function and place it in `examples/utils` so all examples can reuse it?

.gitignore
```diff
@@ -160,3 +160,4 @@ settings.json
 # model artifacts
 
 **/*.pkl
+/examples/datasets/cifar10_data/
```
emersodb, line 163: We have a `#datasets` section in the .gitignore file. Feel free to stick this in there or rename your folder to `cifar_data` so that its covered by the existing ignore file

tests/smoke_tests/run_smoke_test.py
```diff
@@ -19,6 +21,11 @@ async def run_smoke_test(
     client_python_path: str,
     config_path: str,
     dataset_path: str,
+    checkpoint_path: Optional[str] = None,
+    assert_evaluation_logs: Optional[bool] = False,
+    # The param below exists to work around an issue with some clients
+    # not printing the "Current FL Round" log message reliably
```
emersodb, line 27: I assume this is due to the fact that some clients (for example the evaluation clients) don't print this log? Are there other clients that don't log this message? If so, maybe we can edit them to make sure they do.

lotif: Not that only. I noticed the examples in the `dp_fed_examples` folder don't reliably print that message. Sometimes they print everything and some other times some of the messages are missing, which made the test fail intermittently, so I added that option to mitigate the issue. I haven't noticed that in the other examples. I don't know if there is anything in those examples that would contribute to that behaviour, I didn't go as far as investigating that in detail. Any clues?

emersodb: Interesting. I don't know the issue here, but we can investigate later. I plan to rework some of the privacy accounting soon anyway.

emersodb (approved):
Looks good to go to me.

PR #80, opened 2023-12-06
bug fix in basic client

yc7z:
# PR Type
[Fix]

Bug fix in `basic_client.py`

# Tests Added

`test_setup_client_basic`

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -557,7 +557,7 @@ class BasicClient(NumPyClient):
 
         self.wandb_reporter = ClientWandBReporter.from_config(self.client_name, config)
 
-        self.intialized = True
+        self.initialized = True
 
     def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
         """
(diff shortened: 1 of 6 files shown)


fl4health/clients/basic_client.py
```diff
@@ -557,7 +557,7 @@ def setup_client(self, config: Config) -> None:
 
         self.wandb_reporter = ClientWandBReporter.from_config(self.client_name, config)
 
-        self.intialized = True
+        self.initialized = True
```
yc7z, line 560: This was the bug

emersodb: Good catch. That is definitely a problem. I guess we don't have a test covering this. As part of this. Do you mind added a test that confirms that this is properly set after `setup_client.py` is run?

emersodb (approved):
Other than adding the test to cover the case that was missed here, we're good to go.

jewelltaylor (approved):
The added test looks good! Makes sure the code wont regress to make similar errors in the future :)

PR #83, opened 2023-12-12
Fix random seed bug on initialized parameters

lotif:
# PR Type
Fix

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686q02mq

Making an util function instead to set all the seeds that has to be called before any random operation is done. Cleaning up the old set seed code.

Tested by locally running the fedprox example twice and checking the metrics are exactly the same on both runs.

# Tests Added

Moved the original tests to the random utils file


sanaAyrml (commented):
The current changes look good to me. However, I think it better to also add def initialize_parameters for other strategies and update examples to follow one structure and consequently delete previous `initial_parameters` from strategies input or instead add an explanation of why `initialize_parameters_function`  should be used when seed is set. I personally prefer the first approach because of the consistent examples and less confusion.

lotif:
@emersodb it is global, and I think your idea is good, it will also avoid the code duplication we have right now on that. The user just needs to remember to run the function before anything else.

sanaAyrml:
I also agree, If @lotif agree,I can take care of it over this branch now.

emersodb:
> @emersodb it is global, and I think your idea is good, it will also avoid the code duplication we have right now on that. The user just needs to remember to run the function before anything else.

That makes sense. They have to remember to set a seed if they want one. So I don't think it's too burdensome to instead just run the function instead. If we have it in the examples, it's a good demonstration of this mechanism.

lotif:
@sanaAyrml I'm already halfway through changing it, the commit is coming soon :) thanks

lotif:
Thanks for making the changes @sanaAyrml ! I made a small fix there to make sure passing `None` to it works. Please review again. cc @emersodb

emersodb (commented):
I think these changes look good. I think the only thing that needs to be looked into is whether this "breaks" the FLamby code, as I believe Sana also included seed setting in that part of our code base (under the "research" folder)

lotif:
I think FLamby is good, I don't see any mentions of seed there before or after this change. Can you confirm is all good, @sanaAyrml ?

sanaAyrml:
> I think FLamby is good, I don't see any mentions of seed there before or after this change. Can you confirm is all good, @sanaAyrml ?

I have not put up a PR to changes in flamby yet, so we are good for now. I will change them accordingly before putting up PR for it.

sanaAyrml (approved):
Changes look fine for me.

PR #84, opened 2023-12-14
Fix random seed on smoke tests and add asserts on results

lotif:
# PR Type
Other

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686mur37

- Adding `--seed` parameter to APFL and SCAFFOLD examples
- Adding the ability to set that parameter on the smoke tests
- Setting the seed parameter for FedProx, APFL and SCAFFOLD
- Adding functionality to check the logs for the metrics

Because of fluctuations in the results depending on the machine it runs, I had to parse the logs to check for the values with some tolerance. And because of that, the code turned out to be more complicated and fragile than I think it should.

That being said, I don't know if this solution should be merged. A better solution would be to save the results to a json file that can be checked more easily by the tests, it would make the code a lot more simple and robust.

# Tests Added

The smoke tests themselves


examples/apfl_example/client.py
```diff
@@ -15,6 +15,7 @@
 from fl4health.model_bases.apfl_base import ApflModule
 from fl4health.utils.load_data import load_mnist_data
 from fl4health.utils.metrics import Accuracy
+from fl4health.utils.random import set_all_random_seeds
```
emersodb, line 18: I know this is "out-of-scope" for this PR, but do you mind updating the typing annotation on `set_all_random_seeds`? Right now the seed argument is annotated as an `int`, but we really mean `Optional[int] `

examples/apfl_example/client.py
```diff
@@ -40,11 +41,21 @@ def get_criterion(self, config: Config) -> _Loss:
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="FL Client Main")
     parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
+    parser.add_argument(
+        "--seed",
+        action="store",
+        type=int,
+        help="Seed for the random number generator",
```
emersodb, line 48: Super minor: This help string, and the one below, could be slightly more informative. That is, something like "Seed for the random number generators across python, torch, and numpy"

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+        has_loss = any([approx(loss, abs=self.tolerance) == self.loss for loss in losses])
+
+        if has_loss:
+            self.found_loss = True
+
+    def parse_losses(self, log_line: str) -> List[float]:
+        try:
+            if self.loss_type == LossType.LOSS:
+                keyword = "losses_distributed"
+                if keyword in log_line:
```
emersodb, line 95: Some comments on the parsing might make the code faster to understand.

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+            else:
+                keyword = f"{self.loss_type.value}:"
+                if keyword in log_line:
+                    after_keyword = log_line.split(keyword)[-1]
+                    loss_string = after_keyword[: after_keyword.find("\t")]
+                    return [eval(loss_string)]
+        except Exception:
+            # if any parsing error occurs, return empty
+            return []
+        return []
```
emersodb, line 108: Is this final return necessary? It might be, but I would have thought the try and except completed all possible return routes.

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+        self.tolerance = tolerance
+        self.found_accuracy = False
+
+    def should_check(self, log_line: str) -> bool:
+        if self.metric_type == MetricType.TRAINING:
+            has_metrics = "metrics_distributed_fit" in log_line
+        else:
+            has_metrics = "metrics_distributed" in log_line
+
+        has_metrics |= f"Client {self.metric_type.value} Metrics" in log_line
```
emersodb, line 139: Interesting. I learned a new operator today 🙂

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+        accuracy: float,
+        metric_type: MetricType,
+        scope: MetricScope = MetricScope.PREDICTION,
+        tolerance: float = DEFAULT_TOLERANCE,
+    ):
+        self.metric_type = metric_type
+        self.accuracy = accuracy
+        self.scope = scope
+        self.tolerance = tolerance
+        self.found_accuracy = False
```
emersodb, line 131: Super minor: You could make this property something more generic like `found_target` and put it in the `MetricChecker` base class and move it out of the child implementations.

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+            if keyword not in log_line:
+                return []
+
+            if self.metric_type == MetricType.TRAINING:
+                accuracy_type_keyword = f"train - {self.scope.value} - accuracy"
+            else:
+                accuracy_type_keyword = f"val - {self.scope.value} - accuracy"
+
+            if multiple_accuracies:
+                accuracies_string = log_line.split(keyword)[-1]
```
emersodb, line 175: Maybe we can encapsulate this parsing functionality for multiple or single accuracies into functions themselves for readability?

tests/smoke_tests/run_smoke_test.py
```diff
@@ -312,6 +345,28 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
             client_python_path="examples.fedprox_example.client",
             config_path="tests/smoke_tests/fedprox_config.yaml",
             dataset_path="examples/datasets/mnist_data/",
+            seed=42,
+            client_metrics_checkers=[
+                LossChecker(1.0138, LossType.CHECKPOINT, MetricType.TRAINING),
+                LossChecker(1.0138, LossType.BACKWARD, MetricType.TRAINING),
+                LossChecker(0.0000, LossType.PROXIMAL, MetricType.TRAINING),
```
emersodb, line 352: Based on the config for the smoke tests it looks like the initial proximal weight is set to 0.0. If we're only running one server round, this means that the loss contribution is sort of "ignored." For the smoke tests, I'd recommend we set it to something like 0.01 so that this loss is non-zero. Let me know if you think that would be a problem.

emersodb (commented):
I see what you mean by the log parsing being somewhat convoluted, but I do think you did a nice job "wrangling" that parsing into something fairly clean. Parsing the logs is definitely not ideal and producing a json with the results would definitely be cleaner. 

I think this could be accomplished fairly smoothly by adding an optional module to the base clients and servers that handles dumping metrics etc. This would be similar to the `wandb_reporter` that handles sending metrics etc. to Weights and Biases.

I'm good to do that now, and not merge this PR, if you want to do it that way. Alternatively, we can merge this PR, as it may be a bit fragile, but works and then have a ticket to move to the json file functionality instead that we work on afterwards.

Let me know what you think.

tests/smoke_tests/checkers.py
```diff
@@ -0,0 +1,206 @@
...
+        if self.loss_type == LossType.LOSS:
+            return "losses_distributed" in log_line
+
+        # all other loss types:
+        assert self.metric_type is not None
+        return f"Client {self.metric_type.value} Losses" in log_line
+
+    def check(self, log_line: str) -> None:
+        losses = self.parse_losses(log_line)
+        has_loss = any([approx(loss, abs=self.tolerance) == self.loss for loss in losses])
```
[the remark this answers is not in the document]
lotif: If we're running consistently on the same machine with the same config (like, one machine locally), the tolerance is not really necessary. However, if other people run it on different processors, especially on different OSs (Windows vs MacOS vs Linux), the results will change slightly. I have even seen it change a little if it's run on different Tensorflow or Pytorch versions (due to differences in library compilation in C). So we have to use a tolerance in order to keep the tests as low-maintenance as possible.

lotif:
@emersodb I think it's better to do it the right way if it's not too hard. The test code would be so much simpler and resemble this code very little, most of it would be trashed either way. I'm also a bit afraid of this test being too fragile and giving us issues in the meantime, so I'm closing this PR and will start working on the reporter part first before coming back to this.

PR #85, opened 2023-12-15
Add warm up module

sanaAyrml:
# PR Type
[Feature]

# Short Description

Added a new Python file as `warmed_up_module.py` under the `preprocessing` directory. This module can be instantiated in the `__init__ `function of each client and used to match and load weights from a pre-trained model. To enable the loading of weights during the initial initialization, users need to override the `initialize_all_model_weights` function. After loading the server model weights, the warm-up module allows us to load weights from the pretrained model over a subset or all of the layers.

`warmed_up_module.py` takes a path to the pertained model and an optional path to the JSON file with mapping among the stats of the current model and the pertained models' stats. 
If the path to the JSON file is None, the module loads any stats with the same name; however, if weights_mapping is provided, it matches current stats with pertained model stats based on this mapping and loads weights.

Also, sample use cases of this module are implemented in warm_up_example.

Clickup Ticket(s): [Link](https://app.clickup.com/t/8685ea66a) 

# Tests Added

Tests were added to check whether loading using and not using weights_mapping is done correctly.


fl4health/clients/clipping_client.py
```diff
@@ -30,13 +31,15 @@ def __init__(
         device: torch.device,
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
         checkpointer: Optional[TorchCheckpointer] = None,
+        warmed_up_module: Optional[WarmedUpModule] = None,
```
jewelltaylor, line 34: Do you have to update the set_parameters in this example? It does not call the BasicClient set_parameters so I think we have to include the logic you did in BasicClient and DynamicWeightExchangeClient in the set_parameters for this class. I could be missing something though

fl4health/utils/warmed_up_module.py
```diff
@@ -0,0 +1,90 @@
...
+            pretrained_model_dir (Optional[str]): Directory of the pretrained model
+            weights_mapping_dir (Optional[str], optional): Directory of to json file of the weights mapping dict.
+            Defaults to None.
+            If models are not exactly the same, a weights mapping dict is needed to map the weights of the pretrained
+            model to the current model.
+        """
+
+        if pretrained_model_dir is None or not os.path.exists(pretrained_model_dir):
+            log(INFO, "No pretrained model provided")
+            self.pretrained_model_state = None
```
jewelltaylor, line 26: Should we just require pretrained_model_dir as argument and throw an error if we cant load the model? We must have a pretrained model to load if we are using this. Let me know if I am missing something.

fl4health/utils/warmed_up_module.py
```diff
@@ -0,0 +1,90 @@
...
+        current_model_state = model.state_dict()
+
+        matching_state = {}
+        for key in current_model_state.keys():
+            pretrained_key = self.get_matching_component(key)
+            log(INFO, f"Matching: {key} -> {pretrained_key}")
+            original_state = current_model_state[key]
+            if pretrained_key in self.pretrained_model_state.keys():
+                pretrained_state = self.pretrained_model_state[pretrained_key]
+            else:
```
jewelltaylor, line 81: Should we log when we don't use a weight from the pretrained model and resort to using the weight in the current model?

fl4health/utils/warmed_up_module.py
```diff
@@ -0,0 +1,90 @@
...
+                continue
+            if original_state.size() == pretrained_state.size():
+                matching_state[key] = pretrained_state
+            else:
+                log(ERROR, f"Dismatched sizes {original_state.size()} -> {pretrained_state.size()}")
+        log(INFO, f"!!! {len(matching_state)}/{len(current_model_state)} STATS GOT MATCHED !!!")
```
jewelltaylor, line 87: Maybe sticking with the convention we have this far, we can drop the exclamation marks and the caps

fl4health/utils/warmed_up_module.py
```diff
@@ -0,0 +1,90 @@
...
+
+        Returns:
+            Optional[str]: If no weights mapping dict is provided, returns the key. Otherwise, if the key is in the
+            weights mapping dict, returns the matching component of the key. Otherwise, returns None.
+        """
+
+        if self.weights_mapping_dict is None:
+            return key
+
+        components = key.split(".")
```
jewelltaylor, line 52: I am kind of confused why we are splitting the key (corresponding to current model) by dot and reconstructing it by appending component by component and checking each time if the key is present in the weight_mapping_dict. Shouldn't the key be an entry in the dictionary? For example, in the fenda example you provided the config for we would iterate through the keys of the Fenda model. One the keys would be "local_module.conv1" and that could be mapped to "conv1" using the mapping. We dont need to break it apart and reconstruct it piece by piece. 

I may be missing something so please feel free to correct me :) If this is the case, maybe a comment clarifying what is happening here would be helpful

jewelltaylor (commented):
The overall design looks good to me! I have requested some small changes and requested some clarification for certain parts of the code.

Those comments should not involve major changes, so you can go ahead with the tests. It may be a pain, but we should have a good amount of tests for different FL methods and a few should have complex model architectures just to make sure there is no edge cases in the weight mapping that we are not accounting for.

examples/fedprox_example/client.py
```diff
@@ -53,6 +54,20 @@ def get_criterion(self, config: Config) -> _Loss:
...
+        action="store",
+        type=str,
+        help="Path to the pretrained model",
+        required=False,
+    )
+    parser.add_argument(
+        "--weights_mapping_dir",
+        action="store",
+        type=str,
+        help="Path to the weights mapping file",
```
lotif, line 68: That help text in inconsistent with the argument name. Either change the help text to say "files" or "directory" instead of a single file, change the argument name to `--weights_mapping_file`, or something else along those lines.

examples/fenda_example/client.py
```diff
@@ -55,11 +64,41 @@ def get_criterion(self, config: Config) -> _Loss:
...
+        action="store",
+        type=int,
+        help="Seed for the random number generator",
+        required=False,
+    )
+    parser.add_argument(
+        "--weights_mapping_dir",
+        action="store",
+        type=str,
+        help="Path to the weights mapping file",
```
lotif, line 85: Same here.

fl4health/clients/basic_client.py
```diff
@@ -79,6 +81,7 @@ def __init__(
         self.num_train_samples: int
         self.num_val_samples: int
         self.learning_rate: Optional[float] = None
+        self.warmed_up_module = warmed_up_module
```
lotif, line 84: Super minor optional nitpick, but I think it's better organized if this line goes before the line above just to keep all the param setting lines together and the other variables that are not init params at the end.

fl4health/utils/warmed_up_module.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import Optional
+
+import torch
+from flwr.common.logger import log
+
+
+class WarmedUpModule:
+    """This class is used to load a pretrained model into the current model."""
+
+    def __init__(self, pretrained_model_dir: Optional[str], weights_mapping_dir: Optional[str] = None) -> None:
```
lotif, line 13: Add `= None` to `pretrained_model_dir` for consistency.

fl4health/utils/warmed_up_module.py
lotif: my 2 cents: I think keeping it optional is fine but we should definitely fail if it's passed in and not be able to load.

fl4health/utils/warmed_up_module.py
lotif: And I think this looks more like a debug level message than an info level message? Please correct me if I'm wrong.

fl4health/clients/clipping_client.py
sanaAyrml: I changed the place of applying warmup model to initialize_all_model_weights function, so we would prevent from code duplication in files.

fl4health/utils/warmed_up_module.py
sanaAyrml: I agree with both that people should just use it if they actually have a model dir, and in fact I used to had an assertion if no address is provided. The reason that I changed temporarily to this version is to go around smoke test failing in that case. I will fix it when I implement and change test for warmup.

fl4health/utils/warmed_up_module.py
sanaAyrml: Sure that can be helpful, I updated logging based on your comment.

fl4health/utils/warmed_up_module.py
sanaAyrml: Ok, so this function makes module to also support partial mapping. For example now the user can have a mapping such as {"model1":"global_module"}, in that case the model can map anything starting with "model1" to keys starting with "global_module". This is so useful for bigger models which user doesn't one to tell layer by layer mapping.

sanaAyrml: So the idea is that if model gets a key like "model1.conv1.weights" first it separates it and starting from beginning it tries to find a match for "model1", "model1.conv1" and "model1.conv1.weights" in the provided mapping. If it finds a mapping for "model1", it replaces  "model1" in the current key with that mapping.

sanaAyrml: I updated doc string to be more clear, but let me know if it is still confusing.

examples/fedprox_example/client.py
sanaAyrml: Good catch, I will fix it.

fl4health/utils/warmed_up_module.py
sanaAyrml: Again, initially I am thinking to not make it optional at all, but this is just go around for smoke test. Once I change it to support getting address for pretrained model in its args, I will fix this.

lotif: Yeah, it actually makes more sense for it to not be optional.

fl4health/clients/basic_client.py
lotif: Sorry, I overlooked how big this init is in the first place. What I meant by the comment is to put this statement below line 57. But again, minor nitpick.

fl4health/clients/basic_client.py
```diff
@@ -139,14 +143,19 @@ def set_parameters(self, parameters: NDArrays, config: Config) -> None:
 
     def initialize_all_model_weights(self, parameters: NDArrays, config: Config) -> None:
         """
-        If this is the first time we're initializing the model weights, we use the FullParameterExchanger to
-        initialize all model components
+        If this is the first time we're initializing the model weights, we check if the client is using a warmed up
+        module or not. If the client is using a warmed up module and the pretrained model state is loaded, we load
+        the pretrained model state into the current model. Otherwise, we use the FullParameterExchangerto initialize
+        all model.
```
lotif, line 149: missing `components` at the end of this.

fl4health/clients/basic_client.py
sanaAyrml: That makes sense.

fl4health/utils/warmed_up_module.py
jewelltaylor: Ah makes a lot of sense! I think the documentation you provided explains it great :)

fl4health/utils/warmed_up_module.py
jewelltaylor: Updated code looks great!

examples/models/fenda_cnn.py
```diff
@@ -24,9 +24,9 @@ def head_forward(self, input_tensor: torch.Tensor) -> torch.Tensor:
 class LocalCnn(nn.Module):
     def __init__(self) -> None:
         super().__init__()
-        self.conv1 = nn.Conv2d(1, 6, 5)
+        self.conv1 = nn.Conv2d(1, 8, 5)
         self.pool = nn.MaxPool2d(2, 2)
-        self.conv2 = nn.Conv2d(6, 16, 5)
```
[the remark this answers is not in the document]
sanaAyrml: I did that to models be compatible with `MnistNet` in `examples/models/cnn_model.py` so we can load the pretrained weights from fedavg to fenda.

examples/warm_up_example/fedavg_warm_up/client.py
```diff
@@ -0,0 +1,90 @@
...
+        required=False,
+    )
+    parser.add_argument(
+        "--checkpoint_dir",
+        action="store",
+        type=str,
+        help="Path to the directory where the checkpoints are stored",
+        required=True,
+    )
+    parser.add_argument(
```
emersodb, line 64: I don't mind having this argument. However, if you're setting the random seed, each client actually generates a unique hash when it starts up under the property `self.client_name` using the `generate_hash` function. That should be deterministic with the random seed and you could eliminate this argument. I don't care either way, just wanted to mention it as a potential simplification.

examples/warm_up_example/fedavg_warm_up/client.py
```diff
@@ -0,0 +1,90 @@
...
+        "--client_number",
+        action="store",
+        type=int,
+        help="Client number",
```
emersodb, line 68: This help string could be slightly more informative. Perhaps something like "Unique identifier for the client to be used for saving artifacts"?

examples/warm_up_example/fedavg_warm_up/client.py
```diff
@@ -0,0 +1,90 @@
...
+        action="store",
+        type=str,
+        help="Server Address for the clients to communicate with the server through",
+        default="0.0.0.0:8080",
+    )
+    parser.add_argument(
+        "--seed",
+        action="store",
+        type=int,
+        help="Seed for the random number generator",
```
emersodb, line 54: Super minor: This help string, and the one below, could be slightly more informative. That is, something like "Seed for the random number generators across python, torch, and numpy"

examples/warm_up_example/fedavg_warm_up/server.py
```diff
@@ -0,0 +1,128 @@
...
+        action="store",
+        type=str,
+        help="Server Address to be used to communicate with the clients",
+        default="0.0.0.0:8080",
+    )
+    parser.add_argument(
+        "--seed",
+        action="store",
+        type=int,
+        help="Seed for the random number generator",
```
emersodb, line 117: Super Minor: Perhaps a similar help string as the client argument suggestion above?

examples/warm_up_example/warmed_up_fedprox/README.md
```diff
@@ -0,0 +1,40 @@
...
+* `batch_size`: size of the batches each client will train on
+* `n_server_rounds`: The number of rounds to run FL
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the three
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.warm_up_example.warmed_up_fedprox.client --dataset_path /path/to/data --seed "SEED" --pretrained_model_dir /path/to/checkpointing/directory  --weights_mapping_path /path/to/weights/mapping/file --client_number "CLIENT_NUMBER (1, 2, or 3)"
+```
```
emersodb, line 35: Perhaps have a comment reminding the user that they should have the "seed" argument match that of the warmup to ensure consistent dataset splits? I know it's mentioned above, but a reminder close to the example command might also be helpful. Something like 
```
**NOTE**: "SEED" above should match that of the warm up run to ensure the datasets are split consistently.
```

examples/warm_up_example/warmed_up_fedprox/client.py
```diff
@@ -0,0 +1,131 @@
...
+    set_all_random_seeds(args.seed)
+
+    # Start the client
+    client = MnistFedProxClient(
+        data_path,
+        [Accuracy()],
+        DEVICE,
+        args.client_number,
+        Path(args.pretrained_model_dir),
+        Path(args.weights_mapping_path),
```
emersodb, line 126: This is fairly minor, but we have `weights_mapping_path` as an optional argument above. However, will this Path parse work if the arg isn't included? That is, if the arg isn't provided, I think `args.weights_mapping_path` will either fail or be `None`. If it's `None`, I think `Path(None)` will throw an error. Keeping it optional is good with me, but we should probably do something like
```
weights_mapping_path = Path(args.weights_mapping_path) if args.weights_mapping_path else None
```
and pass `weights_mapping_path` to `MnistFedProxClient`.

The arg `pretrained_model_dir` is also marked as `required=False`, but for this example we need it to be defined. So I'd say that we mark it as `required=True`?

examples/warm_up_example/warmed_up_fedprox/server.py
```diff
@@ -0,0 +1,132 @@
...
+from fl4health.utils.config import load_config
+from fl4health.utils.random import set_all_random_seeds
+
+
+def get_initial_model_information() -> Parameters:
+    # Initializing the model parameters on the server side.
+    # Currently uses the Pytorch default initialization for the model parameters.
+    initial_model = MnistNet()
+    model_weights = [val.cpu().numpy() for _, val in initial_model.state_dict().items()]
+    return ndarrays_to_parameters(model_weights)
```
emersodb, line 27: To make it clear that we're not using the server-side initial parameters, perhaps we can simply do something like
```
def get_initial_model_information() -> Parameters:
    # Server-side initial weights are not used for client side models when using a warmed up module. So these can be initialized as empty
    return ndarrays_to_parameters([])
```

Note: I think this will work, but I haven't tested it. 😂

examples/warm_up_example/warmed_up_fenda/README.md
```diff
@@ -0,0 +1,34 @@
...
+* `batch_size`: size of the batches each client will train on
+* `n_server_rounds`: The number of rounds to run FL
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the three
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.warm_up_example.warmed_up_fenda.client --dataset_path /path/to/data --seed "SEED" --pretrained_model_dir /path/to/checkpointing/directory  --weights_mapping_path /path/to/weights/mapping/file --client_number "CLIENT_NUMBER (1, 2, or 3)"
+```
```
emersodb, line 29: A note along the lines described in the fedprox md file above would be useful here 🙂

examples/warm_up_example/warmed_up_fenda/client.py
```diff
@@ -0,0 +1,134 @@
...
+    set_all_random_seeds(args.seed)
+
+    # Start the client
+    client = MnistFendaClient(
+        data_path,
+        [Accuracy("accuracy")],
+        DEVICE,
+        args.client_number,
+        Path(args.pretrained_model_dir),
+        Path(args.weights_mapping_path),
```
emersodb, line 129: Similar to the FedProx example, I think we can make a few small refactors to these arguments to make them a bit more graceful. Let me know if you have any objections.

examples/warm_up_example/warmed_up_fenda/server.py
```diff
@@ -0,0 +1,126 @@
...
+from fl4health.server.base_server import FlServer
+from fl4health.utils.config import load_config
+from fl4health.utils.random import set_all_random_seeds
+
+
+def get_initial_model_parameters() -> Parameters:
+    # Initializing the model parameters on the server side.
+    # Currently uses the Pytorch default initialization for the model parameters.
+    initial_model = FendaModel(LocalCnn(), GlobalCnn(), FendaClassifier(FendaJoinMode.CONCATENATE))
+    return ndarrays_to_parameters([val.cpu().numpy() for _, val in initial_model.state_dict().items()])
```
emersodb, line 27: Similar to the FedProx example, I think we can make it very explicit here that we don't use the initial parameters sent by the serve by setting the parameters constructed by this function to "empty"

examples/warm_up_example/warmed_up_fenda/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+After the warm-up training, clients can load their warmed-up models and continue training with the FENDA algorithm. To maintain consistency in the data loader between both runs, it is crucial to set a fixed seed for both clients and the server, ensuring uniformity in random data points across consecutive runs. Therefore, we ensure a fixed seed is set for these consecutive runs in both the `client.py` and `server.py` files. Additionally, to load the warmed-up models, it's important provide the path to the pretrained models and unique number of each client. In this particular scenario, we assigned a unique number between 1 and 3 to each client, ensuring that we can load the trained local model for each client from the previous example as a warmed-up model. Since models in the two runs can be different, loading weights from the pretrained model requires providing a mapping between the pretrained model and the model used in FL training. This mapping is accomplished through the `weights_mapping.json` file, which contains the names of the pretrained model's layers and the corresponding names of the layers in the model used in FL training.
```
emersodb, line 6: Perhaps we can also describe what we're doing specifically in this example. That is, we're taking the components of a single model trained with FedAvg and loading the weights into both the local and global submodules of a FENDA model.

fl4health/clients/basic_client.py
```diff
@@ -122,17 +123,21 @@ def get_parameters(self, config: Config) -> NDArrays:
...
         Args:
             parameters (NDArrays): Parameters have information about model state to be added to the relevant client
                 model but may contain more information than that.
             config (Config): The config is sent by the FL server to allow for customization in the function if desired.
         """
         assert self.model is not None
-        if not self.model_weights_initialized:
-            self.initialize_all_model_weights(parameters, config)
+
+        if not self.first_parameters_set:
```
emersodb, line 137: I think there is a small edge case that we might not be thinking about here when the warm up module is only a subset of the weights being initialized. Consider a simple example, such as FedProx, but we only load the first layer of the warm up module into the FedProx model to be trained. The rest of the FedProx model will be initialized differently per client (unless the seeds are well set etc.), which isn't what we want. 

One way we could get around this is to initialize the model with the server weights first, then load the warm up model to replace the specified layers. We could have the `BasicClient` accept an optional `WarmedUpModule` and if it's there, we perform a load with the class. This method would then be something like
```
if not self.model_weights_initialized:	
    self.initialize_all_model_weights(parameters, config)
    if self.warmed_up_module:
        self.warmed_up_module.load_from_pretrained(self.model)
...
```
We can talk about this a bit more if you'd like. This would mean we wouldn't need `first_parameters_set` anymore.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+import torch
+from flwr.common.logger import log
+
+
+class WarmedUpModule:
+    """This class is used to load a pretrained model into the current model."""
+
+    def __init__(
+        self,
+        pretrained_model: Optional[torch.nn.Module] = None,
```
emersodb, line 16: Very optional, but if you want to shorten these types (`torch.nn.Module`), we have been doing
```
import torch.nn as nn
```
in other files, which allows you to type these as just `nn.Module`. It's fine as is, just wanted to let you know in case you wanted to.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+from logging import INFO
+from pathlib import Path
+from typing import Optional
+
+import torch
+from flwr.common.logger import log
+
+
+class WarmedUpModule:
+    """This class is used to load a pretrained model into the current model."""
```
emersodb, line 12: Super minor, but I might say `into a target model` since we provide a target into which, we insert the weights.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+            if i == 0:
+                matching_component = components[0]
+            else:
+                matching_component += "." + component
+            if matching_component in self.weights_mapping_dict:
+                return self.weights_mapping_dict[matching_component] + key[len(matching_component) :]
+        return None
+
+    def load_from_pretrained(self, model: torch.nn.Module) -> torch.nn.Module:
+        """Load the pretrained model into the current model.
```
emersodb, line 90: Super minor, but I might say `into a target model` since we provide a target into which, we insert the weights.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+            log(INFO, f"Loading pretrained model from {pretrained_model_path}")
+            self.pretrained_model_state = torch.load(pretrained_model_path).state_dict()
+
+        else:
+            raise AssertionError("At least one of pretrained_model_path and pretrained_model should be provided.")
+
+        if weights_mapping_path is not None:
+            with open(weights_mapping_path, "r") as file:
+                self.weights_mapping_dict = json.load(file)
+                log(INFO, f"Weights mapping dict: {self.weights_mapping_dict}")
```
emersodb, line 53: When the mapping is quite large (think language model), this is going to be a lot to log. Perhaps we just log that it was successfully loaded? If we think it's helpful to log the dictionary, perhaps we can log up to small number of components, like 5-10?

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            log(INFO, "Weights mapping dict is not provided. Matching stats directlly, based on current model's keys.")
```
emersodb, line 55: directlly -> directly 🙂

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+
+        if self.weights_mapping_dict is None:
+            return key
+
+        components = key.split(".")
+
+        for i, component in enumerate(components):
+            if i == 0:
+                matching_component = components[0]
```
emersodb, line 82: Super minor: You already have `components[0]` in `component` here, so no need to index just have `matching_component = component`

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+            else:
+                matching_component += "." + component
```
emersodb, line 84: Perhaps a more explicit way of specifying the format might be 
`matching_component = f"{matching_component}.{component}"`. However, I'm alright with sticking with the += if you'd like.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+            pretrained_key = self.get_matching_component(key)
+            log(INFO, f"Matching: {key} -> {pretrained_key}")
+            if pretrained_key is not None:
+                if pretrained_key in self.pretrained_model_state.keys():
+                    pretrained_state = self.pretrained_model_state[pretrained_key]
+                    if original_state.size() == pretrained_state.size():
+                        matching_state[key] = pretrained_state
+                        log(INFO, "Succesful stats matching.")
+                    else:
+                        log(INFO, f"Dismatched sizes {original_state.size()} -> {pretrained_state.size()}.")
```
emersodb, line 113: Small typo: Dismatched vs. Mismatched.

I would have this log be a WARNING level log, as it means the key was matched but something happened with the state. I would also log that this results in that the `pretrained_state` not being loaded. That is, we skip it so the user knows that it will be the effect.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+                else:
+                    log(INFO, f"Key {pretrained_key} not found in the pretrained model stats.")
```
emersodb, line 115: I would have this as a WARNING level log as well and state that the will not be overwritten.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+            original_state = current_model_state[key]
+
+            pretrained_key = self.get_matching_component(key)
+            log(INFO, f"Matching: {key} -> {pretrained_key}")
+            if pretrained_key is not None:
+                if pretrained_key in self.pretrained_model_state.keys():
+                    pretrained_state = self.pretrained_model_state[pretrained_key]
+                    if original_state.size() == pretrained_state.size():
+                        matching_state[key] = pretrained_state
+                        log(INFO, "Succesful stats matching.")
```
emersodb, line 111: state matching rather than stats?

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+class WarmedUpModule:
+    """This class is used to load a pretrained model into the current model."""
+
+    def __init__(
+        self,
+        pretrained_model: Optional[torch.nn.Module] = None,
+        pretrained_model_path: Optional[Path] = None,
+        weights_mapping_path: Optional[Path] = None,
+    ) -> None:
+        """Initialize the WarmedUpModule with the pretrained model stats and weights mapping dict.
```
emersodb, line 20: states instead of stats?

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,121 @@
...
+                    pretrained_state = self.pretrained_model_state[pretrained_key]
+                    if original_state.size() == pretrained_state.size():
+                        matching_state[key] = pretrained_state
+                        log(INFO, "Succesful stats matching.")
+                    else:
+                        log(INFO, f"Dismatched sizes {original_state.size()} -> {pretrained_state.size()}.")
+                else:
+                    log(INFO, f"Key {pretrained_key} not found in the pretrained model stats.")
+
+        log(INFO, f"{len(matching_state)}/{len(current_model_state)} stats got matched.")
```
emersodb, line 117: states rather than stats here?

tests/preprocessing/test_warm_up_module.py
```diff
@@ -0,0 +1,161 @@
+import copy
+
+from fl4health.model_bases.apfl_base import ApflModule
+from fl4health.model_bases.fenda_base import FendaModel
+from fl4health.model_bases.moon_base import MoonModel
+from fl4health.preprocessing.warmed_up_module import WarmedUpModule
+from tests.test_utils.models_for_test import FeatureCnn, FendaHeadCnn, HeadCnn, SmallCnn, ToyConvNet, ToyConvNet_2
+
+
+def test_warm_up_module_loading_similar_models_without_mapping() -> None:
```
emersodb, line 10: Very minor: I really like the descriptive names of these tests. However, just to make them a little shorter, I think you can omit the `_warm_up_module` part of them since we're in the `test_warm_up_module.py` file.

tests/preprocessing/test_warm_up_module.py
```diff
@@ -0,0 +1,161 @@
...
+            assert (model.state_dict()[key] == pretrained_model.state_dict()[key]).all()
+        else:
+            assert (model.state_dict()[key] == old_model.state_dict()[key]).all()
+
+
+def test_global_warm_up_module_loading_fenda_model_with_mapping() -> None:
+    pretrained_model = SmallCnn()
+    model = FendaModel(FeatureCnn(), FeatureCnn(), FendaHeadCnn())
+    old_model = copy.deepcopy(model)
+    print(pretrained_model.state_dict().keys())
```
emersodb, line 77: I'd say either drop these print statements or turn them into logs?

tests/preprocessing/test_warm_up_module.py
```diff
@@ -0,0 +1,161 @@
...
+from fl4health.model_bases.moon_base import MoonModel
+from fl4health.preprocessing.warmed_up_module import WarmedUpModule
+from tests.test_utils.models_for_test import FeatureCnn, FendaHeadCnn, HeadCnn, SmallCnn, ToyConvNet, ToyConvNet_2
+
+
+def test_warm_up_module_loading_similar_models_without_mapping() -> None:
+    pretrained_model = ToyConvNet()
+    model = ToyConvNet()
+    old_model = copy.deepcopy(model)
+    warmupmodule = WarmedUpModule(pretrained_model=pretrained_model)
```
emersodb, line 14: Really minor: code style conventions typically would have this named `warmup_module` for readability throughout the test file.

emersodb (commented):
Overall, I think this looks great and you did a great job with writing tests! I think the only thing we should consider is the edge case when we don't initialize all of the weights with a warm up model.

examples/warm_up_example/fedavg_warm_up/client.py
sanaAyrml: That's a good point. I was following flabby experiments for this argument, but I will update based on this comment and remove extra arguments.

fl4health/preprocessing/warmed_up_module.py
sanaAyrml: I think we can omit it.

fl4health/preprocessing/warmed_up_module.py
sanaAyrml: Thanks, updated.

tests/preprocessing/test_warm_up_module.py
sanaAyrml: Oops they are extra. Should have been omitted.

examples/warm_up_example/fedavg_warm_up/client.py
emersodb: Yeah I see what you mean! The reason the FLamby experiments use it is because the clients numbers specifically correspond to the data loaders. So we explicitly needed to set the numbers. Again, no worries if you want to leave it though.

fl4health/preprocessing/warmed_up_module.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            log(INFO, "Weights mapping dict is not provided. Matching states directly, based on target model's keys.")
+            self.weights_mapping_dict = None
+
+    def get_matching_component(self, key: str) -> Optional[str]:
+        """Get the matching component of the key from the weights mapping dictionary. Since the provided mapping
+        can contain partial names of the keys, this function is used to split the key of the target model and
+        match it with the partial key in the mapping, returning the complete name of the key in the pretrained model.
+
+        This allows users to provide one mapping for multiple statistics that share the same prefix. For example,
```
yc7z, line 62: Do you mean "model states" instead of statistics?

tests/preprocessing/test_warm_up_module.py
```diff
@@ -0,0 +1,200 @@
...
+def test_loading_different_models_without_mapping() -> None:
+    pretrained_model = SmallCnn()
+    model = ToyConvNet()
+    old_model = copy.deepcopy(model)
+    warmup_module = WarmedUpModule(pretrained_model=pretrained_model)
+    warmup_module.load_from_pretrained(model)
+
+    # Check if the weights are same with previous model as loading should not have any effect
+    for old_model_value, model_value in zip(old_model.state_dict().values(), model.state_dict().values()):
+        assert (old_model_value == model_value).any()
```
yc7z, line 80: I think you want to do

`assert (old_model_value == model_value).all()`

instead.

yc7z (commented):
Overall the design looks great to me! I just left a few pretty minor comments.

tests/preprocessing/test_warm_up_module.py
sanaAyrml: That's true. Thanks for mentioning.

fl4health/preprocessing/warmed_up_module.py
sanaAyrml: Yeah I think states make more sense.

examples/warm_up_example/warmed_up_fedprox/client.py
```diff
@@ -52,21 +51,20 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
...
-        return model
+        return MnistNet().to(self.device)
 
     def get_optimizer(self, config: Config) -> Optimizer:
         return torch.optim.AdamW(self.model.parameters(), lr=0.01)
 
     def get_criterion(self, config: Config) -> _Loss:
         return torch.nn.CrossEntropyLoss()
 
+    # Overriding the initialize_all_model_weights function is required to load the pretrained model
```
emersodb, line 62: This is a nice way of doing the warm up load!

emersodb (approved):
I'm happy with all of the changes that you made in response to my comments. I think there are a few small unresolved comments from John J and Marcelo above w.r.t smoke tests or the like. So I'd just make sure they're good with any of the related changes before merging.

PR #86, opened 2023-12-15
Flamby experiments

sanaAyrml:
# PR Type
[ Other ]

# Short Description

Implementation of FedPer, PerFCL, and Moon experiments with setting fixed seed for fed_heart_disease, fed_isic2019, and fedixi. I also updated the fenda experiments by setting the fixed seed. Finally, as FedPer by default uses moon client, it contains auxiliary contrastive loss; I added an option to be able to turn off this loss for moon client.


fl4health/model_surgery/load_pretrained_weights.py
emersodb, line 1: Perhaps this file is included accidentally, as it is empty?

research/flamby/fed_heart_disease/fedper/client.py
```diff
@@ -0,0 +1,141 @@
...
+        self,
+        data_path: Path,
+        metrics: Sequence[Metric],
+        device: torch.device,
+        client_number: int,
+        learning_rate: float,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        checkpointer: Optional[TorchCheckpointer] = None,
+    ) -> None:
+        super().__init__(
```
emersodb, line 37: Maybe add a comment here that states that the Moon loss weight defaults to zero?

research/flamby/fed_heart_disease/fedper/fedper_model.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.linear(x)
+        x = F.relu(x)
+        x = x.flatten(start_dim=1)
+        return x
+
+
+class HeadClassifier(nn.Module):
+    """Moon head module"""
```
emersodb, line 23: Moon -> FedPer in the comment (to avoid confusion 🙂)

research/flamby/fed_heart_disease/fedper/client.py
```diff
@@ -0,0 +1,141 @@
...
+    def get_model(self, config: Config) -> nn.Module:
+        model: nn.Module = FedHeartDiseaseFedPerModel().to(self.device)
+        return model
+
+    def get_optimizer(self, config: Config) -> Optimizer:
+        return torch.optim.AdamW(self.model.parameters(), lr=self.learning_rate)
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return BaselineLoss()
+
```
emersodb, line 67: I may be missing something, so please correct me if I'm wrong, but is there a place where you define the partial weight exchanger for FedPer (similar to `examples/fedper_example/client.py`). Without that definition, I think we default to a full weight exchange approach, which isn't what we want. That is, we want only the base module to be exchanged and the classifier layer to stay local. I may be missing where that happens though 🙂

research/flamby/fed_heart_disease/fenda/client.py
```diff
@@ -50,11 +51,11 @@ def __init__(
         self.client_number = client_number
         self.learning_rate: float = learning_rate
         if cos_sim_activate:
-            self.cos_sim_loss_weight = 100.0
+            assert isinstance(extra_loss_weights, float)
```
emersodb, line 54: Rather than doing an `isinstance` check here, you can just do `assert extra_loss_weights is not None`.

research/flamby/fed_heart_disease/perfcl/README.md
```diff
@@ -0,0 +1,24 @@
...
+An example is something like
+``` bash
+./research/flamby/fed_heart_disease/perfcl/run_hp_sweep.sh \
+   research/flamby/fed_heart_disease/perfcl/config.yaml \
+   research/flamby/fed_heart_disease/perfcl/ \
+   /Users/david/Desktop/FLambyDatasets/fed_heart_disease/ \
+   /h/demerson/vector_repositories/fl4health_env/
+```
+
+In order to manipulate the grid search being conducted, you need to change the parameters for `lr`, the client-side learning rate, in the `run_hp_sweep.sh` script directly.
```
emersodb, line 24: Since there are two other hyper-parameters for PerFCL (mu and gamma), it's probably worth saying something like 
```
you need to change the parameters for `lr`, `mu`, or `gamma`, in the `run_hp_sweep.sh` script directly. These represent the client-side learning rate and the two PerFCL loss weights respectively.
```

research/flamby/fed_heart_disease/perfcl/client.py
```diff
@@ -0,0 +1,177 @@
...
+)
+from fl4health.clients.fenda_client import FendaClient
+from fl4health.utils.losses import LossMeterType
+from fl4health.utils.metrics import Accuracy, Metric
+from fl4health.utils.random import set_all_random_seeds
+from research.flamby.fed_heart_disease.perfcl.perfcl_model import FedHeartDiseasePerFCLModel
+from research.flamby.flamby_data_utils import construct_fed_heard_disease_train_val_datasets
+
+
+class FedHeartDiseasePerFCLClient(FendaClient):
```
emersodb, line 30: Super minor: For acronyms, we've adopted the style of first letter uppercased and the rest lower. So PerFCL here would be `PerFcl`. This is similar for `FedHeartDiseasePerFCLModel` as well

research/flamby/fed_heart_disease/fedper/client.py
sanaAyrml: That is true, I think I have missed that we should also implement `get_parameter_exchanger` for fedper. I will update and rerun the experiments.

research/flamby/fed_heart_disease/fenda/client.py
sanaAyrml: Sure

research/flamby/fed_isic2019/fedper/fedper_model.py
```diff
@@ -0,0 +1,83 @@
...
+class BaseEfficientNet(nn.Module):
+    """FedPer feature extractor module
+    We use the EfficientNets architecture that many participants in the ISIC
+    competition have identified to work best.
+    See here the [reference paper](https://arxiv.org/abs/1905.11946)
+    Thank you to [Luke Melas-Kyriazi](https://github.com/lukemelas) for his
+    [pytorch reimplementation of EfficientNets]
+    (https://github.com/lukemelas/EfficientNet-PyTorch).
+    When loading the EfficientNet-B0 model, we strip off the FC layer to use the model as a feature extractor.
+    We freeze a subset of the layers in order to make sure that FENDA is not training twice as many parameters as the
```
emersodb, line 48: Based on the implementation for FedPer, I don't think we're freezing any blocks. To avoid confusion, maybe we change this comment to something like
```
There is an option to freeze a subset of the layers to reduce the number of trainable parameters. However, it is not used in the FedPer experiments.
```

research/flamby/fed_isic2019/fedper/run_hp_sweep.sh
```diff
@@ -0,0 +1,58 @@
...
+  SERVER_ADDRESS="0.0.0.0:${SERVER_PORT}"
+  echo "Server Address: ${SERVER_ADDRESS}"
+  SBATCH_COMMAND="research/flamby/fed_isic2019/fedper/run_fold_experiment.slrm \
+    ${SERVER_CONFIG_PATH} \
+    ${EXPERIMENT_DIRECTORY} \
+    ${DATASET_DIR} \
+    ${VENV_PATH} \
+    ${LR_VALUE} \
+    ${SERVER_ADDRESS} \
+    ${MU_VALUE}"
```
emersodb, line 52: I think this probably still works with this here, it just doesn't get parsed and the `${MU_VALUE}` likely just returns a blank, but still worth removing to avoid confusion or unforeseen issues?

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
```diff
@@ -81,8 +83,10 @@ source ${VENV_PATH}bin/activate
 echo "Active Environment:"
 which python
 
-for RUN_NAME in "${RUN_NAMES[@]}";
+for ((i=0; i<${#RUN_NAMES[@]}; i++));
 do
+RUN_NAME="${RUN_NAMES[i]}"
```
emersodb, line 88: I'm not sure, but does the lack of indentation here cause an issue? I forget if bash scripts are whitespace sensitive (like python). Perhaps worth indenting it to be safe.

research/flamby/fed_isic2019/fenda/server.py
```diff
@@ -29,7 +30,7 @@ def main(config: Dict[str, Any], server_address: str) -> None:
     )
 
     client_manager = SimpleClientManager()
-    model = FedIsic2019FendaModel(frozen_blocks=None, turn_off_bn_tracking=False)
```
emersodb, line 32: Good call on setting this to 13 by default.

research/flamby/fed_isic2019/perfcl/run_fold_experiment.slrm
```diff
@@ -0,0 +1,166 @@
...
+echo "GPUs per node: ${NUM_GPUs}"
+
+# Source the environment
+source ${VENV_PATH}bin/activate
+echo "Active Environment:"
+which python
+
+for ((i=0; i<${#RUN_NAMES[@]}; i++));
+do
+RUN_NAME="${RUN_NAMES[i]}"
```
emersodb, line 94: As above, I don't remember if bash is white space sensitive, but I'd say we should indent this either way.

research/flamby/fed_isic2019/perfcl/run_hp_sweep.sh
```diff
@@ -0,0 +1,68 @@
...
+      SERVER_ADDRESS="0.0.0.0:${SERVER_PORT}"
+      echo "Server Address: ${SERVER_ADDRESS}"
+      SBATCH_COMMAND="research/flamby/fed_isic2019/perfcl/run_fold_experiment.slrm \
+        ${SERVER_CONFIG_PATH} \
+        ${EXPERIMENT_DIRECTORY} \
+        ${DATASET_DIR} \
+        ${VENV_PATH} \
+        ${LR_VALUE} \
+        ${SERVER_ADDRESS}\
+            ${MU_VALUE}\
```
emersodb, line 58: I'd fix the indentations on these values, even if it's just for visual presentation.

research/flamby/fed_ixi/fedper/run_hp_sweep.sh
```diff
@@ -0,0 +1,58 @@
...
+  SERVER_ADDRESS="0.0.0.0:${SERVER_PORT}"
+  echo "Server Address: ${SERVER_ADDRESS}"
+  SBATCH_COMMAND="research/flamby/fed_ixi/fedper/run_fold_experiment.slrm \
+    ${SERVER_CONFIG_PATH} \
+    ${EXPERIMENT_DIRECTORY} \
+    ${DATASET_DIR} \
+    ${VENV_PATH} \
+    ${LR_VALUE} \
+    ${SERVER_ADDRESS} \
+    ${MU_VALUE}"
```
emersodb, line 52: Probably should remove this input for the fedper experiments, given that we don't have a mu value (I don't think this effects the results, just to avoid confusion)

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
sanaAyrml: I don't get error. So it should be fine but I will fix it to make it cleaner.

research/flamby/utils.py
```diff
@@ -109,8 +109,8 @@ def evaluate_fed_isic_model(
 ) -> float:
     meter = evaluate_model_on_dataset(model, dataset, metrics, device, is_apfl)
     computed_metrics = meter.compute()
-    assert "test_meter_FedIsic2019_balanced_accuracy" in computed_metrics
-    balanced_accuracy = computed_metrics["test_meter_FedIsic2019_balanced_accuracy"]
+    assert "test_meter - prediction - FedIsic2019_balanced_accuracy" in computed_metrics
+    balanced_accuracy = computed_metrics["test_meter - prediction - FedIsic2019_balanced_accuracy"]
```
sanaAyrml, line 113: For some reason (probably due to code refactor) I needed to change these evaluation files to make it work.

research/flamby/utils.py
```diff
@@ -121,8 +121,8 @@ def evaluate_fed_heart_disease_model(
     meter = evaluate_model_on_dataset(model, dataset, metrics, device, is_apfl)
 
     computed_metrics = meter.compute()
-    assert "test_meter_FedHeartDisease_accuracy" in computed_metrics
-    accuracy = computed_metrics["test_meter_FedHeartDisease_accuracy"]
+    assert "test_meter - prediction - FedHeartDisease_accuracy" in computed_metrics
+    accuracy = computed_metrics["test_meter - prediction - FedHeartDisease_accuracy"]
```
sanaAyrml, line 125: Same here.

research/flamby/utils.py
```diff
@@ -132,8 +132,8 @@ def evaluate_fed_ixi_model(
 ) -> float:
     meter = evaluate_model_on_dataset(model, dataset, metrics, device, is_apfl)
     computed_metrics = meter.compute()
-    assert "test_meter_FedIXI_dice" in computed_metrics
-    dice = computed_metrics["test_meter_FedIXI_dice"]
+    assert "test_meter - prediction - FedIXI_dice" in computed_metrics
+    dice = computed_metrics["test_meter - prediction - FedIXI_dice"]
```
sanaAyrml, line 136: And here.

research/flamby/utils.py
```diff
@@ -98,8 +98,8 @@ def evaluate_model_on_dataset(
             if is_apfl:
                 preds = model(input)["personal"]
             else:
-                preds = model(input)
-            preds = preds if isinstance(preds, dict) else {"predictions": preds}
+                preds = model(input)[0]
+            preds = preds if isinstance(preds, dict) else {"prediction": preds}
```
sanaAyrml, line 102: Due to refactor the output of model version is changed so I needed to update this file.

research/flamby/find_best_hp.py
```diff
@@ -21,16 +21,7 @@ def get_weighted_loss_from_server_log(run_folder_path: str, experiment_name: str
...
-        if experiment_name == "fed_heart_disease":
-            line_to_convert = files_lines[-1].strip()
-        elif experiment_name == "fed_ixi":
-            line_to_convert = files_lines[-1].strip()
-        else:
-            if is_fenda:
-                # FENDA doesn't log the same way as other Fed-ISIC methods
-                line_to_convert = files_lines[-1].strip()
-            else:
-                line_to_convert = files_lines[-3].strip()
```
sanaAyrml, line 33: I don't know why fenda was having this condition previously, but it raised error for me. So I needed to update it.

research/flamby/fed_heart_disease/moon/server.py
emersodb, line 1: I know these Moon Server files existed before we started doing the FLamby experiments, so I think that's why the structure is a bit different. I may be mis-remembering, but because MOON is a "global" model method, like FedAvg where each client receives the same model at each federated round, we want to also do server-side checkpointing like with, for example, FedAvg. That means rather than using a `PersonalServer` we want to use the `FullExchangeServer` class where it also saves checkpoints on the server-side based on the loss function. This would mean the clients save models based on their observed loss and the server saves a model based on its observed loss. The implementation is pretty straightforward (see `research/flamby/fed_heart_disease/fedavg/server.py`). Let me know if I'm wrong here though

research/flamby/fed_isic2019/fenda/run_fold_experiment.slrm
emersodb: Sounds good. Bash is sometimes finicky and sometimes not with this stuff. Thanks for fixing it for visual purposes anyway.

research/flamby/utils.py
emersodb: Yeah, we made some changes to the way this stuff is named in the metrics classes. I'm glad you were able to fix it in a simple way.

research/flamby/find_best_hp.py
emersodb: If I remember correctly, I was getting a few extra lines appearing at the bottom of the logs that looked like this for models that we're FENDA

```
INFO flwr 2023-07-05 13:43:44,075 | server.py:146 | Best Aggregated (Weighted) Loss seen by the Server: 
4.862437355255427
Loaded pretrained weights for efficientnet-b0
Number of features output by EfficientNet 1280
```
It's possible that the logging changed a bit and these are being logged in their proper place now so that this message isn't popping up at the bottom anymore.

emersodb: It's also possible that it only appears at the bottom of the logs when you load the full EfficientNet model, which we do for all methods other than FENDA, FedPer, MOON, and PerFCL. So I think what we actually want to do to fix this is have these methods be considered as the "exceptions." Rather than the argument being `is_fenda` maybe we can change it to `is_fenda_fed_per_moon_perfcl`? It's long but covers all of the methods that don't load a full EfficientNet?

research/flamby/fed_heart_disease/moon/server.py
sanaAyrml: True, if that is the way we do for global models, we should still follow the same way and change. Is the difference only checkpointing for server or there is also difference in checkpointing for client's models? I mean I have this feeling that it would not change anything in results but I am not sure.

emersodb: The difference would only be checkpointing for the server. It doesn't change what the clients decide to checkpoint. So your results are still valid. We just don't necessarily have results from the model that the server would have chosen from its perspective on the aggregated loss.

emersodb: It is unlikely that there is a large improvement in the results. For most methods, we see that local checkpointing is better, but it's good to have another datapoint in that regard.

research/flamby/find_best_hp.py
sanaAyrml: Oh I see what you are talking about, I didn't see it as none of my methods used full EfficientNet models, so we should actually keep it and update this if to some other names. Thanks for pointing it out.

research/flamby/fed_heart_disease/moon/config.yaml
```diff
@@ -6,3 +6,4 @@ n_clients: 4 # The number of clients in the FL experiment
 local_epochs: 1 # The number of epochs to complete for client (NOT USED FOR FLAMBY)
 batch_size: 64 # The batch size for client training (NOT USED FOR FLAMBY)
 local_steps: 100 # The number of local training steps to perform.
+federated_checkpointing: True # Indicates whether intermediate or latest checkpointing is used on the server side
```
emersodb, line 9: I think this is worth adding to the `fed_ixi/moon/config.yaml`. Because of the default for the config.get function, it doesn't change the functionality for `federated_checkpointing` but it makes it obvious how to turn it off.

research/flamby/fed_isic2019/moon/server.py
```diff
@@ -47,23 +53,35 @@ def main(config: Dict[str, Any], server_address: str) -> None:
...
-    server = PersonalServer(client_manager, strategy)
+    server = FullExchangeServer(client_manager, model, strategy, checkpointer=checkpointer)
 
     fl.server.start_server(
         server=server,
         server_address=server_address,
         config=fl.server.ServerConfig(num_rounds=config["n_server_rounds"]),
     )
 
-    log(INFO, "Training Complete")
```
emersodb, line 57: No a big deal at all, but we lost the `log(INFO, "Training Complete")` here in the refactor.

emersodb (approved):
I think all of my comments have been addressed. I had two very minor additional comments that should be very quick. Otherwise, I think we're good to go.

PR #88, opened 2023-12-18
Use poetry as the only source of dependency control

lotif:
# PR Type
Fix

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686rqncr

We use a mix of `requirements.txt` and `pyproject.toml` (from Poetry) for dependency management. This is not a great practice as it makes it difficult to maintain dependencies up to date if we are managing them in multiple places with multiple tools.

Poetry is a far superior tool especially when it comes to dependency conflict management and dependency version locking.

In this PR I am:
- Removing the `requirements.txt` from the project
- Moving all the missing dependencies to `pyproject.toml`
- Updating `README.md` to use Poetry
- Updating the workflows to use Poetry

# Tests Added

Not applicable.


jewelltaylor:
I am able to configure environment and run examples and tests with no issues! I have added a follow up ticket to [clickup](https://github.com/VectorInstitute/FL4Health/tree/main/research/flamby) to explore updating the environment instruction in the FLamby experiments as we discussed. I will wait to take a final pass through this afternoon once you have a chance to update the documentation but all in all the changes look good to me.

jewelltaylor (approved):
LGTM!

PR #89, opened 2023-12-19
Federated PCA

yc7z:
# PR Type
[Feature]

Implementation of the federated principal component analysis functionality. This is a new design different from the previous draft PR. Things that are new in this PR include:

1. An example of performing federated PCA as well as an example that leverages it for dimensionality reduction.
2. A small module `preprocessing/pca_preprocessor.py` that enables the usage of pre-computed principal components for downstream tasks. 
3. A separate PyTorch module `pca.py` that is responsible for locally performing PCA.
4. On the server side, two distinct merging algorithms are implemented in `fedpca.py`.

Clickup Ticket(s): [Link](https://app.clickup.com/t/8686dx41x)

The following two papers contain details of the merging algorithms implemented in `fedpca.py`:

SVD-based merging: [Link](https://arxiv.org/pdf/1601.07010.pdf)

QR-based merging: [Link](https://link.springer.com/chapter/10.1007/978-3-642-20161-5_29)

# Tests Added
`test_fedpca.py` which tests the two merging algorithms in this PR.


fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+        val_data_tensor = self.get_data_tensor(self.val_loader)
+        val_data_tensor_prepared = self.model.centre_data(self.model.maybe_reshape(val_data_tensor))
+        reconstruction_loss = self.model.compute_reconstruction_error(val_data_tensor_prepared, num_components_eval)
+        projection_variance = self.model.compute_projection_variance(val_data_tensor_prepared, num_components_eval)
+        metrics: Dict[str, Scalar] = {}
+        metrics["projection_variance"] = projection_variance
+        return (reconstruction_loss, self.num_val_samples, metrics)
+
+    def save_model(self) -> None:
```
sanaAyrml, line 140: Why do we need that function? Can't we rely on saving with checkpointing?

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,143 @@
...
+from torch import Tensor
+from torch.utils.data import DataLoader
+
+from fl4health.model_bases.pca import PCAModule
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+
+T = TypeVar("T")
+
+
+class FedPCAClient(NumPyClient):
```
sanaAyrml, line 20: Why did you decide to use NumPyClient instead of basic_client?

examples/fedpca_examples/dim_reduction/README.md
```diff
@@ -0,0 +1,42 @@
...
+This example leverages federally computed principal components of the MNIST dataset to perform dimensionality reduction on the images, before proceeding with normal training.
+
+This example assumes that the principal components of MNIST have already been computed and saved (run the example in `exampes/fedpca_examples/perform_pca` to do this), and the user supplies a path to the saved principal components to perform dimensionality reduction.
+
+Each client performs Dirichlet subsampling on the whole dataset to produce heterogeneous local datasets.
+
+## Running the Example
+In order to run the example, first ensure you have the virtual env of your choice activated and run
+```
+pip install --upgrade pip
```
jewelltaylor, line 11: Small point: Marcelo updated the installation instructions of the examples to reflect we are only using poetry for package management in the most recently merged PR. You can take a look what he did and just paste it in here

fl4health/clients/fed_pca_client.py
yc7z: Because `BasicClient` assumes we are performing mini-batch training with a PyTorch model, which I thought was too distinct from how federated PCA works. I could use `BasicClient` instead, but I would still need to overwrite most of its methods

examples/fedpca_examples/dim_reduction/README.md
yc7z: Good point!

fl4health/clients/fed_pca_client.py
yc7z: I didn't use a checkpointer because we only need to save the principal components once after merging, so we're not technically doing checkpointing. I could change this if you think using checkpointing is better.

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -0,0 +1,69 @@
...
+        [
+            transforms.ToTensor(),
+            transforms.Normalize((0.5), (0.5)),
+        ]
+    )
+    return MNISTDataset(data_dir, train=train, transform=transform)
+
+
+class MnistFedPCAClient(BasicClient):
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
```
fatemetkl, line 33: Should we make sure the dim_reduction client has the same data as perform_pca since it uses the components calculated in the previous example? maybe by setting the random seed in both examples.

yc7z: Yes that's a good point! I think fixing the seed makes more sense as well

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+        training data so it can be used to center validation/test data later,
+        if needed.
+        """
+        self.data_mean = torch.mean(X, dim=0)
+
+    def centre_data(self, X: Tensor) -> Tensor:
+        assert self.data_mean is not None
+        return X - self.data_mean
+
+    def prepare_data_forward(self, X: Tensor, center_data: bool) -> Tensor:
```
yc7z, line 83: As discussed in yesterday's meeting, I'm not sure whether we should include preprocessing steps such as centering and reshaping data as a part of the module or leave it to the user. If anyone has any thoughts on this please let me know!

examples/fedpca_examples/dim_reduction/README.md
```diff
@@ -0,0 +1,37 @@
...
+* `n_clients`: number of clients the server waits for in order to run the FL training.
+* `local_epochs`: number of local epochs each client will train.
+* `batch_size`: size of the batches each client will train on.
+* `new_dimension`: new dimension after reduction.
+* `pca_path`: path to pre-computed principal components.
+* `checkpoint_path`: path where the model weights are saved.
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the two
+clients. This is done by simply running (remembering to activate your environment)
```
emersodb, line 27: Very minor, but based on the config, it looks like we're expecting the user to start three clients?

examples/fedpca_examples/dim_reduction/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+python -m examples.fedpca_examples.dim_reduction.client --dataset_path /path/to/data
+```
+**NOTE**:
+
+* The argument `dataset_path` has two functions, depending on whether the dataset exists locally or not. If
+the dataset already exists at the path specified, it will be loaded from there. Otherwise, the dataset will be
+automatically downloaded to the path specified and used in the run.
+
+After both clients have been started federated learning should commence.
```
emersodb, line 37: If we are starting three clients, we can just edit this to be, `After the clients have been started...`

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def get_mnist_dataset(data_dir: Path, train: bool) -> MNISTDataset:
+    transform = transforms.Compose(
+        [
+            transforms.ToTensor(),
+            transforms.Normalize((0.5), (0.5)),
+        ]
+    )
+    return MNISTDataset(data_dir, train=train, transform=transform)
```
emersodb, line 29: As a convention, we've settled on acronyms in classes being upper-case for the first letter and lowercase thereafter (this is just a choice we've made, the conventions allow for either). So this would mean that"
- `PCAPreprocessor` -> `PcaPreprocessor`
- `MNISTDataset` -> `MnistDataset`
- `MnistFedPCAClient` -> `MnistFedPcaClient`
If there are other places in the code, this change should be made there too 🙂

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -0,0 +1,78 @@
...
+        pca_path = Path(self.narrow_config_type(config, "pca_path", str))
+        new_dimension = self.narrow_config_type(config, "new_dimension", int)
+        pca_preprocessor = PCAPreprocessor(pca_path)
+        sampler = DirichletLabelBasedSampler(list(range(10)), sample_percentage=0.6, beta=0.75)
+
+        train_dataset = get_mnist_dataset(data_dir=self.data_path, train=True)
+
+        validation_dataset = get_mnist_dataset(data_dir=self.data_path, train=False)
+
+        train_loader = pca_preprocessor.reduce_dimension(
```
emersodb, line 44: I would argue that we don't want to support sampler is the reduce_dimension functionality. Unless I'm mistaken, the sampling doesn't have anything to do with dimensionality reduction. So in the interest of "single responsibility," I think we should just have it take a dataset and produce a dataset (with the dimensionality reduced. We can support sampling in `get_mnist_dataset` and create the dataloader from the dimensionality reduced dataset in this function, which wouldn't be a huge refactor I don't think?

emersodb: Similarly, we could support shuffling and batch_size setting when creating the dataloader, rather than having the `reduce_dimension` function being responsible for that.

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -0,0 +1,78 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="FL Client Main")
+    parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
+    args = parser.parse_args()
+
+    DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    data_path = Path(args.dataset_path)
+
+    torch.manual_seed(47)
```
emersodb, line 75: I think it's probably best to use the functionality that Sana created in `from fl4health.utils.random import set_all_random_seeds` since it handles setting all of the relevant seeds for deterministic training? Perhaps you didn't do that for a reason. If so, a comment would be useful

examples/fedpca_examples/dim_reduction/config.yaml
```diff
@@ -0,0 +1,12 @@
+# Parameters that describe server
+n_server_rounds: 3 # The number of rounds to run FL
+
+# Parameters that describe clients
+n_clients: 3 # The number of clients in the FL experiment
+local_epochs: 2 # The number of epochs to complete for client
+batch_size: 64 # The batch size for client training
+new_dimension: 12
```
emersodb, line 8: Perhaps a comment here on what this is.

examples/fedpca_examples/dim_reduction/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+# checkpointing
```
emersodb, line 10: Maybe elaborate that these are the paths to the pre-computed principal components (`pca_path`) and the path where FL trained models will be saved during training (`checkpoint_path`)

examples/fedpca_examples/dim_reduction/mnist_model.py
```diff
@@ -0,0 +1,17 @@
+import torch
```
emersodb, line 1: Not super important, but we've been sticking models that the examples use in `examples/models`

examples/fedpca_examples/dim_reduction/server.py
```diff
@@ -0,0 +1,89 @@
...
+
+def main(config: Dict[str, Any]) -> None:
+    # This function will be used to produce a config that is sent to each client to initialize their own environment
+    fit_config_fn = partial(
+        fit_config, config["local_epochs"], config["batch_size"], config["new_dimension"], config["pca_path"]
+    )
+
+    # Initializing the model on the server side
+    model = MnistNet(input_dim=config["new_dimension"])
+    # To facilitate checkpointing
```
emersodb, line 49: I'd move this comment down one line, as the parameter exchanger isn't strictly to facilitate checkpointing 🙂

examples/fedpca_examples/dim_reduction/client.py
yc7z: Yes that's a good point!

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
+# Federated Principal Component Analysis Example
+This example performs federated principal component analysis. The goal is to compute the principal components of a subset of the MNIST dataset, under the condition that the data is distributed across two distinct clients.
```
emersodb, line 2: It looks like in the config.yaml we actually have four clients taking part in the FedPCA calculations? Perhaps I am misinterpreting the comment though.

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This is achieved by each client performing PCA locally at first, then the principal components are sent to a central server to be merged.
+
+Each client performs Dirichlet subsampling on the whole dataset to produce heterogeneous local datasets. This is done to ensure that local principal compoents are distinct enough across different clients.
```
emersodb, line 6: I would just say "distinct" instead of "distinct enough." We don't really define what we mean by "enough" in this instance

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
...
+python -m examples.fedpca_examples.perform_pca.server  --config_path /path/to/config.yaml
+```
+from the FL4Health directory. The following arguments must be present in the specified config file:
+* `n_clients`: number of clients the server waits for in order to run the FL training
+* `batch_size`: size of the batches each client will train on
+* `low_rank`: whether the data matrix is low rank or not. If the user has prior knowledge that the data matrix is low rank, then this parameter can be set to True so that each client can leverage this knowledge to perform local PCA more efficiently.
+* `full_svd`: determines whether full SVD or reduced SVD is performed by each client.
+* `rank_estimation`: an estimation of the rank of the data matrix. This is only used if `low_rank` is set to True.
+* `center_data`: if set to True, the mean of data will be subtracted from all data points before local PCA is performed.
+* `num_components`: by default, after merging is completed, each client will use the merged principal components to compute the reconstruction error on its validation set. This parameter specify the number of principal components that are used in this evaluation process.
```
emersodb, line 24: specify -> specifies

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the two
```
emersodb, line 28: From the config it seems like we might be starting four instead of two clients?

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
...
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.fedpca_examples.perform_pca.client --dataset_path /path/to/data --model_save_path /path/to/save/model
```
emersodb, line 31: It might be better to name with something like `principal_components_save_path` or just `components_save_path` to avoid confusion with a full ML model? There might be a reason that I'm not foreseeing to not use these names though.

examples/fedpca_examples/perform_pca/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+**NOTE**:
+
+* The argument `dataset_path` has two functions, depending on whether the dataset exists locally or not. If
+the dataset already exists at the path specified, it will be loaded from there. Otherwise, the dataset will be
+automatically downloaded to the path specified and used in the run.
+
+* The argument `model_save_path` specifies the directory in which the merged principal components will be saved, so they can be leveraged for other downstream tasks. An example of dimensionality reduction can be found at `exampes/fedpca_examples/dim_reduction`.
+
+After both clients have been started federated pca should commence.
```
emersodb, line 41: If we're starting more than two we can just rewrite this is `After the clients have been started`

examples/fedpca_examples/perform_pca/README.md
yc7z: Ah yes. Originally both of the FedPCA examples involve only two clients. After I locally tested for more clients, I committed those changes but forgot to update the README. I'll update them so they are consistent with the config file. Sorry for the confusion 😂

examples/fedpca_examples/perform_pca/client.py
```diff
@@ -0,0 +1,40 @@
...
+    parser = argparse.ArgumentParser(description="FL Client Main")
+    parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
+    parser.add_argument("--model_save_path", action="store", type=str, help="Path to saving model parameters")
+    args = parser.parse_args()
+
+    DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    data_path = Path(args.dataset_path)
+    model_save_path = Path(args.model_save_path)
+
+    torch.manual_seed(47)
```
emersodb, line 38: Perhaps fix the seed with `set_all_random_seeds` here as well.

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def get_data_tensor(self, data_loader: DataLoader) -> Tensor:
+        raise NotImplementedError
+
+    def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict[str, Scalar]]:
+        """Perform PCA using the locally held dataset."""
+        if not self.initialized:
+            self.setup_client(config)
+        center_data = self.narrow_config_type(config, "center_data", bool)
+        train_data_tensor = self.get_data_tensor(self.train_loader)
```
emersodb, line 102: these two lines can be merged into a single one without a large loss in readability I think.
```
train_data_tensor = self.get_data_tensor(self.train_loader).to(self.device)
```

examples/fedpca_examples/perform_pca/config.yaml
```diff
@@ -0,0 +1,12 @@
+# Parameters that describe server
+n_server_rounds: 1 # The number of rounds to run FL
+
+# Parameters that describe clients
+n_clients: 4 # The number of clients in the FL experiment
+batch_size: 32 # The batch size for client training
```
emersodb, line 6: I'd change the comment on this variable here, since we're not really using the batch_size for training in the PCA example.

examples/fedpca_examples/perform_pca/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+low_rank: False
```
emersodb, line 8: Some brief inline comments (along the same lines as the readme, but can be less informative) would be useful.

examples/fedpca_examples/perform_pca/server.py
```diff
@@ -0,0 +1,89 @@
...
+    strategy = FedPCA(
+        # Server waits for min_available_clients before starting FL rounds
+        min_available_clients=config["n_clients"],
+        min_fit_clients=config["n_clients"],
+        on_fit_config_fn=fit_config_fn,
+        # We use the same fit config function, as nothing changes for eval
+        on_evaluate_config_fn=fit_config_fn,
+        fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+        evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        initial_parameters=get_initial_model_parameters(),
```
emersodb, line 62: Since we have control over this strategy, we can just set initial_parameters to `Parameters(tensors=[], tensor_type="numpy.ndarray")` in the strategy to enforce this strictly and not have it as an argument.

examples/fedpca_examples/perform_pca/server.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument(
+        "--config_path",
+        action="store",
+        type=str,
+        help="Path to configuration file.",
+        default="examples/fedpca_examples/perform_pca/config.yaml",
+    )
+    args = parser.parse_args()
+
+    config = load_config(args.config_path)
```
emersodb, line 87: Since we only intend to perform a single server round for FedPCA, perhaps we can assert that the `config["n_server_rounds"]` is one here?

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+            raise ValueError(f"{config_key} is not present in the Config.")
+
+        config_value = config[config_key]
+        if isinstance(config_value, narrow_type_to):
+            return config_value
+        else:
+            raise ValueError(f"Provided configuration key ({config_key}) value does not have correct type")
+
+    def get_parameter_exchanger(self, config: Config) -> FullParameterExchanger:
+        return FullParameterExchanger()
```
emersodb, line 59: Is there a situation where we might consider only exchanging some of the principal components? If not, then I think we can just set `self.parameter_exchanger = FullParameterExchanger()` in the __init__ of this class? If there is a setting with partial exchange, I'm good with this structure.

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def generate_hash(self, length: int = 8) -> str:
+        return "".join(random.choice(string.ascii_lowercase) for _ in range(length))
+
+    def get_parameters(self, config: Config) -> NDArrays:
+        assert self.model is not None and self.parameter_exchanger is not None
+        return self.parameter_exchanger.push_parameters(self.model, config=config)
+
+    def set_parameters(self, parameters: NDArrays, config: Config) -> None:
+        self.parameter_exchanger.pull_parameters(parameters, self.model, config)
```
emersodb, line 45: This is a bit different than the standard flow, as we're saving the model on setting the parameters. A comment would be useful to explain why.

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+        train_data_tensor = self.get_data_tensor(self.train_loader)
+        train_data_tensor = train_data_tensor.to(self.device)
+
+        principal_components, singular_values = self.model(train_data_tensor, center_data)
+        self.model.set_principal_components(principal_components, singular_values)
+
+        metrics: Dict[str, Scalar] = {}
+        cumulative_explained_variance = self.model.compute_cumulative_explained_variance()
+        explained_variance_ratios = self.model.compute_explained_variance_ratios()
+        metrics["cumulative_explained_variance"] = cumulative_explained_variance
```
emersodb, line 111: Not super important, but since you're not dynamically creating the metrics dictionary you could just do something like
```
metrics = {
    "cumulative_explained_variance": cumulative_explained_variance,
    "top_explained_variance_ratio": explained_variance_ratios[0].item()
}
```
and forgo declaring the empty dictionary. The same could be done in `evaluate` below.

examples/fedpca_examples/perform_pca/README.md
yc7z: Yes I agree that would be clearer. I originally named it `model_save_path` because the PCA module is implemented as a PyTorch "model"

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+            train_data_tensor = self.get_data_tensor(self.train_loader)
+            self.model.set_data_mean(self.model.maybe_reshape(train_data_tensor))
+        self.set_parameters(parameters, config)
+        num_components_eval = (
+            self.narrow_config_type(config, "num_components_eval", int)
+            if "num_components_eval" in config.keys()
+            else None
+        )
+        val_data_tensor = self.get_data_tensor(self.val_loader)
+        val_data_tensor_prepared = self.model.centre_data(self.model.maybe_reshape(val_data_tensor)).to(self.device)
```
emersodb, line 134: Correct me if I'm wrong, but this will explicitly center the data whether or not the data was centered during training? If the data wasn't centered during training, that might lead to a mismatch? I may be missing something though.

fl4health/clients/fed_pca_client.py
```diff
@@ -0,0 +1,144 @@
...
+        reconstruction_loss = self.model.compute_reconstruction_error(val_data_tensor_prepared, num_components_eval)
+        projection_variance = self.model.compute_projection_variance(val_data_tensor_prepared, num_components_eval)
+        metrics: Dict[str, Scalar] = {}
+        metrics["projection_variance"] = projection_variance
+        return (reconstruction_loss, self.num_val_samples, metrics)
+
+    def save_model(self) -> None:
+        final_model_save_path = f"{self.model_save_path}/client_{self.generate_hash()}_pca"
```
emersodb, line 142: We might want a file type at the end of this file name like `.pkl` or `cpt` or something?

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+    def __init__(self, low_rank: bool = True, full_svd: bool = False, rank_estimation: int = 6) -> None:
+        """
+        PyTorch module for performing Principal Component Analysis.
+
+        Args:
+            low_rank (bool, optional): Indicates whether the data matrix is low-rank. If the user has
+             prior knowledge that it is, then this parameter can be set to True to allow for more efficient
+              computation of SVD. Defaults to True.
+            full_svd (bool, optional): Indicates whether full SVD or reduced SVD is performed. Defaults to False.
+            rank_estimation (int, optional): An estimation of the rank of the data matrix.
```
emersodb, line 21: This will probably be answered below. However, it would be good to have a comment covering it here. Is a rank_estimation value required if `low_rank` is true? I could see a situation where one suspects that the matrix has a low-rank representation but the true rank is unknown. Moreover, what is the "penalty" or "downside" to being off in this estimation?

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+from torch.nn.parameter import Parameter
+
+
+class PCAModule(nn.Module):
+    def __init__(self, low_rank: bool = True, full_svd: bool = False, rank_estimation: int = 6) -> None:
+        """
+        PyTorch module for performing Principal Component Analysis.
+
+        Args:
+            low_rank (bool, optional): Indicates whether the data matrix is low-rank. If the user has
```
emersodb, line 17: Is this approximately low rank or exactly low rank. That is, does the matrix have to have an analytical low rank representation? I imagine it is "approximately low rank" where this means that the matrix can be well-approximated with a significantly smaller set of principal components than the full rank, but it would be good to be explicit for the user.

Additionally, what is the cost to getting this wrong. That is, I think the matrix is of low approximate rank, but it turns out it isn't. That's worth mentioned I think?

emersodb: **Note**: You can also just link to the SVD implementation documentation if you think they do a good job explaining such things 🙂

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+            performing PCA.
+
+        Returns:
+            Tuple[Tensor, Tensor]: The principal components (i.e., left singular vectors) and their corresponding
+            singular values.
+        """
+        X_prime = self.prepare_data_forward(X, center_data=center_data)
+        if self.low_rank:
+            log(INFO, "Assuming data matrix is low rank, using low-rank PCA implementation.")
+            q = min(self.rank_estimation, X_prime.size(0), X_prime.size(1))
```
emersodb, line 48: If the rank estimation is larger than the appropriate matrix dimension, we should probably log a warning, as the user sort of did something weird?

fl4health/model_bases/pca.py
emersodb: I'm alright with supporting the centering functionality. However, I think we need to make sure that we return the data in the way it was provided. That is, if centering is true, when we do the data reconstruction we add back the mean (for example, to compute reconstruction error). 

Another option is to assume that the user has centered the data before and throw an error is there is a non-zero mean (up to machine precision)

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+        self.data_mean: Tensor
+
+    def forward(self, X: Tensor, center_data: bool) -> Tuple[Tensor, Tensor]:
+        """
+        Perform PCA on the data matrix X by computing its SVD.
+
+        Args:
+            X (Tensor): Data matrix.
+            center_data (bool): If true, then the data mean will be subtracted from all data points prior to
+            performing PCA.
```
emersodb, line 39: I would also state that if `center_data` is false, it is expected that the data is already centered and an exception will be thrown if it is not.

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+            return self.centre_data(X)
+        else:
+            return X
+
+    def project_lower_dim(self, X: Tensor, k: Optional[int] = None) -> Tensor:
+        """
+        Project input data X onto the top k principal components.
+
+        Args:
+            X (Tensor): Input Data.
```
emersodb, line 99: This assumes that the data has been appropriately centered correct? We should probably support centering here as well (that is, applying the learned data mean to the input)?

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+            if self.full_svd:
+                log(INFO, "Performing full SVD on data matrix.")
+            else:
+                log(INFO, "Performing reduced SVD on data matrix.")
+            _, S, Vh = torch.linalg.svd(X_prime, full_matrices=self.full_svd)
+            principal_components = Vh.T
+        singular_values = S
+        return principal_components, singular_values
+
+    def maybe_reshape(self, X: Tensor) -> Tensor:
```
emersodb, line 61: Perhaps a bit more comment on why this reshape is potentially required?

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+from typing import Optional, Tuple
+
+import torch
+import torch.nn as nn
+from flwr.common.logger import log
+from torch import Tensor
+from torch.nn.parameter import Parameter
+
+
+class PCAModule(nn.Module):
```
emersodb, line 11: PcaModule by our convention 🙂

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Returns:
+            Tuple[Tensor, Tensor]: The principal components (i.e., left singular vectors) and their corresponding
+            singular values.
+        """
+        X_prime = self.prepare_data_forward(X, center_data=center_data)
+        if self.low_rank:
+            log(INFO, "Assuming data matrix is low rank, using low-rank PCA implementation.")
+            q = min(self.rank_estimation, X_prime.size(0), X_prime.size(1))
+            _, S, V = torch.pca_lowrank(X_prime, q=q, center=False)
```
emersodb, line 49: I think you can just do
```
_, singular_values, principal_components = torch.pca_lowrank(X_prime, q=q, center=False)
```
Similarly below I think you can do
```
_, singular_values, Vh = torch.linalg.svd(X_prime, full_matrices=self.full_svd)
principal_components = Vh.T
```
and get ride of the `singular_values = S` on line 58

examples/fedpca_examples/dim_reduction/client.py
yc7z: Yes that makes sense, and I also agree that decoupling the dataloader creation from this method is better from a "single responsibility" perspective. I just want to point out that due to the way PyTorch DataLoader class is designed, the dimension reduction transform won't actually be applied to the data until the user iterate through the dataloader in case you had something different in mind 🙂

emersodb: Gotcha. So it's a lazy application. I think we can still modify the transform of the dataset using this functionality in the same way, just with a separation of the functionality.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+        self, client_singular_vectors: NDArrays, client_singular_values: NDArrays
+    ) -> Tuple[NDArray, NDArray]:
+        """
+        Produce the principal components (PCs) for all the data distributed across clients by merging the PCs
+        belonging to each local dataset.
+
+        Each clients sends a matrix whose columns are the local principal components to the server. The corresponding
+        singular values are also shared.
+
+        The server then arranges the local PCs into a block matrix, then perform SVD.
```
emersodb, line 164: perform -> performs

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+        return torch.matmul(X_lower_dim_prime, self.principal_components[:, :k].T)
+
+    def compute_reconstruction_error(self, X: Tensor, k: Optional[int]) -> float:
+        """
+        Compute the reconstruction error of X under PCA reconstruction.
+
+        More precisely, if X is n by d, and U is the matrix whose columns are the
+        k principal components of X (thus U is d by k), then the reconstruction loss
+        is defined as
+            1 / n * | X @ U @ U.T - X| ** 2.
```
emersodb, line 129: Is the `@` symbol here denoting matrix multiplication? If this is standard, we can leave it, I'm just not used to the notation.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+
+        For example, if U_i denotes the matrix of PCs and S_i denotes the diagonal matrix of
+        singular values for client i, and there are n clients, then merging is done by
+        performing SVD on the matrix
+
+        B = [U_1 @ S_1 | U_2 @ S_2 | ... | U_n @ S_n],
```
emersodb, line 170: I think the more standard notation here would be `diag([U_1 @ S_1, U_2 @ S_2, ..., U_n @ S_n])`, but perhaps I'm wrong.

fl4health/model_bases/pca.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        Perform PCA on the data matrix X by computing its SVD.
+
+        Args:
+            X (Tensor): Data matrix.
+            center_data (bool): If true, then the data mean will be subtracted from all data points prior to
+            performing PCA.
+
+        Returns:
+            Tuple[Tensor, Tensor]: The principal components (i.e., left singular vectors) and their corresponding
```
emersodb, line 42: It looks like we're returning V (or V^T if computed with SVD), which is contains the right singular vector rather than the left if I'm not mistaken.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+
+        where the new (left) singular vectors are returned as the merging result.
+
+        This implementation assumes that the *columns* of U_i are the PCs.
+
+        If
+
+        U @ S @ V.T = B
+
+        is the SVD of B, then it turns out that U = A @ U',
```
emersodb, line 180: In the implementation of pca.py, we see to be returning the right singular vectors rather than the left singular vectors. For standard PCA, we can perform projection with either through different computations. However, does this procedure still work?

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+        Args:
+            client_singular_vectors (NDArrays): Local PCs.
+            client_singular_values (NDArrays): Singular values corresponding to local PCs.
+
+        Returns:
+            Tuple[NDArray, NDArray]: merged PCs and corresponding singular values.
+        """
+        X = [U @ np.diag(S) for U, S in zip(client_singular_vectors, client_singular_values)]
+        svd_input = np.concatenate(X, axis=1)
+        new_singular_vectors, new_singular_values, _ = np.linalg.svd(svd_input, full_matrices=True)
```
emersodb, line 196: If I recall correctly, the projection procedure is slightly different for right vs. left singular vectors. So we definitely want to make sure the math here is fully aligned.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+        Args:
+            client_singular_vectors (NDArrays): Local PCs.
+            client_singular_values (NDArrays): Singular values corresponding to local PCs.
+
+        Returns:
+            Tuple[NDArray, NDArray]: merged PCs and corresponding singular values.
+        """
+        assert len(client_singular_values) >= 2
+        if len(client_singular_values) == 2:
+            singular_values_diagonal_matrices = [
```
emersodb, line 236: A more straightforward construction might be clearer given the length assertion. That is, 
```
U1, S1 = client_singular_vectors[0], np.diag(client_singular_values[0])
U2, S2 = client_singular_vectors[1], np.diag(client_singular_values[1])
```
and drop the list comprehension above.

fl4health/strategies/fedpca.py
```diff
@@ -0,0 +1,269 @@
...
+                np.diag(singular_values_vector) for singular_values_vector in client_singular_values
+            ]
+            U1, S1 = client_singular_vectors[0], singular_values_diagonal_matrices[0]
+            U2, S2 = client_singular_vectors[1], singular_values_diagonal_matrices[1]
+            return self.merge_subspaces_qr_helper((U1, S1), (U2, S2))
```
emersodb, line 241: Rather than helper, maybe we rename this two, `merge_two_subspaces_qr` to make it more explicit?

tests/strategies/test_fedpca.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    # svd_merged_vectors is only guaranteed to be the same as V up to the application
+    # of a blockwise unitary transformation, thus we do not check whether it is close to V.
+    assert np.allclose(svd_merged_singular_values, S)
+    assert np.allclose(qr_merged_singular_values, S)
+
+    # Instead, we verify that the merging results are unitary matrices.
+    assert np.allclose(svd_merged_vectors.T @ svd_merged_vectors, np.identity(svd_merged_vectors.shape[1]))
+    assert np.allclose(qr_merged_vectors.T @ qr_merged_vectors, np.identity(qr_merged_vectors.shape[1]))
```
emersodb, line 51: Some tests of the PCA module would be good too. That is, does PCA without truncating the singular value results in zero reconstruction error. If the data is of low rank, say (x, y, 1) where x and y vary randomly but they all have a z value of 1, are the singular values 0 for the third dimension. Do we get exact reconstruction error when truncating to 2 dimensions?

emersodb: Some tests around centering and not centering the data would be useful as well.

emersodb (commented):
Overall I think the design is pretty good! However, there are a few minor comments and design choices I think that we can discuss. One of the main concerns I have is around the alignment of using left and right singular value matrices. It seems like we have some misalignment as to which values are required for merging (or at least the comments are misaligned).

fl4health/clients/fed_pca_client.py
yc7z: I believe we always want to exchange all of the principal components, so I think your suggestion makes more sense.

fl4health/clients/fed_pca_client.py
yc7z: Yes, good point

fl4health/clients/fed_pca_client.py
yc7z: Yes for sure. Since federated PCA only runs for one round, the parameters obtained here are the merged principal components, and because currently for downstream tasks we need to save these parameters, I thought this is the most appropriate place to save them. I've added a comment to this method as well.

fl4health/model_bases/pca.py
yc7z: Good point!

fl4health/model_bases/pca.py
yc7z: I think these are important questions. I've added more comments addressing them, and we can discuss whether the updates have cleared things up 🙂

fl4health/model_bases/pca.py
yc7z: Yes, `@` is matrix multiplication. I thought it's standard since you can use it to do matrix multiplication in both NumPy and PyTorch but maybe it's not commonly used...?

fl4health/strategies/fedpca.py
yc7z: I meant for this to mean that `B` is a block matrix consisting of blocks `U1S1, U2S2, ...,  UnSn` from arranged from left to right. I wasn't sure what a good notation for this would be in the docstring

fl4health/model_bases/pca.py
yc7z: Yes absolutely

fl4health/strategies/fedpca.py
yc7z: Ah I see the confusion. I should've made this more clear. Here we are returning the left singular vectors because the algorithm assumes that the **columns** of the data matrix (sent by clients to the server) are the local principal components. 

In contrast, `pca.py`, assumes that the **rows** of the data matrix are the data points. Basically the input data matrices are "transposes" of each other in the two cases, so their SVDs are also flipped. That's why we are returning the left singular vectors here and right singular vectors in `pca.py`. 

I did this because the first dimension of a dataset is usually the batch dimension, so it makes sense that the data points would be rows, hence the design in `pca.py`. As for `fedpca.py`, I assumed the local principal components come in columns because all the papers I read on this seem to assume so. 

I agree this is pretty confusing. So I think we can either clarify things better in the comments or change the implementations to be consistent, I'm okay with either approach

tests/strategies/test_fedpca.py
yc7z: Yes I think this is important. I'll add more tests for `pca.py`

fl4health/model_bases/pca.py
emersodb: That's good with me. I didn't realize that their docs used this symbol for matrix multiplication. Good to know!

fl4health/strategies/fedpca.py
emersodb: Gotcha. So if this is supposed to be a block diagonal matrix, I think a standard notation for this would be similar to the diag notation I put above. That is, if n = 2, you want to indicate a matrix that looks like this
```
U_1 @ S_1           0
      0            U_2 @ S_2
```
where the zeros are blocks of 0s with the appropriate dimensions.

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -11,38 +11,45 @@
...
-        train_loader = pca_preprocessor.reduce_dimension(
-            new_dimension, batch_size, True, sampler, get_mnist_dataset, self.data_path, True
-        )
-        val_loader = pca_preprocessor.reduce_dimension(
-            new_dimension, batch_size, False, sampler, get_mnist_dataset, self.data_path, True
-        )
+        # Get training and validation datasets.
+        train_dataset: BaseDataset = get_mnist_dataset(data_dir=self.data_path, train=True)
+        validation_dataset: BaseDataset = get_mnist_dataset(data_dir=self.data_path, train=False)
+        train_dataset = sampler.subsample(dataset=train_dataset)
```
emersodb, line 44: Maybe this is intentional, but it looks like the validation dataset is no longer being subsampled as it had been in the previous implementation.

examples/fedpca_examples/dim_reduction/client.py
```diff
@@ -11,38 +11,45 @@
...
+
+        # Apply dimensionality reduction/.
```
emersodb, line 46: Very minor: I think the slash here is a small typo.

examples/fedpca_examples/perform_pca/config.yaml
```diff
@@ -3,10 +3,11 @@ n_server_rounds: 1 # The number of rounds to run FL
...
-batch_size: 32 # The batch size for client training
+batch_size: 32 # The batch size used for dataloader creation. (This parameter is otherwise not used in the execution of federated PCA.)
 
-low_rank: False
-full_svd: False
-rank_estimation: 6
-center_data: True
-num_components_eval: 16
+# The following are relevant technical parameters related to performing PCA.
+low_rank: False # Indicates whether the user wish to assume the input data has low rank.
```
emersodb, line 9: Super minor: wishes rather than wish?

fl4health/model_bases/pca.py
```diff
@@ -8,18 +8,46 @@
...
 
         Args:
-            low_rank (bool, optional): Indicates whether the data matrix is low-rank. If the user has
-             prior knowledge that it is, then this parameter can be set to True to allow for more efficient
-              computation of SVD. Defaults to True.
+            low_rank (bool, optional): Indicates whether the data matrix can be well-approximated
+            by a low-rank singular value decomposition. If the user has
+            good reasons to believe so, then this parameter can be set to True to allow for more efficient
+            computations. Defaults to False.
             full_svd (bool, optional): Indicates whether full SVD or reduced SVD is performed. Defaults to False.
```
emersodb, line 21: Maybe we can also state that `If low_rank is set to true, this argument is ignored.`?

fl4health/strategies/fedpca.py
yc7z: Actually B wouldn't be a block diagonal matrix. It would just be a matrix with blocks `U_1 @ S_1` and `U_2 @ S_2` placed from left to right (i.e., there won't be any zero blocks).

So for example, if `U_1 @ S_1` has size d by n1 and `U_2 @ S_2` has size d by n2, then `B` would have size d by (n1 + n2).

examples/fedpca_examples/perform_pca/config.yaml
```diff
@@ -3,10 +3,11 @@ n_server_rounds: 1 # The number of rounds to run FL
...
+full_svd: False # Indicates whether to perform full SVD or reduced SVD.
+rank_estimation: 6 # Indicates the estimated rank of the data matrix. Only used if low_rank is True.
+center_data: True #  Indicates whether the PCA module should center the data before performing PCA.
+num_components_eval: 16 # Indicates the number of principal components used during evaluation when computing reconstruction error.
```
emersodb, line 13: Perhaps state that if it isn't in the config it defaults to none and all components are used?

fl4health/strategies/fedpca.py
emersodb: Ah okay I see. Your comments are clear on this. I'm not sure why I assume we needed it to be block diagonal. Apologies.

fl4health/model_bases/pca.py
```diff
@@ -35,32 +63,40 @@ def forward(self, X: Tensor, center_data: bool) -> Tuple[Tensor, Tensor]:
 
         Args:
             X (Tensor): Data matrix.
-            center_data (bool): If true, then the data mean will be subtracted from all data points prior to
-            performing PCA.
+            center_data (bool): If true, then the data mean will be subtracted
+            from all data points prior to performing PCA. If center_data is false,
+            it is expected that the data has already been centered and an exception
```
emersodb, line 68: I don't think we're throwing an exception. So I assume the underlying method calls will throw one?

fl4health/model_bases/pca.py
```diff
@@ -35,32 +63,40 @@ def forward(self, X: Tensor, center_data: bool) -> Tuple[Tensor, Tensor]:
...
+            will be thrown if it is not.
 
         Returns:
-            Tuple[Tensor, Tensor]: The principal components (i.e., left singular vectors) and their corresponding
-            singular values.
+            Tuple[Tensor, Tensor]: The principal components (i.e., right singular vectors)
+            and their corresponding singular values.
+
+        Note: the algorithm assumes that the rows of X_prime are the data points.
```
emersodb, line 75: Just a small thing. `X_prime` is only mentioned in the code, but in generated documentation, the user will only be aware of X. So unless we introduce X_prime briefly, it might be confusing.

fl4health/model_bases/pca.py
yc7z: Yes you are right. I've updated `prepare_data_forward()` to check whether the data already has zero mean and throw an exception accordingly.

fl4health/model_bases/pca.py
yc7z: Good point. I think it might be clearer to just use `X` in the documentation then (mentioning that we assume it is reshaped as needed)

fl4health/strategies/fedpca.py
yc7z: No worries at all! The notation was a bit confusing

examples/fedpca_examples/dim_reduction/client.py
yc7z: I'm not sure why I deleted the subsampling for validation data... It's been added back now

emersodb (approved):
All of my remaining comments here are minor, so I'm good to merge once those small changes are complete!

PR #87, opened 2023-12-15
Add pip-audit to the CI checks

lotif:
# PR Type
Other

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686rqngw

[`pip-audit`](https://github.com/pypa/pip-audit) is a tool that scans the installed libraries in the project in search of versions that have reported security vulnerabilities.

In this PR, I am:
- adding `pip-audit` as a step in the static code analysis CI workflow
- Fixing the libraries that it reports as vulnerable, including those that are requirements of other requirements.

Here is a list of vulnerabilities that I am fixing in this PR in the `poetry.lock` file:

| Dependency  | Affected Verison  | Sec Report  |
| ------------- | -----------------  | --------------  |
| aiohttp | 3.8.6  | [GHSA-qvrw-v9rv-5rjx](https://github.com/aio-libs/aiohttp/security/advisories/GHSA-qvrw-v9rv-5rjx) |
| aiohttp | 3.8.6 | [GHSA-q3qx-c6g2-7pw2](https://github.com/aio-libs/aiohttp/security/advisories/GHSA-q3qx-c6g2-7pw2) |
| certifi | 2022.12.7 | [PYSEC-2023-135](https://vulners.com/osv/OSV:PYSEC-2023-135) |
| urllib3 | 1.26.15 | [PYSEC-2023-192](https://deps.dev/advisory/osv/PYSEC-2023-192) |
| urllib3 | 1.26.15 | [PYSEC-2023-212](https://github.com/advisories/GHSA-g4mx-q9vg-27p4) |
| werkzeug | 2.3.2 | [PYSEC-2023-221](https://github.com/pypa/advisory-database/blob/main/vulns/werkzeug/PYSEC-2023-221.yaml) |

And here are 4 vulnerabilities on Pillow that I am ignoring because I could not fix, pending theresolution of the issue below:
https://github.com/SeldonIO/alibi/issues/991

| Dependency  | Affected Verison  | Sec Report  |
| ------------- | -----------------  | --------------  |
| pillow | 9.5.0 | [PYSEC-2023-175](https://scout.docker.com/vulnerabilities/id/PYSEC-2023-175) |
| pillow | 9.5.0 | [PYSEC-2023-227](https://scout.docker.com/vulnerabilities/id/PYSEC-2023-227) |
| pillow | 9.5.0 | [GHSA-j7hp-h8jx-5ppr](https://secalerts.co/vulnerability/GHSA-j7hp-h8jx-5ppr) |
[...]

PR #90, opened 2023-12-20
Fix cyclops module import, adjust dependencies

amrit110:
# PR Type
Other

# Short Description
- With some cleanup of dependency management happening in recent PRs such as #88, #87 I took a look at the direct package imports in the repo (attached)
[requirements.txt](https://github.com/VectorInstitute/FL4Health/files/13722627/requirements.txt).
- I propose the following: Any dependency used in the `fl4health` package is a core dependency. Hence in the `pyproject.toml` file, it will be listed as so. Sometimes, this might not be desirable and one would want optional dependencies inside the core package itself. Here is an [example](https://github.com/VectorInstitute/cyclops/blob/main/cyclops/models/__init__.py#L13) of that. However for now, assuming no optional dependencies, the core ones would be `flwr`, `numpy`, `opacus`, `pandas`, `pycyclops`, `PyYAML`, `scikit-learn`, `torch`, `torchvision`, `dp-accounting` and `wandb`. Everything else that is used in other directories like `examples`, `research`, etc could be placed under development dependencies i.e., `[tool.poetry.group.dev.dependencies]`.
- Anyways in this PR I'm just proposing the above and made some slight changes towards that, specifically:
    - Removed `tensorflow-privacy` as a dependency since I didn't see this used in the repo directly. Instead i added `dp-accounting` which is the package actually being used (a sub-dependency of `tensorflow-privacy`)
    - Bumped up pandas to 2.0 and above. pandas 2.0 is nice since it supports a pyarrow backend which would be good moving forward. But i did this mainly since `cyclops` now used 2.0 and above. The update should be mostly backwards compatible but good to verify this before merging this PR. So, would need some help from the other folks to check if this doesn't break anything.
    - Bumped up `cyclops` to 0.2.2 and above. `cyclops` currently pulls a lot of dependencies, and its only being used in the feature alignment module. This older version of cyclops is really heavy; installing all kinds of things like monai, torchvision, etc. The current one from `0.2.2` is much more lightweight and offers additional features using extra installs.
    - Fixed the import since the `TabularFeatures` class moved to a unified `data` package in cyclops

When the package installation is heavy, it prevents users from using it. Hence, I hope this PR can help to trim things down. Its mainly caused by cyclops, which we have now fixed after the `0.2.0` release, hence I figured I'd help migrate to that!

# Tests Added


lotif:
That's great, thanks @amrit110 !! I will defer approval to other team members who might be able to verify the pandas upgrade better than me. Thanks for jumping into it and pushing this out.

emersodb (approved):
Since this PR passes the smoke tests, I think we're good to go here. The only part that I think that might be affected is the FLamby research code. However, we currently install those environments from a fixed requirements file to get around a MonAI/FLamby dependency issue anyway.

PR #92, opened 2024-01-10
Add a reporter to save the metrics to a parseable file

lotif:
# PR Type
Feature

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686xytj6

- Adding the `MetricsReporter` class to store and eventually report the training metrics from both server and client
- Hooking it up to the `fit` and `evaluate` methods of server and client

# Tests Added

Added full unit test coverage for all added functionality.

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -267,7 +267,7 @@ class BasicClient(NumPyClient):
         self.metrics_reporter.add_to_metrics_at_round(
             current_server_round,
             data={
-                "metrics": metrics,
+                "fit_metrics": metrics,
                 "loss_dict": loss_dict,
             },
         )
@@ -307,7 +307,7 @@ class BasicClient(NumPyClient):
         self.metrics_reporter.add_to_metrics_at_round(
             current_server_round,
             data={
-                "metrics": metric_values,
+                "evaluate_metrics": metric_values,
                 "loss": loss,
             },
         )
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -59,12 +59,13 @@ class BasicClient(NumPyClient):
         self.metrics = metrics
         self.checkpointer = checkpointer
 
+        self.client_name = self.generate_hash()
+
         if metrics_reporter is not None:
             self.metrics_reporter = metrics_reporter
         else:
-            self.metrics_reporter = MetricsReporter()
+            self.metrics_reporter = MetricsReporter(run_id=self.client_name)
 
-        self.client_name = self.generate_hash()
         self.initialized = False  # Whether or not the client has been setup
         self.model_weights_initialized = False
 
@@ -295,7 +296,7 @@ class BasicClient(NumPyClient):
         if not self.initialized:
             self.setup_client(config)
 
-        _, _, current_server_round = self.process_config(config)
+        current_server_round = self.narrow_config_type(config, "current_server_round", int)
         self.metrics_reporter.add_to_metrics_at_round(
             current_server_round,
             data={"evaluate_start": datetime.datetime.now()},
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -1,4 +1,5 @@
 import copy
+import datetime
 import random
 import string
 from logging import INFO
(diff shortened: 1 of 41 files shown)


fl4health/clients/basic_client.py
```diff
@@ -270,8 +295,23 @@ def evaluate(self, parameters: NDArrays, config: Config) -> Tuple[float, int, Di
         if not self.initialized:
             self.setup_client(config)
 
+        _, _, current_server_round = self.process_config(config)
```
emersodb, line 298: If you just need to extract the `current_server_round` here you could just use
```
current_server_round = self.narrow_config_type(config, "current_server_round", int)
```
As it's slightly more pointed.

fl4health/clients/basic_client.py
```diff
@@ -541,6 +581,8 @@ def setup_client(self, config: Config) -> None:
 
         self.wandb_reporter = ClientWandBReporter.from_config(self.client_name, config)
 
+        self.metrics_reporter.add_to_metrics({"type": "client", "initialized": datetime.datetime.now()})
```
emersodb, line 584: It might be useful to log the "client_name" in the metrics somewhere (i.e. the unique hash)? However, if you don't think that would be helpful, no need.

fl4health/clients/basic_client.py
```diff
@@ -47,13 +50,20 @@ def __init__(
...
+            metrics_reporter (Optional[MetricsReporter], optional): A metrics reporter instance to record the metrics
+                during the execution. Defaults to an instance of MetricsReporter with default init parameters.
         """
 
         self.data_path = data_path
         self.device = device
         self.metrics = metrics
         self.checkpointer = checkpointer
 
+        if metrics_reporter is not None:
```
emersodb, line 62: Maybe this will be answered as I go through, but does this mean we're always going to be dumping metrics, regardless of whether a reporter is provided?

fl4health/clients/basic_client.py
```diff
@@ -47,13 +50,20 @@ def __init__(
...
+            self.metrics_reporter = metrics_reporter
+        else:
+            self.metrics_reporter = MetricsReporter()
```
emersodb, line 65: Each client generates a hash as its `client_name` (see line 66 just below). Do you think we should use that as the `run_id` for the metrics reporter if one isn't provided?

fl4health/clients/basic_client.py
emersodb: I think I see. This will track metrics, but doesn't necessarily write until `dump` is called. I assume that we'll modify our examples to dump the reporter when you refactor the smoke tests?

fl4health/server/base_server.py
```diff
@@ -132,6 +178,15 @@ def evaluate_round(
             if loss_aggregated:
```
emersodb, line 178: I know this isn't your code, but one line 177, I think you can replace `(results, failures)` with just `_`

fl4health/clients/basic_client.py
lotif: I don't think it's necessary. For the file name, when it's saved, it's already given with a UUID if an id is not provided.

fl4health/clients/basic_client.py
lotif: Yes, correct.

fl4health/clients/basic_client.py
lotif: Yes, that's a good idea. Will do.

tests/reporting/test_metrics.py
```diff
@@ -0,0 +1,88 @@
...
+    metrics_reporter.add_to_metrics_at_round(2, test_data_2)
+    assert metrics_reporter.metrics == {
+        "rounds": {
+            2: {**test_data_1, **test_data_2},
+            4: test_data_1,
+        },
+    }
+
+
+def test_metrics_reporter_dump() -> None:
```
emersodb, line 63: pytest supports using a tmp folder that is created and cleaned up on the backend so you don't have to do it yourself. See `tests/checkpointing/test_save_load.py`. Just a suggestion.

fl4health/clients/basic_client.py
emersodb: Fair enough. Wasn't sure if it would be useful for mapping the metrics back to a specific client.

emersodb (approved):
I left some minor comments. Otherwise, I think this looks good to go.

PR #95, opened 2024-01-11
Add federated checkpointing

sanaAyrml:
# PR Type
[Fix]

# Short Description

Implemented federated checkpointing for additional ablation studies. 
Also addressed a minor issue in the evaluation process, as some federated algorithms do not return tuples like the built-in models like moon_base and fenda_base, which could potentially lead to errors.


emersodb (approved):
Looks good to me!

fatemetkl (approved):
Looks good, thanks for addressing the bug!

PR #97, opened 2024-01-15
Smoke tests: Fix random seed on smoke tests and add asserts on results

lotif:
# PR Type
Other

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8686mur37

- Passing a random seed to FedProx, APFL and SCAFFOLD
- Making them save the metrics to a file at the end of their execution
- Adding assertions to their metrics values in the smoke tests

# Tests Added

The smoke tests themselves.


tests/smoke_tests/run_smoke_test.py
```diff
@@ -69,6 +75,23 @@ async def run_smoke_test(
                 dataset_path="examples/datasets/cifar_data/",
                 checkpoint_path="examples/assets/best_checkpoint_fczjmljm.pkl",
                 assert_evaluation_logs=True,
+                seed=42,
+                server_metrics={
+                    "rounds": {
+                        "1": {
+                            "metrics_aggregated": {"val - prediction - accuracy": 0.4744},
+                            "loss_aggregated": (2.001, 0.05), # to override default tolerance, pass it in as a tuple
```
emersodb, line 83: Rather than using a tuple here, which might be a little hard to understand without the comment you put here (i.e. without always replicating that comment), maybe we can just use a nested dictionary? That is
```
"loss_aggregated": {"target_value": 2.001, "custom_tolerance": 0.05},
```

Then in the parsing of `_assert_metrics_dict`, we could do
```
tolerance = DEFAULT_TOLERANCE
if isinstance(value, dict):
    if "custom_tolerance" in value:
        tolerance = value["custom_tolerance"]
    value = value["target_value"]
```
Maybe it belabors the parsing a bit. Just trying to think of a way to make the custom tolerance override a bit more "obvious."

tests/smoke_tests/run_smoke_test.py
```diff
@@ -304,6 +342,70 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
...
+            continue
+
+        with open(file) as f:
+            metrics = json.load(f)
+
+        if metrics["type"] != metric_type.value:
+            continue
+
+        metrics_found = True
+        _assert_metrics_dict(metrics_to_assert, metrics)
```
emersodb, line 370: I may be missing something, so correct me if I'm wrong, but this looks like we're going to compare any metrics json dumped by a client to a single `metrics_to_assert` dictionary. For some of our example, that's definitely fine, because the different clients are loading the same dataset and running the same training cycle. However, there are instances where we might want to test when that isn't the case. That is, each client loads a distinct dataset. So their metrics won't be the same as each other.

Again, correct me if that interpretation is wrong.

emersodb (commented):
Overall it looks really great. Some fairly minor comments. The only one of moderate significance is the assumption that each client will produce matching metrics. In the general case, they won't. So we probably want that flexibility built into the smoke test infra.

tests/smoke_tests/run_smoke_test.py
lotif: What about a class? It would make things even more organized. I thought about it but went with the tuple, but I think it makes sense to do it.
```python
class MetricValue:
    def __init__(self.value, self.tolerance=DEFAULT_TOLERANCE):
        self.value = value
        self.tolerance = tolerance
```
```python
"loss_aggregated": MetricValue(2.001, 0.05),
```

tests/smoke_tests/run_smoke_test.py
```diff
@@ -312,6 +414,57 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
             client_python_path="examples.fedprox_example.client",
             config_path="tests/smoke_tests/fedprox_config.yaml",
             dataset_path="examples/datasets/mnist_data/",
+            seed=42,
+            server_metrics={
```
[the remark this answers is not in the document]
lotif: Makes sense. Will do.

tests/smoke_tests/run_smoke_test.py
lotif: oh well, just doing the json file change and it doesn't make sense to make it a class. Will do the dictionary change.

tests/smoke_tests/run_smoke_test.py
emersodb: Yes, I think that makes sense. We can do another smoke test where two clients load slightly different datasets and include it as a separate test.

emersodb (approved):

PR #98, opened 2024-01-23
Mkmmd loss

sanaAyrml:
# PR Type
[Feature ]

# Short description

This is a preliminary PR just to review implementation of MKMDD loss to check whether it is aligned with [paper](http://www.gatsby.ucl.ac.uk/~gretton/papers/GreSriSejStrBalPonFuk12.pdf) or not.

# Tests Added

NA


fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,177 @@
+# multi-kernel maximum mean discrepancy
+# cao bin, HKUST, China, binjacobcao@gmail.com
+# free to charge for academic communication
```
lotif, line 3: Typo, maybe? `free of charge` makes more sense

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,177 @@
...
+
+import torch
+from qpth.qp import QPFunction
+
+
+class MKMMDLoss(torch.nn.Module):
+    def __init__(
+        self,
+        device: torch.device,
+        gamma_list: List[float] = [
```
lotif, line 15: Mutable values should not be set as default values for parameters, see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments . Not a big problem since on line 32 you immediately make a tensor out of it but it's good to avoid in case that line is changed in the future.

I'm surprised mypy didn't pick up on it, maybe it's smarter than I think it is?

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,177 @@
...
+            1 / 2,
+            1 / 4,
+            1 / 8,
+        ],
+        min_distance: Optional[bool] = False,
+    ) -> None:
+        """Compute the multi-kernel maximum mean discrepancy (MK-MMD) between the source and target domains.
+        Args:
+
+            gamma_list (List[float]): list of length scales for rbf kernels
```
lotif, line 27: Missing some parameters in the docstring.

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,177 @@
...
+            self.sign = 1
+
+    def rbf_multi_kernel(self, X1: torch.Tensor, X2: torch.Tensor) -> torch.Tensor:
+        """
+        Compute the Radial Basis Function (RBF) kernel between two sets of data points.
+
+        Args:
+            X1 (torch tensor): shape (n_samples1, n_features)
+            X2 (torch tensor): shape (n_samples2, n_features)
+            gamma (torch tensor): the kernel parameters
```
lotif, line 46: `gamma` is not one of the parameters of the function.

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,177 @@
...
+        self,
+        XX: torch.Tensor,
+        YY: torch.Tensor,
+        YX: torch.Tensor,
+        XY: torch.Tensor,
+    ) -> torch.Tensor:
+
+        return XX.mean(dim=(1, 2)) + YY.mean(dim=(1, 2)) - XY.mean(dim=(1, 2)) - YX.mean(dim=(1, 2))
+
+    def estimate_η_k(
```
lotif, line 72: I think special characters like `η` should not be used on function or variable names.

fl4health/clients/fenda_client.py
```diff
@@ -120,6 +125,14 @@ def get_cosine_similarity_loss(self, local_features: torch.Tensor, global_featur
         assert len(local_features) == len(global_features)
         return torch.abs(self.cos_sim(local_features, global_features)).mean()
 
+    def get_mkmmd_loss(self, local_features: torch.Tensor, global_features: torch.Tensor) -> torch.Tensor:
+        """
```
emersodb, line 129: This docstring is probably a copy-paste oversight, as it discusses cosine similarity?

fl4health/clients/fenda_client.py
```diff
@@ -217,6 +230,14 @@ def compute_loss(
             total_loss += self.cos_sim_loss_weight * cos_sim_loss
             additional_losses["cos_sim_loss"] = cos_sim_loss
 
+        if self.mkmmd_loss_weight:
```
emersodb, line 233: Another way that the mkmmd loss can be incorporated is actually in the contrastive loss itself. That is, we use it as a replacement for the `cos_sim` measures in the contrastive calculations. It's probably easier to incorporate it into MOON first, as that approach only has two `cos_sim` calculations. We would have different bases for the two positive and negative measures.

emersodb: This comment is no longer relevant. Closing 🙂

fl4health/clients/fenda_client.py
```diff
@@ -108,8 +114,31 @@ def update_before_train(self, current_server_round: int) -> None:
...
+            self.set_optimized_betas(self.mkmmd_loss)
+
         return super().update_before_train(current_server_round)
 
+    def set_optimized_betas(self, mkmmd_loss: MkMmdLoss) -> None:
+        """Set the optimized betas for the MK-MMD loss."""
+        assert isinstance(self.model, FendaModel)
+
+        local_dist = []
```
emersodb, line 126: I'm not quite sure what the suffix `_dist` here is indicating? Perhaps distance? However, these are storing features, so that seems a bit misleading.

fl4health/clients/fenda_client.py
```diff
@@ -118,24 +147,13 @@ def get_cosine_similarity_loss(self, local_features: torch.Tensor, global_featur
...
-        posi = self.cos_sim(features, positive_pairs)
-        logits = posi.reshape(-1, 1)
-        assert len(features) == len(negative_pairs)
-        nega = self.cos_sim(features, negative_pairs)
-        logits = torch.cat((logits, nega.reshape(-1, 1)), dim=1)
-        logits /= self.temperature
-        labels = torch.zeros(features.size(0)).to(self.device).long()
-
-        return self.ce_criterion(logits, labels)
+        assert len(local_features) == len(global_features)
```
emersodb, line 155: We probably want to assert that the shape of these features is equal, not just the batch size?

fl4health/clients/fenda_client.py
```diff
@@ -46,13 +49,16 @@ def __init__(
...
         self.perfcl_loss_weights = perfcl_loss_weights
         self.cos_sim_loss_weight = cos_sim_loss_weight
         self.contrastive_loss_weight = contrastive_loss_weight
-        self.cos_sim = torch.nn.CosineSimilarity(dim=-1).to(self.device)
-        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
+        self.mkmmd_loss_weight = mkmmd_loss_weight
         self.temperature = temperature
+        self.cos_sim = torch.nn.CosineSimilarity(dim=-1).to(self.device)
+        self.contrastive_loss = ContrastiveLoss(self.device, temperature=self.temperature).to(self.device)
+        self.mkmmd_loss = MkMmdLoss(device=self.device, minimize_type_two_error=False).to(self.device)
```
emersodb, line 61: It's probably worth experimenting with this a bit, but I sort of imagined we'd have at least two MkMMD losses to more closely mimic the PerFCL setup. That is, on MkMMD loss to ensure the new global features don't move too far from the old global features (we would minimize the type two error there) and one to make sure the new global features and the new local features were far from each other (as you're doing here). We could theoretically add a third that tried to make sure the the new local features don't drift too far from the old local features.

fl4health/clients/fenda_client.py
```diff
@@ -144,10 +162,10 @@ def get_contrastive_loss(
         Compute contrastive over current local features (z_p) with old local features (hat{z_p}) as positive pairs
         and current global features (z_s) as negative pairs.
         """
-        return self.compute_contrastive_loss(
+        return self.contrastive_loss(
             features=local_features,  # (z_p)
-            positive_pairs=old_local_features,  # (\hat{z_p})
-            negative_pairs=global_features,  # (z_s)
+            positive_pairs=old_local_features.unsqueeze(0),  # (\hat{z_p})
```
emersodb, line 167: I may be missing something, but why do we need to inject this unsqueeze here now?

fl4health/clients/moon_client.py
```diff
@@ -35,10 +38,11 @@ def __init__(
             loss_meter_type=loss_meter_type,
             checkpointer=checkpointer,
         )
-        self.cos_sim = torch.nn.CosineSimilarity(dim=-1).to(self.device)
-        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
         self.contrastive_weight = contrastive_weight
+        self.mkmmd_loss_weights = mkmmd_loss_weights
         self.temperature = temperature
```
emersodb, line 43: Since temperature is just being used to initialize the ContrastiveLoss class, I don't think we need to keep it around as a class variable?

fl4health/clients/fenda_client.py
```diff
@@ -46,13 +49,16 @@ def __init__(
...
             contrastive_loss_weight: Weight to be used for contrastive loss.
+            mkmmd_loss_weight: Weight to be used for mkmmd loss.
         """
         self.perfcl_loss_weights = perfcl_loss_weights
         self.cos_sim_loss_weight = cos_sim_loss_weight
         self.contrastive_loss_weight = contrastive_loss_weight
-        self.cos_sim = torch.nn.CosineSimilarity(dim=-1).to(self.device)
-        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
+        self.mkmmd_loss_weight = mkmmd_loss_weight
         self.temperature = temperature
```
emersodb, line 58: Since temperature is just being used to initialize the ContrastiveLoss class, I don't think we need to keep it around as a class variable?

fl4health/clients/moon_client.py
```diff
@@ -106,8 +89,36 @@ def update_before_train(self, current_server_round: int) -> None:
...
         return super().update_before_train(current_server_round)
 
+    def set_optimized_betas(
+        self, mkmmd_loss: MkMmdLoss, old_model: torch.nn.Module, global_model: torch.nn.Module
+    ) -> None:
+        """Set the optimized betas for the MK-MMD loss."""
+        assert isinstance(old_model, MoonModel)
+        assert isinstance(global_model, MoonModel)
+
+        old_dist = []
```
emersodb, line 104: A similar comment to the FENDA client, where I think "_dist" might be a misleading suffix?

fl4health/clients/moon_client.py
```diff
@@ -106,8 +89,36 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        global_dist = []
+
+        # Compute the old features before aggregation and global features
+        old_model.eval()
```
emersodb, line 108: Perhaps, rather than explicitly setting these to eval here (they should be eval after our clone and freeze) maybe we can do 
``` 
assert not old_model.training
assert not global_model.training
```
If these fail then we know something is wrong.

fl4health/clients/moon_client.py
```diff
@@ -135,13 +146,22 @@ def compute_loss_and_additional_losses(
...
+        if self.contrastive_weight and "old_features" in features:
+            contrastive_loss = self.contrastive_loss(
+                features["features"], features["global_features"].unsqueeze(0), features["old_features"]
             )
             total_loss += self.contrastive_weight * contrastive_loss
             additional_losses["contrastive_loss"] = contrastive_loss
 
+        elif self.mkmmd_loss_weights and "old_features" in features:
+            min_mkmmd_loss = self.mkmmd_loss(features["features"], features["global_features"])
+            max_mkmmd_loss = self.mkmmd_loss(features["features"], features["old_features"][-1])
```
emersodb, line 158: If we're going to be using an MkMMD loss to push the new local features away from the old local features, we likely want to have a different set of optimized betas. Remember that when we optimized them above, we're finding the kernels that best distinguish the between the old local model and the aggregated model. We would want to find the kernels that have the hardest time distinguishing them for the `max_mkmmd_loss`.

As a side note, I would be interested in knowing how well we perform if we just try to keep the local features close to the global features. I.e. only having the one loss.

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,280 @@
...
+    def beta_with_extreme_kernel_base_values(
+        self, hat_d_per_kernel: torch.Tensor, hat_Q_k: torch.Tensor, minimize_type_two_error: bool = True
+    ) -> torch.Tensor:
+        kernel_base_values = torch.tensor(
+            [hat_d_per_kernel[i] / hat_Q_k[i][i] for i in range(len(hat_d_per_kernel))]
+        ).to(self.device)
+        if minimize_type_two_error:
+            log(
+                INFO,
+                "Rather than optimizing, we select a single kernel with largest hat_d_k",
```
emersodb, line 169: We can update this log string and the one below to be `hat_d_k/hat_Q_k_lambda`

fl4health/clients/fenda_client.py
sanaAyrml: Distribution, because input for mkmmd loss is distribution

fl4health/clients/fenda_client.py
sanaAyrml: That is better assertion

fl4health/clients/fenda_client.py
sanaAyrml: I agree I am just experimenting various option for fenda and moon

fl4health/clients/fenda_client.py
sanaAyrml: I have implemented contrastive loss based on MOON implementation, where the negative pair of a single sample can be multiple samples based on the list of saved previous models. For the consistency in design, the contrastive loss here gets a tensor of (n_pairs ,n_samples, n_features) as negative_pairs and a tensor of (1, n_samples, n_features) as positive pairs as each sample can only have one positive pair. Such a squeeze here is to fit in these shape sizes.

fl4health/clients/moon_client.py
sanaAyrml: That is a fair point

fl4health/clients/fenda_client.py
sanaAyrml: same here

PR #100, opened 2024-01-26
Splitting Loss into EvaluationLoss and TrainLoss

lotif:
# PR Type
[Fix]

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8685q1dbw

- Making `Loss` into an abstract class and splitting its code into `TrainingLoss` and `EvaluationLoss`
- Making `backward` exist only in `TrainingLoss` and `checkpoint` exist only in `EvaluationLoss`
- Splitting `BasicClient`'s `compute_loss` into `compute_training_loss` and `compute_evaluation_loss` and making them return their respective subclass instances.
- Changing all clients accordingly

# Tests Added

Transformed the `Losses` test into `TrainingLosses` and `EvaluationLosses` tests, and also fixed the existing tests.


lotif:
@emersodb please check. This is one idea I had, not a big fan of it so I am open to better suggestions.

fl4health/utils/losses.py
```diff
@@ -2,12 +2,47 @@
```
emersodb, line 2: I think maybe we want to make an explicit EvalLosses class that inherits from Losses, similar to TrainLosses to force and explicit "fork." One thing that I think can be in EvalLosses but not in TrainLosses is checkpoint. It's possible that a person might want to checkpoint in the training loop, but they can do that explicitly through additional_losses if they really want. I think checkpointing during validation is what makes the most sense to support by default.

fl4health/clients/basic_client.py
```diff
@@ -625,8 +626,12 @@ def predict(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[st
             raise ValueError("Model forward did not return a tensor or dictionary or tensors")
 
     def compute_loss(
```
emersodb, line 628: I see some advantage to using the boolean here. If we do that, the user need only define one `compute_loss` function. 

However, I wonder what you think have having a `compute_training_loss` and `compute_eval_loss`. This would allow for more explicit typing and specific workflows, like only constructing backward or checkpoint when necessary. In the more complex overrides, code duplication could be reduce by defining a unified `compute_loss` function that both call, for example in `fenda_client.py`

fl4health/utils/losses.py
```diff
@@ -128,26 +152,36 @@ def aggregate_loss_by_type(
...
-            if loss_meter_type == LossMeterType.AVERAGE:
-                loss = loss / num_losses
-            loss_dict[key] = loss
-
-        return loss_dict
+        additional_loss_list = [losses.additional_losses for losses in losses_list]
+        backward_loss_list = None
+        # Assuming the losses are all the same type, check for the type of the first one only
+        if len(losses_list) > 0 and isinstance(losses_list[0], TrainLosses):
+            backward_loss_list = [losses.backward for losses in losses_list]  # type: ignore
```
emersodb, line 163: Perhaps for each of the Losses classes we can have a class function that takes all of the losses present and rolls them into lists that can all be processed by this aggregation and then "reconstituted from the lists thereafter. That way we could just call these class functions to deconstruct, aggregate, and reconstruct by just calling something like Losses.deconstruct(...) and Losses.reconstruct(...).

emersodb: We can talk about this in a bit more detail if you'd like.

lotif: That makes a lot of sense.

lotif:
@emersodb thanks for the comments. I will refactor it further and reach out again.

lotif:
@emersodb Code changes are done, thanks for the feedback! I still need to add the docstrings, which I'll do tomorrow, but the code itself is ready for review.

fl4health/clients/basic_client.py
```diff
@@ -391,23 +391,23 @@ def train_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Losses,
...
         # Clear gradients from optimizer if they exist
         self.optimizers["global"].zero_grad()
 
         # Call user defined methods to get predictions and compute loss
         preds, features = self.predict(input)
-        losses = self.compute_loss(preds, features, target)
+        losses = self.compute_training_loss(preds, features, target)
 
-        # Compute backward pass and update paramters with optimizer
+        # Compute backward pass and update parameters with optimizer
```
emersodb, line 404: Nice! 😂

fl4health/utils/losses.py
```diff
@@ -84,18 +204,20 @@ def clear(self) -> None:
         """
         raise NotImplementedError
 
-    @abstractmethod
     def compute(self) -> Losses:
         """
-        Compute aggregation of current list of losses if non-empty.
+        Compute average of current list of losses if non-empty.
```
emersodb, line 209: This isn't always the average if the meter type is just accumulation, which is why I think we used the term aggregation here originally.

fl4health/utils/losses.py
```diff
@@ -84,18 +204,20 @@ def clear(self) -> None:
...
 
         Returns:
-            Losses: New Losses object with aggregate of losses in losses_list.
-
-        Raises:
-            NotImplentedError: To be impemented by child class.
+            Losses: New Losses object with average of losses in losses_list.
```
emersodb, line 212: Similarly average vs. aggregate here.

fl4health/utils/losses.py
```diff
@@ -84,18 +204,20 @@ def clear(self) -> None:
...
         """
-        raise NotImplementedError
+        assert len(self.losses_list) > 0
+
+        # assume the type of the first loss object is the type of all loss objects
+        if isinstance(self.losses_list[0], EvaluationLosses):
```
emersodb, line 217: This is sort of a work around because we're allowing LossMeter to be "loosely" type in that it handles "Losses" objects. This might be more trouble than it's worth, but I feel like the "proper" thing for us to do is generically type the LossMeter with a typevar T that has a bound of Losses, then when we construct a `self.train_loss_meter = LossMeter.get_meter_by_type(loss_meter_type)`
and the validation analogue, we provide the type TrainingLosses and EvaluationLosses to each respectively.

Then this becomes just
```
T.aggregate(self)
```

There might be unforseen consequences of that. So let me know what you think?

fl4health/utils/losses.py
```diff
@@ -216,22 +301,20 @@ def clear(self) -> None:
...
-        checkpoint_loss = torch.sum(torch.FloatTensor([losses.checkpoint for losses in self.losses_list]))
-        backward_loss = LossMeter.aggregate_loss_by_type(
-            self.losses_list, loss_type=LossType.BACKWARD, loss_meter_type=LossMeterType.ACCUMULATION
-        )
-        additional_loss = LossMeter.aggregate_loss_by_type(
-            self.losses_list, loss_type=LossType.ADDITIONAL, loss_meter_type=LossMeterType.ACCUMULATION
-        )
-        losses = Losses(backward=backward_loss, checkpoint=checkpoint_loss, additional_losses=additional_loss)
-        return losses
+def _aggregate_losses_dict(
```
emersodb, line 305: I'm not 100% sure on the "proper" design, so please disagree if the following seems silly. However, since this is working specifically on "Loss Dictionaries", should we make this a `@staticmethod` within the Losses class rather than a "file-wide" function?

fl4health/clients/evaluate_client.py
```diff
@@ -157,21 +157,21 @@ def validate_on_model(
...
                 preds = {"prediction": model(inputs)}
-                losses = self.compute_loss(preds, {}, targets)
+                losses = self.compute_evaluation_loss(preds, {}, targets)
 
                 metric_meter.update(preds, targets)
                 loss_meter.update(losses)
 
         metrics = metric_meter.compute()
-        losses = loss_meter.compute()
+        losses = cast(EvaluationLosses, loss_meter.compute())
```
emersodb, line 166: If we do the generic typing refactor that I suggested below for the LossMeter class, I don't think this cast will be necessary, as .compute() will return the type T of the loss meter.

However, even if we don't do the generic typing refactor, I think we probably want to do
```
losses = loss_meter.compute()
assert isinstance(losses, EvaluationLosses)
```
rather than a `cast` for mypy. If I recall correctly, cast doesn't actually do any checking that the underlying object is actually of type EvaluationLosses. It essentially tells mypy to ignore any issues. The isinstance is a verification of the type and mypy infers the type from there.

I may be wrong though.

fl4health/clients/fed_prox_client.py
```diff
@@ -108,14 +121,14 @@ def compute_loss(
...
         """
-        loss = self.criterion(preds["prediction"], target)
-        proximal_loss = self.get_proximal_loss()
-        total_loss = loss + proximal_loss
-        losses = Losses(checkpoint=loss, backward=total_loss, additional_losses={"proximal_loss": proximal_loss})
-        return losses
+        loss, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
+        total_loss = loss + additional_losses["proximal_loss"]
+        # adding the vanilla loss to the additional losses to be used by update_after_train
+        additional_losses["loss"] = loss
```
emersodb, line 130: Nice catch!

emersodb: In this example, I wonder if it makes more sense to just override `compute_training_loss` and `compute_evaluation_loss` themselves such that the body of 
`compute_training_loss`
```
loss = self.criterion(preds["prediction"], target)
proximal_loss = self.get_proximal_loss()
additional_losses = {"proximal_loss": proximal_loss, "loss": loss}
return TrainingLosses(backward=loss + proximal_loss, additional_losses=additional_losses)
```
and the body of 
`compute_evaluation_loss`
```
loss = self.criterion(preds["prediction"], target)
return EvaluationLosses(checkpoint=loss, additional_losses={})
```
The reason is that proximal loss doesn't really make sense for the validation set (it's always zero). Just a thought.

fl4health/clients/fenda_client.py
```diff
@@ -243,6 +232,30 @@ def compute_loss(
...
+            features: (Dict[str, torch.Tensor]): Feature(s) of the model(s) indexed by name.
+            target: (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            EvaluationLosses: an instance of EvaluationLosses containing checkpoint loss and additional losses
+                indexed by name. Additional losses may include cosine_similarity_loss, contrastive_loss
+                and perfcl_loss.
+        """
+        _, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
+        loss = self.criterion(preds["prediction"], target)
```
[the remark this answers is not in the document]
emersodb: Since we're already computing the `total_loss`, we might as well add it to `additional_losses` rather than discarding it?

fl4health/clients/moon_client.py
```diff
@@ -117,19 +141,38 @@ def compute_loss(
...
+            target: (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            EvaluationLosses: an instance of EvaluationLosses containing checkpoint loss and
+                additional losses indexed by name.
+        """
+        if len(self.old_models_list) == 0:
+            return super().compute_evaluation_loss(preds, features, target)
+
+        loss = self.criterion(preds["prediction"], target)
```
emersodb, line 176: Similar to the FENDA example, perhaps you can just store the loss in additional losses and extract from there. You could also pull out the total loss and put it into additional losses as well.

fl4health/utils/losses.py
```diff
@@ -2,15 +2,103 @@
...
+        """
+        Aggregates the losses in the given LossMeter into an instance of EvaluationLosses
+
+        Args:
+            loss_meter (LossMeter): The loss meter object with the collected losses.
+
+        Returns:
+            EvaluationLosses: An instance of EvaluationLosses with the aggregated losses.
+        """
+        assert all([isinstance(losses, EvaluationLosses) for losses in loss_meter.losses_list])
```
emersodb, line 86: I think, though I'm not certain, that if we generically type the LossMeter, we can also drop this? I'm not 100% sure though.

fl4health/clients/basic_client.py
jewelltaylor: Any spelling errors in the repo we can assume came from 😭

fl4health/utils/losses.py
lotif: I think this is a great idea. Maybe the only thing is that we will probably need 2 loss meters, one for evaluation and one for training, for each client.

fl4health/utils/losses.py
lotif: Yes, that's a better idea. Will do.

fl4health/clients/fed_prox_client.py
lotif: That makes sense, will do.

fl4health/utils/losses.py
lotif: Yes, I think this can be dropped once we do the generics change, which is also one more reason to do it.

fl4health/utils/losses.py
emersodb: If I remember correctly, I think basic client has 2 loss meters already 🤔.
```
self.train_loss_meter = LossMeter.get_meter_by_type(loss_meter_type)
self.val_loss_meter = LossMeter.get_meter_by_type(loss_meter_type)
```
So I'm hoping it sort of seamlessly drops in. However, definitely let me know if this creates a massive headache.

fl4health/utils/losses.py
```diff
@@ -2,15 +2,103 @@
...
+            checkpoint_loss /= len(loss_meter.losses_list)
+
+        additional_losses_list = [losses.additional_losses for losses in loss_meter.losses_list]
+        additional_losses_dict = _aggregate_losses_dict(additional_losses_list, loss_meter.get_type())
+
+        return EvaluationLosses(checkpoint=checkpoint_loss, additional_losses=additional_losses_dict)
+
+
+class TrainingLosses(Losses):
     def __init__(
```
jewelltaylor, line 100: Very minor documentation detail flagging just in case: "A class to store the checkpoint, backward and additional_losses of a model along with a method to return a dictionary representation." 

We can remove checkpoint because the new TrainingLosses has removed that from what I can see

fl4health/utils/losses.py
lotif: I just found out it's impossible in python to do `T.aggregate(self)` because on runtime the real type of `T` is lost, there is currently an open issue about it:
https://github.com/python/typing/issues/629
I will need to store the type in the class so the solution will be a bit uglier but I still think it's worth it.

emersodb: Oh very interesting. We're on the cutting edge of generics usage 😂. I agree, storing the type is a little less fun, but hopefully not too much uglier.

fl4health/clients/fed_prox_client.py
lotif: I did it a bit differently. please check.

fl4health/clients/fenda_client.py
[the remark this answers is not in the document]
lotif: Not sure if what I did is what you had in mind, please check.

fl4health/clients/moon_client.py
lotif: Not sure if what I did is what you had in mind, please check.

lotif:
@jewelltaylor @emersodb all comments have been addressed and I have added the missing docstrings, please check again when you have a minute.

emersodb:
> @jewelltaylor @emersodb all comments have been addressed and I have added the missing docstrings, please check again when you have a minute.

It's next on my list of things to tackle today 🙂

fl4health/clients/fed_prox_client.py
emersodb: Looks good to me.

fl4health/utils/losses.py
```diff
@@ -165,156 +162,74 @@ class LossMeterType(Enum):
...
-        # assume the type of the first loss object is the type of all loss objects
-        if isinstance(self.losses_list[0], EvaluationLosses):
-            return EvaluationLosses.aggregate(self)
 
-        return TrainingLosses.aggregate(self)
-
-    @classmethod
-    def get_meter_by_type(cls, meter_enum: LossMeterType) -> LossMeter:
+class LossMeter(Generic[LossesType]):
+    def __init__(self, loss_meter_type: LossMeterType, losses_type: type[Losses]) -> None:
```
emersodb, line 169: I might be wrong, but can `losses_type` here actually be annotated losses_type: LossesType? That is, it must be the same time as the generic for LossMeter?

tests/clients/test_fenda_client.py
```diff
@@ -85,12 +85,15 @@ def test_computing_loss(get_fenda_client: FendaClient) -> None:  # noqa
         "old_global_features": old_global_features,
         "aggregated_global_features": aggregated_global_features,
     }
+
     training_loss = fenda_client.compute_training_loss(preds=preds, target=target, features=features)
     evaluation_loss = fenda_client.compute_evaluation_loss(preds=preds, target=target, features=features)
     assert isinstance(training_loss.backward["backward"], torch.Tensor)
     assert pytest.approx(0.8132616, abs=0.0001) == evaluation_loss.checkpoint.item()
-    assert evaluation_loss.checkpoint.item() != training_loss.backward["backward"].item()
```
emersodb: Do you mind if we reinstate a check like this? It could simply be 
```
assert training_loss.additional_losses["loss"].item() == training_loss.backward["backward"].item()
```
We previously had a bit of a nasty bug where the two tensors shared memory and inadvertently ended up being the same value. This ensures that isn't happening.

emersodb (approved):
@lotif: Looks pretty much ready to go. They only thing I would strongly suggest considering is adding back in the small comparison I suggested to avoid a bit of a nuanced but from creeping in again. The other comment I left is more minor.

fl4health/utils/losses.py
lotif: good point, thanks

lotif:
@emersodb those were some great catches, thanks! I've fixed them and will merge as soon as tests pass.

PR #101, opened 2024-01-29
Picai example

jewelltaylor:
# PR Type

First pass at adding support for experimentation on the [PICAI Dataset](https://www.google.com/search?q=picai+grand+challenge&oq=PICAI+gran&gs_lcrp=EgZjaHJvbWUqBwgAEAAYgAQyBwgAEAAYgAQyBggBEEUYQDIGCAIQRRg5MgYIAxBFGDsyBggEEEUYPKgCALACAA&sourceid=chrome&ie=UTF-8). Specifically, there are scripts to train a U-Net on the PICAI dataset using both a centralized and a federated setup via FedAvg. As part of this PR, I have also downloaded and preprocessed the PICAI dataset. I have included a README.md describing the raw and preprocessed datasets and where they are stored on the Vector cluster.

Since well-performing models on the PICAI dataset tend to be trained for large epoch counts, a central consideration of this PR was to be resilient to pre-emption (and in the future broader fault tolerance). To this end, I defined a simple `PicaiServer` and `PicaiClient` endowed with a custom `PerRoundCheckpointer` to save server and client state. In a future PR, I will work to include the logic in our base server and base client to add fault tolerance but I will have to still think of some aspects of the design.

The focus of this PR is to get a simple pipeline for centralized and federated training on the PICAI dataset. Next, I will work on a more intelligent transformation pipeline, scaling up the data loading to multiple workers, exploring more recent 3D segmentation models and federated learning approaches.

Clickup Ticket(s): [Train and Evaluate baseline model on PICAI with centralized and federated setup](https://app.clickup.com/t/8686t9ztp)

# Tests Added
- Tests added to custom per round checkpointers that I defined


research/picai/fedavg/server.py
```diff
@@ -0,0 +1,91 @@
...
+from flwr.common.logger import log
+from flwr.server.client_manager import SimpleClientManager
+from flwr.server.strategy import FedAvg
+
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.config import load_config
+from research.picai.fl_utils import (
+    evaluate_metrics_aggregation_fn,
+    fit_config,
+    fit_metrics_aggregation_fn,
```
fatemetkl, line 16: Can we instead directly use `examples.simple_metric_aggregation` for `evaluate_metrics_aggregation_fn` and `fit_metrics_aggregation_fn`?
And instead of  `get_initial_model_parameters` can we use `fl4health.utils.functions import get_all_model_parameters`? In that case, we probably would not need  to have`fl_utils.py`

research/picai/picai_server.py
```diff
@@ -0,0 +1,164 @@
...
+from flwr.server.strategy import Strategy
+
+from fl4health.checkpointing.checkpointer import ServerPerRoundCheckpointer, TorchCheckpointer
+from fl4health.parameter_exchange.parameter_exchanger_base import ParameterExchanger
+from fl4health.reporting.fl_wanb import ServerWandBReporter
+from fl4health.server.base_server import FlServerWithCheckpointing
+from research.picai.fl_utils import get_initial_model_parameters
+
+
+class PicaiServer(FlServerWithCheckpointing):
```
fatemetkl, line 19: I think this is a very nice extension of `FlServerWithCheckpointing`!  I was wondering if there is a way to have this included in the core library as a pre-emption resistant server (either in this PR or in a follow-up ticket).

fatemetkl (commented):
This PR is a great research addition to the library, and `PerRoundCheckpointer` is very useful and practical for training larger models! I had some comments but mostly are minor and some suggestions for improvement.

research/picai/losses.py
```diff
@@ -0,0 +1,52 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+class FocalLoss(nn.Module):
```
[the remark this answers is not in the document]
jewelltaylor: I think I could also inherit from _Loss, but usually classes or functions with _ prepended are meant to be private and not accessed externally. To be consistent, I will change it to _Loss for this PR but I may update this in an upcoming one!

research/picai/fedavg/client.py
```diff
@@ -0,0 +1,121 @@
...
+        metrics: Sequence[Metric],
+        device: torch.device,
+        intermediate_checkpoint_dir: Path,
+        overviews_dir: Path,
+    ) -> None:
+
+        super().__init__(data_path, metrics, device, intermediate_checkpoint_dir=intermediate_checkpoint_dir)
+        self.overviews_dir = overviews_dir
+
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
```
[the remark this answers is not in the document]
jewelltaylor: Yeah for sure, that is a great idea! I have added a naive partitioning of the dataset based on the number of clients where each sample gets randomly assigned to one of the clients. In a follow up PR, I will aim to have more interesting ways of splitting the client dataset. Unfortunately, we do not have access to the site that a sample was captured at but we have information about the MRI machine used to capture it, so we can split on that or other variables we have access too.

fl4health/utils/config.py
```diff
@@ -4,7 +4,6 @@
 
 REQUIRED_CONFIG = {
     "n_server_rounds": int,
-    "n_clients": int,
```
jewelltaylor, line 7: I remove this as a required config param because I wanted to have the number of clients to be dynamic and passed in the launching script. If we put it in a config, then we cannot (easily) check the value of n_clients from the launch script to dynamically launch the given number of clients.

research/picai/picai_server.py
jewelltaylor: Thanks for the great feedback! I will definitely have a follow up PR to incorporate this into the main. I decided not to do that for this PR, just to not tackle too much at once

research/picai/losses.py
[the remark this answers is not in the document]
fatemetkl: That makes a lot of sense! 
Another suggestion for a future PR is to dedicate a loss module in fl4health where we keep all these loss implementations (there is another one for preprocessing.autoencoders for example). Do you think it would make sense to move FocalLoss under the main library as well?

research/picai/fedavg/server.py
```diff
@@ -0,0 +1,120 @@
...
+from flwr.server.strategy import FedAvg
+
+from examples.simple_metric_aggregation import evaluate_metrics_aggregation_fn, fit_metrics_aggregation_fn
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.config import load_config
+from research.picai.model_utils import get_model
+from research.picai.picai_server import PicaiServer
+
+
+def get_initial_model_parameters(client_model: nn.Module) -> Parameters:
```
fatemetkl, line 21: Minor: this can be imported from `fl4health.utils.functions`

research/picai/central/README.md
```diff
@@ -0,0 +1,14 @@
+# Running Centralized Example
+
+The following instructions outline training and validating a simple U-Net model on the Preprocessed Dataset described in the [PICAI Documentation](/research/picai/README.md) using a centralized setup. An example of the usage is below. Note that the script needs to be run from the top level of the FL4Health repository. Moreover, a python environment with the required libraries must already exist.  See the main PICAI documentation Cluster [PICAI Documentation](/research/picai/README.md) for instructions on creating and activating environment required to exectute the following code. The commands below should be run from the top level directory:
+
+```bash
+./research/picai/central/launch.slrm folder_for_artifacts/ path_to_desired_venv/ fold_id
```
fatemetkl, line 6: Super minor: since we are passing run_name now, should we include it in the example here?

fl4health/utils/config.py
fatemetkl: Makes sense to me!

research/picai/fedavg/client.py
```diff
@@ -0,0 +1,121 @@
...
+            get_img_transform(),
+            get_seg_transform(),
+            num_workers=0,
+        )
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        return get_model(device=self.device)
+
+    def get_criterion(self, config: Config) -> nn.Module:  # type: ignore
```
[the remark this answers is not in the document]
fatemetkl: I think it makes sense to do this (maybe in an upcoming PR).

fatemetkl (approved):
All my comments are addressed, just some minor things left that should be very quick. Thanks John for the changes, everything looks good to me!

PR #102, opened 2024-01-29
Proposed changes to the mkmmd loss function

emersodb:
This is a proposed refactor to the mkmmd loss function constructions. There are a lot of tests here and I can definitely improve the documentation. However, I wanted to get it in front of you ASAP.

Note: Most of the other files are miscellaneous stuff that was failing for me in mypy, black, or flake.

### DIFF
--- a/fl4health/losses/mkmmd_loss.py
+++ b/fl4health/losses/mkmmd_loss.py
@@ -247,7 +247,7 @@ class MkMmdLoss(torch.nn.Module):
             unnormalized_betas = self.form_and_solve_qp(hat_d_per_kernel, regularized_Q_k)
             return (1.0 / torch.sum(unnormalized_betas)) * unnormalized_betas
         else:
-            # If we're trying to maximize the type II error, then we are trying to maximize a concave function over a
+            # If we're trying to maximize the type II error, then we are trying to maximize a convex function over a
             # convex polygon of beta values. So the maximum is found at one of the vertices
             return self.get_best_vertex_for_objective_function(hat_d_per_kernel, regularized_Q_k)
 
--- a/fl4health/losses/mkmmd_loss.py
+++ b/fl4health/losses/mkmmd_loss.py
@@ -2,153 +2,202 @@
 # cao bin, HKUST, China, binjacobcao@gmail.com
 # free to charge for academic communication
 
+from logging import INFO
 from typing import Optional
 
 import torch
-from qpth.qp import QPFunction
+from flwr.common.logger import log
+from qpth.qp import QPFunction, QPSolvers
 
 
-class MKMMDLoss(torch.nn.Module):
+class MkMmdLoss(torch.nn.Module):
     def __init__(
         self,
         device: torch.device,
-        gamma_list: Optional[torch.Tensor] = None,
-        min_distance: Optional[bool] = False,
+        gammas: Optional[torch.Tensor] = None,
+        betas: Optional[torch.Tensor] = None,
+        minimize_type_two_error: bool = True,
     ) -> None:
-        """Compute the multi-kernel maximum mean discrepancy (MK-MMD) between the source and target domains.
-        Args:
+        """
+        Compute the multi-kernel maximum mean discrepancy (MK-MMD) between the source and target domains. Also allows
+        for optimization of the coefficients, beta
 
-            device (torch.device): device to use for computation
-            gamma_list (List[float]): list of length scales for rbf kernels
-            min_distance (bool): if True, the MK-MMD is computed as the minimum distance between the source and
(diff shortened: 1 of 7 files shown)


fl4health/losses/mkmmd_loss.py
```diff
@@ -2,153 +2,202 @@
...
+        # all_h_u_per_vi has shape number kernels x number of v_is. Each row is h_u(v_i) for kernel u with the ith
+        # column being that kernel evaluated on v_i.
+        # output shape is number of kernels x 1
+        return torch.mean(all_h_u_per_v_i, dim=1, keepdim=True)
+
+    def compute_mkmmd(self, X: torch.Tensor, Y: torch.Tensor, beta: torch.Tensor) -> torch.Tensor:
+        # In this function, we assume that X, Y: n_samples, n_features are the same size and that beta is a tensor of
+        # shape number of kernels x 1
+        all_h_u_per_vi = self.compute_all_h_u_per_v_i(X, Y)
+        hat_d_per_kernel = self.compute_hat_d_per_kernel(all_h_u_per_vi)
```
sanaAyrml, line 128: Do we want to also compute estimated hat_d for mlmmd computation? In my original code I basically computed over all pairs and only used estimation (pairing i and i+1) for optimizing beta.
```
def compute_mmd(
        self,
        XX: torch.Tensor,
        YY: torch.Tensor,
        YX: torch.Tensor,
        XY: torch.Tensor,
    ) -> torch.Tensor:

        return XX.mean(dim=(1, 2)) + YY.mean(dim=(1, 2)) - XY.mean(dim=(1, 2)) - YX.mean(dim=(1, 2))
```

fl4health/losses/mkmmd_loss.py
```diff
@@ -167,12 +216,50 @@ def forward(self, Xs: torch.Tensor, Xt: torch.Tensor) -> torch.Tensor:
...
+        # check to see that at least one of them is positive. If none of them are positive, then select a single kernel
+        # with largest hat_d, similar to the suggestion of Gretton et al. in "Optimal Kernel Choice for Large-Scale
+        # Two-Sample Tests", 2012
+        if not torch.any(hat_d_per_kernel > 0):
+            log(
+                INFO,
+                f"None of the estimates for hat_d are positive: {hat_d_per_kernel}. Rather than optimizing, we select "
+                "a single kernel with largest hat_d_k",
+            )
+            return self.beta_with_largest_hat_d(hat_d_per_kernel)
```
sanaAyrml, line 239: shouldn't we return beta with smallest hat d if we want to maximize the distance between two distributions? (minimize_type_two_error=False)

fl4health/losses/mkmmd_loss.py
```diff
@@ -167,12 +216,50 @@ def forward(self, Xs: torch.Tensor, Xt: torch.Tensor) -> torch.Tensor:
...
         #   h is (kernel_num, 1)
         #   A is (1, kernel_num)
-        #   b is ()
-        beta = self.sign * QPFunction(verbose=False)(Q, p, G, h, A, b)
+        #   b is a scalar
+
+        # QPFunction returns betas in the shape 1 x num_kernels to we take the transpose for consistency
+        return QPFunction(verbose=False, solver=QPSolvers.CVXPY)(regularized_Q_k, p, G, h, hat_d_per_kernel.t(), b).t()
+
+    def optimize_betas(self, X: torch.Tensor, Y: torch.Tensor, lambda_m: float) -> torch.Tensor:
```
sanaAyrml, line 224: Maybe have a default value of 1e-5 for  lambda_m?

fl4health/losses/mkmmd_loss.py
sanaAyrml: or maybe we should first the condition `if self.minimize_type_two_error` ?

fl4health/losses/mkmmd_loss.py
emersodb: hmmm...that's true about the single batch. I'd be good with forming XX, YY, XY for batch mkmmd calculations. I think that's a good suggestions.

fl4health/losses/mkmmd_loss.py
emersodb: Yes, you're right. We should only return this if we're minimizing the type two error. Good call.

emersodb: Otherwise, we're still good to make it to the else statement below.

fl4health/losses/mkmmd_loss.py
emersodb: Yup that's good with me.

PR #103, opened 2024-01-30
Sparse Parameter Exchange

yc7z:
# PR Type
[Feature]

# Short Description

The main feature of this PR is the implementaiton of a sparse parameter exchange mechanism, which allows each client to select an arbitrary subset of parameters to be exchanged with the server at the end of each training round according to some sparsity level and selection criterion. The design of the parameter exchanger is meant to allow for flexibility in choosing the selection criterion function. A simple example that demonstrates the usage of the new exchanger is included as well.

This PR also implements some selection criteria functions that were considered in the [super-mask paper](https://arxiv.org/abs/1905.01067), which I want to test on the Flamby datasets.

Finally, I redesigned/refactored the dynamic layer exchanger component, and fixed the bug for the partial exchange client class.

Clickup Ticket(s): [Link](https://app.clickup.com/t/860pettx7)

# Tests Added

Tests have been added to test the sparse exchanger and the corresponding strategy.


fl4health/clients/partial_weight_exchange_client.py
```diff
@@ -109,9 +107,15 @@ def set_parameters(self, parameters: NDArrays, config: Config) -> None:
                 unwound properly by the parameter exchanger
             config (Config): configuration if required to control parameter exchange.
         """
-        if not self.model_weights_initialized:
+        # Since PartialWeightExchangeClient may exchange different sets parameters
+        # across different training round, the full parameter exchanger is only used
+        # if a client is selected to participate in the first *fitting* round.
+        current_server_round = self.narrow_config_type(config, "current_server_round", int)
+        fl_fit_round = self.narrow_config_type(config, "fl_fit_round", bool)
+        if (not self.model_weights_initialized) and current_server_round == 1 and fl_fit_round:
```
yc7z, line 115: In our current design, each client uses a full parameter exchanger to pull the initial model from the server at the very beginning of FL. However, in partial exchange regimes, clients can exchange a different set of parameters with the server, so if some client is not selected to participate in the first training round and it later tries to use the full parameter exchanger to pull the partial weights, there would be a problem. The change highlighted here fixes this bug.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+        all_scores = torch.cat([val.flatten() for _, val in all_parameter_scores.items()])
+        # Sorting all scores and determining the threshold.
+        sorted_scores, _ = torch.sort(all_scores, descending=True)
+        top_scores = sorted_scores[: math.ceil(len(sorted_scores) * self.sparsity_level)]
+        score_threshold = top_scores[-1].item()
+
+        names_to_masks = {}
+        for name, param_scores in all_parameter_scores.items():
+            # Use score_threshold to produce masks.
+            threshold_result = torch.where(param_scores >= score_threshold, input=param_scores, other=0)
```
yc7z, line 79: The mask is created by thresholding the parameter scores. So in theory, if all parameters have the same score (for example, when the model parameters are one constant) then the number of parameters selected won't necessarily match the sparsity level. But I think it is reasonable to assume that this won't happen in practice.

examples/sparse_tensor_partial_exchange_example/server.py
```diff
@@ -0,0 +1,116 @@
...
+    local_epochs: Optional[int] = None,
+    local_steps: Optional[int] = None,
+) -> Config:
+    config: Config = {
+        **make_dict_with_epochs_or_steps(local_epochs, local_steps),
+        "batch_size": batch_size,
+        "sparsity_level": sparsity_level,
+        "current_server_round": current_server_round,
+    }
+    config["fl_fit_round"] = False
```
yc7z, line 54: This config parameter is used to inform the client whether the current round is an eval or fit round so it can decide which parameter exchanger to use.

emersodb: With the changes we discussed on zoom to the get_parameters function in `basic_client`, leveraging def fit or def eval to send this boolean rather than the configs. We should be able to unify these config generator functions 🙂

fl4health/clients/partial_weight_exchange_client.py
```diff
@@ -109,9 +113,15 @@ def set_parameters(self, parameters: NDArrays, config: Config) -> None:
...
             self.initialize_all_model_weights(parameters, config)
```
emersodb, line 122: Just placing a note here with the summary of our zoom conversation. Essentially, we decided that even for the BasicClient, we will only do FullParameterExchange on the first **fit** round of federated learning. That is, the server will fully sync all FL client models who participate in the first round of training. Thereafter, any models that join federated training for evaluation or subsequent training will only receive the weights transferred by their "correct" exchanger (partial or not). Get parameters will receive the boolean as to whether we're fitting or not from the fit or eval functions that call it and we'll know what round # we're on using the config

fl4health/clients/partial_weight_exchange_client.py
jewelltaylor: > However, in partial exchange regimes, clients can exchange a different set of parameters with the server, so if some client is not selected to participate in the first fitting round and it later tries to use the full parameter exchanger to pull the partial weights, there would be a problem.
> 
> Thus, the client should only use the full parameter exchanger if it is in the first fitting round.

This implies we are sampling all clients in the first round right?

yc7z: If all clients are sampled in the first fitting round, the old code would work without any problem. But if let's say client 1 is not sampled in the first fitting round, and it is later sampled for the first time in some fitting round, then the server would have partial parameters while the client would try to use the full exchanger to pull the weights, which is a problem.

Also, because in this case the server does not necessarily have all the model parameters when a client tries to pull the weights from it for the first time, we can't follow what we did in for example `scaffold_client.py`.

fl4health/clients/partial_weight_exchange_client.py
jewelltaylor: > Essentially, we decided that even for the BasicClient, we will only do FullParameterExchange on the first fit round of federated learning. That is, the server will fully sync all FL client models who participate in the first round of training. 

Isn't this already the [default behaviour for BasicClient](https://github.com/VectorInstitute/FL4Health/blob/6472d0c477796edf1122943f6200e53f167ef6ab/fl4health/clients/basic_client.py#L151)?

> Thereafter, any models that join federated training for evaluation or subsequent training will only receive the weights transferred by their "correct" exchanger (partial or not).

If the correct exchanger is a FullParameterExchanger I assume we will still use it in subsequent rounds?

> Get parameters will receive the boolean as to whether we're fitting or not from the fit or eval functions that call it and we'll know what round # we're on using the config

All ParameterExchangers will follow this right? Have get_parameters take in the fitting bool? And Yuchong is going to update this to be true because its not currently the case?

emersodb: Let's talk about this a little more deeply together, but to answer your questions above
1) Sort of. We do use a FullParameterExchange whenever a client's weights are first initialized. This is problematic for partial weight exchange if a client does not participate in the first round of training.
2) Yes, we will always use the clients parameter exchange in subsequent rounds, as we do now. This includes when it's a FullParameterExchanger
3) We're only going to be changing the signature `get_parameters` function in the client. It won't change anything about how the ParameterExchangers function under the hood.

fl4health/clients/partial_weight_exchange_client.py
jewelltaylor: Ahh I see, thanks @yc7z! thats a much nicer property anyway (ie not having to have full participation in the first round)! I will dig into the code to understand the details more, but just initially while I wrap my head around things, I wanted to ask you: 

Why does the server model not have access to a complete set of weights to send to client who joins FL after the first round. On the first round the whole model is sent out. Subsequently, each client is going to share a subset of weights with server but I assume the server has to keep a copy of the entire model structure to map the partial weight updates to and aggregate across clients to redistribute a model the following round. Am I wrong in this assumption?

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
         assert 0 < exchange_percentage <= 1
         assert 0 < norm_threshold
-        self.threshold = norm_threshold
+        self.norm_threshold = norm_threshold
         self.exchange_percentage = exchange_percentage
         self.set_normalization_mode(normalize)
         self.set_filter_mode(filter_by_percentage)
+        self.set_select_drift_more(select_drift_more)
+
+    def set_layer_selection_function(
```
emersodb, line 118: Any reason that we have this set function rather than just doing `self.layer_selection_function = layer_selection_function` in the `__init__`? I know we use it below in the tests, but there we could simply do
```
exchanger.layer_selection_function = layer_selection_function
```
as well, since it's not a private attribute.

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
         model.load_state_dict(current_state, strict=True)
 
 
-class NormDriftParameterExchanger(ParameterExchangerWithPacking[List[str]]):
+class DynamicLayerExchanger(PartialParameterExchanger[List[str]]):
```
jewelltaylor, line 94: Not a huge deal, but curious why you changed the name of this class. It still seems be accurate to call it the NormDriftParameterExchanger. Is DynamicLayerExchanger potentially too general of a name (ie I would expect you can have a DynamicLayerExchanger that does not use a norm based selection criteria. Maybe I am wrong so let me know :)

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
     def set_filter_mode(self, filter_by_percentage: bool) -> None:
         self.filter_by_percentage = filter_by_percentage
 
     def set_normalization_mode(self, normalize: bool) -> None:
         self.normalize = normalize
 
+    def set_select_drift_more(self, select_drift_more: bool) -> None:
+        self.select_drift_more = select_drift_more
+
     def _calculate_drift_norm(self, t1: torch.Tensor, t2: torch.Tensor) -> float:
```
jewelltaylor, line 132: Is this method needed anymore? I don't see it being used anywhere

jewelltaylor: I also see there is a functional counterpart `parameter_selection_criteria.py`

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
     def __init__(
         self,
+        layer_selection_function: Callable[..., Tuple[NDArrays, List[str]]],
         norm_threshold: float,
         exchange_percentage: float = 0.1,
         normalize: bool = True,
         filter_by_percentage: bool = True,
+        select_drift_more: bool = True,
     ) -> None:
         """
```
jewelltaylor, line 104: Small documentation ask. It may be helpful to have some documentation on the arguments just clarify what these are in this class and downstream function that leverage them.

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
-            drift_norm = self._calculate_drift_norm(layer_param, layer_param_past)
-            names_to_norm_drift[layer_name] = drift_norm
-
-        total_param_num = len(names_to_norm_drift.keys())
-        num_param_exchange = int(math.ceil(total_param_num * self.exchange_percentage))
-        param_to_exchange_names = sorted(names_to_norm_drift.keys(), key=lambda x: names_to_norm_drift[x])[
-            :(num_param_exchange)
-        ]
-        return [model_states[name].cpu().numpy() for name in param_to_exchange_names], param_to_exchange_names
+    def select_parameters(
```
jewelltaylor, line 146: Since we have two different select parameters functionality (filter_by_percentage and filter_by_threshold) is it better to just have separate classes? Perhaps you can extend PartialWeightExchanger to add functionality that is common between these two drift based exchangers (ie push_parameters and pull_parameters) and then extend that class and implement the select_parameters method?

The only reason I suggest this is because its a bit awkward the fact you can potentially pass both a norm_threshold and an exchange_percentage even though you only use one and then you have to pass an extra bool (filter_by_percentage) to clarify. 

I could be completely missing something so feel free to fill me in :)

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
         assert 0 < norm_threshold
-        self.threshold = norm_threshold
+        self.norm_threshold = norm_threshold
         self.exchange_percentage = exchange_percentage
         self.set_normalization_mode(normalize)
         self.set_filter_mode(filter_by_percentage)
+        self.set_select_drift_more(select_drift_more)
+
+    def set_layer_selection_function(
+        self, layer_selection_function: Callable[..., Tuple[NDArrays, List[str]]]
```
emersodb, line 119: Correct me if I'm wrong, but below, this function actually has a fixed signature. That is, it takes 
`model, initial_model, self.exchange_percentage, self.normalize, self.select_drift_more`
So the type is `Callable[nn.Module, nn.Module, float, bool, bool, Tuple[NDArrays, List[str]]]`. It's a bit verbose, but you could create a type alias to simplify the annotation if you want.

fl4health/parameter_exchange/layer_exchanger.py
```diff
@@ -92,88 +91,76 @@ def pull_parameters(self, parameters: NDArrays, model: nn.Module, config: Option
...
-            names_to_norm_drift[layer_name] = drift_norm
-
-        total_param_num = len(names_to_norm_drift.keys())
-        num_param_exchange = int(math.ceil(total_param_num * self.exchange_percentage))
-        param_to_exchange_names = sorted(names_to_norm_drift.keys(), key=lambda x: names_to_norm_drift[x])[
-            :(num_param_exchange)
-        ]
-        return [model_states[name].cpu().numpy() for name in param_to_exchange_names], param_to_exchange_names
+    def select_parameters(
+        self, model: nn.Module, initial_model: Optional[nn.Module] = None
```
emersodb, line 147: For the signature here, I think you want to _force_. initial_model to have type `nn.Module` in the typing (ie. not optional) as the assertion below implies that it is required. In push_parameters, where this function is called, you already do an assertion, so the typing here can be stronger.

fl4health/parameter_exchange/layer_exchanger.py
emersodb: An alternative to this is that the selection function that is provided needs to be "pre-constructed" using partial with all but `model` and `initial_model`. That is, it only has a signature of `Callable[[nn.Module, nn.Module], Tuple[NDArrays, List[str]]]`. Then the `parameter_selection_criteria.py` file can have constructors based on the auxiliary inputs. We would then just have a `filter_by_percentage` and `filter_by_threshold` function constructor.

jewelltaylor: @emersodb Yes, I like that! We don't have to impose as strong of constraint on the type signature through this "pre-constructing". It worked out nicely in this case as you [asserted above](https://github.com/VectorInstitute/FL4Health/pull/103/files#r1476423922) but its nice to have more flexibility if needed

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
jewelltaylor: Is it worth checking if all elements are equal with torch.unique or something and logging a warning or raising an error if this is the case? Im thinking it would be helpful to surface if the provided scoring function produces the same value cause it might caused by an underlying bug

fl4health/parameter_exchange/parameter_packer.py
```diff
@@ -71,3 +73,47 @@ def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, List
         model_parameters = packed_parameters[:split_size]
         param_names = packed_parameters[split_size:][0].tolist()
         return model_parameters, param_names
+
+
+class SparseCooParameterPacker(ParameterPacker[Tuple[NDArrays, NDArrays, List[str]]]):
+    """
+    This parameter packer is responsible for selecting an arbitrary set of parameters
+    and then representing them in the sparse COO tensor format, which requires knowing
```
emersodb, line 81: If you have a good documentation link for COO sparse representations, maybe you can also include a link 🙂

fl4health/parameter_exchange/parameter_packer.py
```diff
@@ -71,3 +73,47 @@ def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, List
...
+    """
+
+    def pack_parameters(
+        self, model_parameters: NDArrays, additional_parameters: Tuple[NDArrays, NDArrays, List[str]]
+    ) -> NDArrays:
+        parameter_indices, tensor_shapes, tensor_names = additional_parameters
+        return model_parameters + parameter_indices + tensor_shapes + [np.array(tensor_names)]
+
+    def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, Tuple[NDArrays, NDArrays, List[str]]]:
+        assert len(packed_parameters) % 3 == 1
```
emersodb, line 93: A quick comment on why we expect a modulus of 1 here would be useful, I think

fl4health/parameter_exchange/parameter_packer.py
```diff
@@ -71,3 +73,47 @@ def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, List
...
+        split_size = (len(packed_parameters) - 1) // 3
+        model_parameters = packed_parameters[:split_size]
+        parameter_indices = packed_parameters[split_size : (2 * split_size)]
+        tensor_shapes = packed_parameters[(2 * split_size) : (3 * split_size)]
+        tensor_names = packed_parameters[(3 * split_size) :][0].tolist()
+        return model_parameters, (parameter_indices, tensor_shapes, tensor_names)
+
+    def extract_coo_info_from_dense(self, x: Tensor) -> Tuple[NDArray, NDArray, NDArray]:
```
emersodb, line 101: Pretty minor, but I think we can make this an `@staticmethod`, as it isn't specific to a particular instance.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+            model (nn.Module): Current model.
+            initial_model (nn.Module): Initial model.
+
+        Returns:
+            Dict[str, Tensor]: Dictionary that maps the name of each tensor to its corresponding mask.
+        """
+        all_parameter_scores = self.generate_parameter_scores(model, initial_model)
+        all_scores = torch.cat([val.flatten() for _, val in all_parameter_scores.items()])
+        # Sorting all scores and determining the threshold.
+        sorted_scores, _ = torch.sort(all_scores, descending=True)
```
jewelltaylor, line 72: Can we use torch.quantile to avoiding having to sort all the parameters in the model?

fl4health/parameter_exchange/parameter_packer.py
jewelltaylor: I was going to comment this, because it took me a minute to realize the the tensor names is wrapped in a list so its length 1. But I thought maybe I was just slow hah

fl4health/parameter_exchange/parameter_selection_criteria.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    Return those layers of model that deviate (in l2 norm) away from corresponding layers of
+    self.initial_model by at least (or at most) self.threshold.
+    """
+    layer_names = []
+    layers_to_transfer = []
+    initial_model_states = initial_model.state_dict()
+    model_states = model.state_dict()
+    for layer_name, layer_param in model_states.items():
+        layer_param_past = initial_model_states[layer_name]
```
emersodb, line 31: I move that we call this variable `ghost_of_layer_params_past`. @jewelltaylor, do you second this motion?

fl4health/parameter_exchange/layer_exchanger.py
yc7z: So I changed the name for two reasons. The first is that I want to emphasize that this exchanger only exchange entire layers, and the second is that now the exchanger can take in customized layer selection criterion function, which may not select layers based on their drift in l2 norms. I agree it's a little unclear what "DynamicLayerExchanger" refers to, but I couldn't think of a better name 😂  If you have other suggestions please let me know!

fl4health/parameter_exchange/layer_exchanger.py
yc7z: I included these setters to give the user the ability to reset the selection criterion etc during training, but it's possible the situation where this is needed doesn't arise in practice... Currently they are only used in the tests. We can remove them if you think that is better.

fl4health/parameter_exchange/layer_exchanger.py
yc7z: Yes currently these setters are not used. David also made a similar comment, see:

https://github.com/VectorInstitute/FL4Health/pull/103#discussion_r1476695267

fl4health/parameter_exchange/partial_parameter_exchanger.py
```diff
@@ -0,0 +1,35 @@
...
+
+import torch.nn as nn
+from flwr.common.typing import Config, NDArrays
+
+from fl4health.parameter_exchange.packing_exchanger import ParameterExchangerWithPacking
+
+T = TypeVar("T")
+
+
+class PartialParameterExchanger(ParameterExchangerWithPacking[T]):
```
emersodb, line 12: This is a bit of an odd inheritance because `ParameterExhangerWithPacking` inherits from `FullParameterExchanger`, which defines push and pull parameters. That makes setting them `abstractmethods` here feel a bit odd. We should probably have a `PartialParameterExchanger` that inherit from `ParameterExchanger` and ` Generic[T]` similar to the `ParameterExhangerWithPacking`. Then we need not define the push/pull parameters here and just tack on the select_parameters requirement.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+        assert 0 < sparsity_level <= 1
+        self.set_sparsity_level(sparsity_level)
+        self.parameter_packer: SparseCooParameterPacker = SparseCooParameterPacker()
+        self.set_score_gen_function(score_gen_function)
+
+    def set_sparsity_level(self, sparsity_level: float) -> None:
+        """This method exists to allow for setting new sparsity levels throughout training."""
+        self.sparsity_level = sparsity_level
+
+    def set_score_gen_function(self, score_gen_function: Callable[..., Dict[str, Tensor]]) -> None:
```
emersodb, line 42: I think this setter function can be removed and the `score_gen_function` property can be set in the `__init__`

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+from flwr.common.typing import Config, NDArrays
+from torch import Tensor
+from torch.nn.modules import Module
+
+from fl4health.parameter_exchange.parameter_packer import SparseCooParameterPacker
+from fl4health.parameter_exchange.partial_parameter_exchanger import PartialParameterExchanger
+
+
+class SparseCooParameterExchanger(PartialParameterExchanger[Tuple[NDArrays, NDArrays, List[str]]]):
+    def __init__(self, sparsity_level: float, score_gen_function: Callable[..., Dict[str, Tensor]]) -> None:
```
emersodb, line 17: Unless I'm mistaken, I think we can be more explicit in the Callable signature here, as all `score_gen_functions` should be `Callable[[nn.Module, nn.Module], Dict[str, Tensor]]`

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def generate_parameter_scores(self, model: nn.Module, initial_model: Optional[nn.Module]) -> Dict[str, Tensor]:
+        """Calling the score generating function to produce parameter scores."""
+        return self.score_gen_function(model, initial_model)
+
+    def create_masks(self, model: nn.Module, initial_model: Optional[nn.Module]) -> Dict[str, Tensor]:
+        """
+        Produce masks for model's tensors.
+
+        More precisely, this method first leverage a score generating function
```
emersodb, line 53: "first leverages" perhaps?

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
emersodb, line 79: I might be wrong, but I think `threshold_result` is the same tensor as 
```
sparse_tensor = torch.multiply(model_tensor, tensor_mask)
```
that you're computing below, which is what you really want anyway. Rather than forming the names_to_masks dictionary here and then multiplying by the mask to get the sparse_tensor, maybe we can just form a names_to_sparse_tensor_dictionary here? You can still not add them to the dictionary using `if not (threshold_result == 0).all():` 

Correct me if I'm wrong though.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -0,0 +1,154 @@
...
+    def pull_parameters(self, parameters: NDArrays, model: Module, config: Optional[Config] = None) -> None:
+        selected_parameters, additional_info = self.parameter_packer.unpack_parameters(parameters)
+        indices, shapes, names = additional_info
+        current_state = model.state_dict()
+
+        for param_values, param_indices, param_shape, param_name in zip(selected_parameters, indices, shapes, names):
+            # Use parameter values, indices, and shape to create
+            # a sparse coo tensor, which is then converted to a dense tensor
+            # to allow for loading.
+            param_coo = torch.sparse_coo_tensor(
```
emersodb, line 148: Cool!

fl4health/strategies/fedavg_sparse_coo_tensor.py
```diff
@@ -0,0 +1,302 @@
...
+            Dict[str, Tensor]: A dictionary mapping the name of the tensor that was aggregated to the aggregated
+                weights.
+        """
+        names_to_dense_tensors: DefaultDict[str, List[Tensor]] = defaultdict(list)
+        total_num_examples: DefaultDict[str, int] = defaultdict(int)
+
+        for packed_parameters, num_examples in results:
+            nonzero_parameter_values, additional_info = self.parameter_packer.unpack_parameters(packed_parameters)
+            parameter_indices, tensor_shapes, tensor_names = additional_info
+            for tensor_params, tensor_param_indices, tensor_shape, tensor_name in zip(
```
emersodb, line 230: Since zip has burned us in the past (myself very much included), can we just do a quick assert to make sure the sizes of all of the pieces being zipped together are equal and non-zero?

emersodb (commented):
A nice implementation of something quite tricky. Also great catch on the partial client participation bug catch. Most of my comments are fairly minor. There are a few possible refactors that I think will improve the PR in a few ways and it'll be ready to go

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
jewelltaylor: I thought maybe under the hood, quantile could do some binning to make it more efficient than sort but I did some profiling and the performance gain seems to be negligible so I think you can keep as is

fl4health/parameter_exchange/partial_parameter_exchanger.py
yc7z: Yes that's a good point. Does it also make sense then to not require this class to have `pack_parameters` or `unpack_parameters` methods? Otherwise it would be a bit strange to repeat those definitions here, although both of the partial exchangers we have need those methods...

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
yc7z: Yes, good point!

fl4health/parameter_exchange/layer_exchanger.py
yc7z: Just an update: I've removed all the setters for both `SparseCooParameterExchanger` and `DynamicLayerExchanger`

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
yc7z: `threshold_result` here is not exactly the result obtained after multiplying `model_tensor` by `tensor_mask`,  since  `torch.where()` is applied on the parameters scores tensor rather than `model_tensor` itself. But you are right that we can rewrite this part to avoid having to create a names_to_masks dictionary, so I'll reimplement it this way.

examples/partial_weight_exchange_example/client.py
```diff
@@ -40,7 +43,7 @@ def evaluate(self, parameters: NDArrays, config: Config) -> Tuple[float, int, Di
         if not self.initialized:
             self.setup_client(config)
 
-        self.set_parameters(parameters, config)
+        self.set_parameters(parameters, config, fitting_round=False)
         testing = self.narrow_config_type(config, "testing", bool)
         num_classes = num_classes = self.narrow_config_type(config, "num_classes", int)
```
fatemetkl, line 48: I think we have an extra num_classes here.

yc7z: Yes thanks for spotting it!

fl4health/clients/basic_client.py
```diff
@@ -130,19 +130,27 @@ def get_parameters(self, config: Config) -> NDArrays:
...
             config (Config): The config is sent by the FL server to allow for customization in the function if desired.
+            fitting_round (bool): Boolean that indicates whether the current federated learning round is a
+                fitting round or an evaluation round.
+                This is used to help determine which parameter exchange should be used for pulling parameters.
+                A full parameter exchanger is only used if the current federated learning round is the very
+                first fitting round.
         """
         assert self.model is not None
-        if not self.model_weights_initialized:
+        current_server_round = self.narrow_config_type(config, "current_server_round", int)
```
emersodb, line 152: Since we're no longer using `self.model_weights_initialized`, I think you can delete the property from `BasicClient`?

yc7z: Yes good catch. Sorry I thought I deleted it already but apparently I forgot.

emersodb: There is one place that this was actually being used `examples/warm_up_example/warmed_up_fenda/client.py` where we need to address this. I think there we can edit `get_model` and `initialize_all_model_weights` as follows to get the same effect:
```
    return FendaModel(LocalCnn(), GlobalCnn(), FendaClassifier(FendaJoinMode.CONCATENATE)).to(self.device)
```
and
```
# Overriding the initialize_all_model_weights function is required to load the pretrained model
def initialize_all_model_weights(self, parameters: NDArrays, config: Config) -> None:
    # Load the pretrained model
    self.warmed_up_module.load_from_pretrained(self.model)
```

fl4health/parameter_exchange/parameter_selection_criteria.py
```diff
@@ -1,13 +1,58 @@
...
+            select_layers_by_percentage,
+            self.exchange_percentage,
+            self.normalize,
+            self.select_drift_more,
+        )
+
+
+# Selection criteria functions for selecting entire layers. Intended to be used
+# by the DynamicLayerExchanger class via the LayerSelectionFunctionConstructor class.
 def _calculate_drift_norm(t1: torch.Tensor, t2: torch.Tensor, normalize: bool) -> float:
```
emersodb, line 56: I might be wrong, but I think we can drop the underscore prefix here? I know the underscore is used to indicate "private" class functions. I'm not sure if a similar convention holds for stand-alone functions?

fl4health/parameter_exchange/parameter_selection_criteria.py
```diff
@@ -28,8 +77,8 @@ def select_layers_by_threshold(
     initial_model_states = initial_model.state_dict()
     model_states = model.state_dict()
     for layer_name, layer_param in model_states.items():
-        layer_param_past = initial_model_states[layer_name]
-        drift_norm = _calculate_drift_norm(layer_param, layer_param_past, normalize)
+        ghost_of_layer_params_past = initial_model_states[layer_name]
```
emersodb, line 80: LOL. I really hope someone finds this in the source code some day and laughs a little bit.

fl4health/parameter_exchange/partial_parameter_exchanger.py
emersodb: I think the way you implemented this looks good to me.

fl4health/clients/basic_client.py
yc7z: I think this fix makes sense too. I've applied it in the latest commit.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -12,117 +12,114 @@
...
-        model_states = model.state_dict()
+        all_parameter_scores = self.generate_parameter_scores(model, initial_model)
+        all_scores = torch.cat([val.flatten() for _, val in all_parameter_scores.items()])
+        # Sorting all scores and determining the threshold.
+        sorted_scores, _ = torch.sort(all_scores, descending=True)
+        n_top_scores = math.ceil(len(sorted_scores) * self.sparsity_level)
+        # Sanity check.
+        assert n_top_scores >= 1
+        top_scores = sorted_scores[:n_top_scores]
+        score_threshold = top_scores[-1].item()
```
emersodb, line 87: Perhaps, I'm wrong, but could we just do 
```
score_threshold = sorted_scores[n_top_scores-1]
```
rather than slicing sorted_scores and taking the last element?

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
```diff
@@ -12,117 +12,114 @@
...
-            selected_parameters, selected_indices, tensor_shape = self.parameter_packer.extract_coo_info_from_dense(
-                sparse_tensor
-            )
+            # Sanity check.
+            assert model_tensor.shape == param_scores.shape
 
-            selected_parameters_all_tensors.append(selected_parameters)
-            selected_indices_all_tensors.append(selected_indices)
-            tensor_shapes.append(tensor_shape)
+            unique_score_values = torch.unique(
```
emersodb, line 100: I'd suggest pulling this check/warning into a helper function so as not to clutter the calculations here

emersodb (approved):
This looks pretty much ready to go from my perspective. I left a few small comments and small refactors, but otherwise I think it's ready.

fl4health/parameter_exchange/parameter_selection_criteria.py
yc7z: I think the same convention holds for stand-alone functions as well. For example, if you do a star import on the module, python will ignore the functions whose names begin with an underscore.

fl4health/parameter_exchange/sparse_coo_parameter_exchanger.py
yc7z: Yes absolutely

fl4health/parameter_exchange/parameter_selection_criteria.py
emersodb: I didn't know that! Thanks for the info.

fatemetkl (commented):
I had just some minor comments, everything else looks good!

fl4health/clients/partial_weight_exchange_client.py
```diff
@@ -93,25 +106,29 @@ def get_parameters(self, config: Config) -> NDArrays:
...
+                first fitting round. Otherwise, use a PartialParameterExchanger.
         """
-        if not self.model_weights_initialized:
-            self.initialize_all_model_weights(parameters, config)
-        else:
-            self.parameter_exchanger.pull_parameters(parameters, self.model, config)
-        # stores the values of the new model parameters at the beginning of each client training round.
-        self.initial_model.load_state_dict(self.model.state_dict(), strict=True)
+        super().set_parameters(parameters, config, fitting_round)
+        if self.initial_model:
```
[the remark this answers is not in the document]
yc7z: Yes I agree it makes more sense to use `self.store_initial_model` here.

PR #108, opened 2024-02-08
Fixing a small, but significant, bug with MOON.

emersodb:
# PR Type
Fix

# Short Description

Fixing a small bug in the MOON implementation in line with Sana's fix. Adding some tests to ensure the fix is properly functioning.

# Tests Added

Some tests to make sure the loss calculations etc. function as expected given the change.

### DIFF
--- a/fl4health/clients/moon_client.py
+++ b/fl4health/clients/moon_client.py
@@ -60,12 +60,14 @@ class MoonClient(BasicClient):
             the old model are returned. All predictions included in dictionary will be used to compute metrics.
         """
         preds, features = self.model(input)
-        old_features = torch.zeros(self.len_old_models_buffer, *features["features"].size()).to(self.device)
-        for i, old_model in enumerate(self.old_models_list):
-            _, old_model_features = old_model(input)
-            old_features[i] = old_model_features["features"]
-        _, global_model_features = self.global_model(input)
-        features.update({"global_features": global_model_features["features"], "old_features": old_features})
+        # If there are no models in the old_models_list, we don't compute the features for the contrastive loss
+        if len(self.old_models_list) > 0:
+            old_features = torch.zeros(len(self.old_models_list), *features["features"].size()).to(self.device)
+            for i, old_model in enumerate(self.old_models_list):
+                _, old_model_features = old_model(input)
+                old_features[i] = old_model_features["features"]
+            _, global_model_features = self.global_model(input)
+            features.update({"global_features": global_model_features["features"], "old_features": old_features})
         return preds, features
 
     def get_contrastive_loss(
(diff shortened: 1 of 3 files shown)


fl4health/clients/moon_client.py
```diff
@@ -89,14 +91,17 @@ def get_contrastive_loss(
 
         return self.ce_criterion(logits, labels)
 
-    def set_parameters(self, parameters: NDArrays, config: Config) -> None:
+    def get_parameters(self, config: Config) -> NDArrays:
         assert isinstance(self.model, MoonModel)
-        # Save the parameters of the old local model
+        # Save the parameters of the old LOCAL model
```
lotif, line 96: Why is the saving being done on get now instead of set?

lotif (commented):
Just one question

fl4health/clients/moon_client.py
emersodb: That's a good question and the crux of the issue that Sana identified. `set_parameters` is called to start both the `fit` and `eval` stages. The eval stage happens after aggregation. So in the simplified case, when `old_models_list` is supposed to have size 1, the `old_local_model` that MOON needs to compute features is replaced with the global model produced by aggregation. `get_parameters` is only called at the end of each `fit`, which means that old_models_list will always have the most recent model produced by **local** training, which is what we want.

lotif: Just for the record, I think this would be better if moved to its own function, but as per offline conversation, this will be done in a follow up ticket (https://app.clickup.com/t/86879vn75)

lotif (approved):

tests/clients/test_moon_client.py
```diff
@@ -20,12 +20,64 @@ def test_setting_parameters(get_client: MoonClient) -> None:  # noqa
...
 
+    # Global model parameters should match those of the new_params
+    global_model_params = [val.cpu().numpy() for _, val in moon_client.global_model.state_dict().items()]
+    # Make sure the MOON model parameters are equal to the global model parameters
+    new_moon_model_params = [val.cpu().numpy() for _, val in moon_client.model.state_dict().items()]
+
+    for i in range(len(params)):
+        assert (new_params[i] == global_model_params[i]).all()
+        assert (new_params[i] == new_moon_model_params[i]).all()
+
```
sanaAyrml, line 31: Maybe we can add an assertion here to check whether global_model and old_models_list[-1] is different. (That's what I figured out the bug from)

sanaAyrml:
Changes seems good to me, Thanks for applying changes.

sanaAyrml (approved):
Changes seems good to me, Thanks for applying changes.

tests/clients/test_moon_client.py
emersodb: So I can't do it in this part, because I assume that `len(moon_client.old_models_list) == 0` just below. However, below, on both lines 59 and 78, there is an assertion that they are different. I'll add in setting the model twice in the last check, just to be certain.

fatemetkl (approved):

PR #112, opened 2024-03-02
Ditto FLamby implementation scripts.

emersodb:
# PR Type
Other

# Short Description

This PR constructs the client/server code and the hyper-parameter sweep scripts needed to run the Ditto experiments on FLamby. I haven't tested this yet, but they should be pretty close to ready to go, as our framework for FLamby experiments is fairly robust.

**Note**: When evaluating the models using the `evaluate_on_holdout.py` script, there may be some small tweaks that need to be made, given that this model produces two predictions (global and local). We **only care about the local predictions in evaluation**. However, hopefully they are minor if necessary at all.

# Tests Added

None


research/flamby/fed_isic2019/ditto/client.py
```diff
@@ -0,0 +1,150 @@
...
+    def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
+        train_dataset, validation_dataset = construct_fedisic_train_val_datasets(
+            self.client_number, str(self.data_path)
+        )
+        train_loader = DataLoader(train_dataset, batch_size=BATCH_SIZE, shuffle=True)
+        val_loader = DataLoader(validation_dataset, batch_size=BATCH_SIZE, shuffle=False)
+        return train_loader, val_loader
+
+    def get_model(self, config: Config) -> nn.Module:
+        model: nn.Module = Baseline().to(self.device)
```
fatemetkl, line 60: I might be totally wrong, but are we training the same number of parameters in Ditto as in other methods (similar to the case in APFL or FENDA)? Given that in Ditto, we are training both a global and local model in each client, should the model size be approximately half of the baseline model? 
If that's true, we can probably achieve this by using the APFL model structure.

emersodb: I think this is a good discussion point and thought. I considered this when writing the FedISIC scripts and would be interested in hearing both your thoughts. The model being trained isn't "twice as large" in that it has the same number of parameters as a centralized method (similar to FedPer). So it doesn't have additional expressiveness w.r.t the number of DoFs. However, we are technically training two models of the same size (global and local, but only one is used in prediction, as opposed to APFL). So the **cost of training** is higher for Ditto but the model isn't bigger. We should definitely method at least mention this in the paper discussion either way.

fatemetkl: Right, I think this makes sense, since only one model is used for prediction, and the other is trained to facilitate the main one (local model) in the training process. So, I will also use the full model for GEMINI experiments if John has the same opinion.

jewelltaylor: I think your raise a good point Fatemeh! But I agree with David that we shouldn't consider the global models parameter counts since at inference only the local model is used. Maybe SCAFFOLD is a somewhat useful analogy (although not perfect obviously), we don't consider control variates in the parameter count. That said, definitely worth mentioning the additional cost of training in the paper

fatemetkl: Yeah SCAFFOLD is a good example! 👍 I will also use the full model then.

jewelltaylor (approved):
Went through the code and everything looks good to me! Was also able to run the fedheart disease experiments with no issues so I think this is good to go. If I run into any issues with other datasets I will make a new PR to update experiment scripts

PR #114, opened 2024-03-11
Small fix for when sampling dataloaders produce an empty batch.

emersodb:
# PR Type
Fix

# Short Description
Fixing a few small typos and addressing a small bug that can occur when the data loaders are statistical samplers.

For example, if you're using a small batch size and a small Poisson probability some batches may be empty strictly due to sampling. Consider a dataset size of 100 and a Poisson probability of 3%. Most of the time you'll get batches of around three, but you can theoretically get a batch size of 0.


fl4health/clients/basic_client.py
```diff
@@ -456,6 +456,11 @@ def train_by_epochs(
             self.train_loss_meter.clear()
             for input, target in self.train_loader:
                 input, target = input.to(self.device), target.to(self.device)
+                # Assume first dimension is batch size. Sampling iterators (such as Poisson batch sampling), can
+                # construct empty batches. We skip the iteration if this occurs.
+                if len(input) == 0:
```
yc7z, line 461: Would it be better if we check for this before moving `input` and `target` to the GPU?

emersodb: Yes it would!

yc7z (approved):
Looks good to me!

PR #116, opened 2024-03-12
Mkmmd loss

sanaAyrml:
# PR Type
[Feature ]

# Short Description

This is an updated version of MK-MMD loss implementation with additions to Fenda, MOON and Ditto models. 

- Fenda: Two mk-mmd losses are implemented, one to maximize the distance between local features and the aggregated model and the other to minimize the distance between global features and the aggregated model. In this setting, betas are only updated at the beginning of each step.

- MOON: Two mk-mmd loss is implemented one to minimize distance between local features and aggregated model and other to maximize distance between local features and old-local features model. In this setting betas are updated even during training after a certain number of step based on `beta_update_interval`. Also in order to prevent from feature value boost a l2-norm regularizer has been added which can be controlled via `feature_l2_norm` parameter.

- Ditto: One mk-mmd loss is implemented one to minimize distance between local features and aggregated global model. In this setting betas are updated even during training after a certain number of step based on `beta_update_interval`. Also in order to prevent from feature value boost a l2-norm regularizer has been added which can be controlled via `feature_l2_norm` parameter.

Also I reorganized the implementation of contrastive loss as a separate file in `fl4health/losses` folder. And generalized loss reporting function in `fl4health/utils/losses` as it was failing to report correctly due to addition of new keys during further steps in MOON setup.

Flamby MK-MMD loss experimental setup has also been added to the `research/flamby/fedisic2019` folder.

# Test

MK-MMD and contrastive loss test added. Also fixed some client's tests to ensure the state of the model during validation and training loss computation.


fl4health/clients/basic_client.py
```diff
@@ -832,7 +832,8 @@ def update_after_step(self, step: int) -> None:
         """
         Hook method called after local train step on client. step is an integer that represents
         the local training step that was most recently completed. For example, used by the APFL
-        method to update the alpha value after a training a step.
+        method to update the alpha value after a training a step. Also used by the Moon, Fenda
```
emersodb, line 835: Since MOON, FENDA, MK-MMD are acronyms, we can capitalize them in this comment. Similarly Ditto can be capitalized as a proper name.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        Returns:
+            torch.Tensor: Returns the L2 inner product between the initial global weights of the round and the current
+                local model weights.
+        """
+        assert self.init_global_model is not None and self.model is not None and self.lam is not None
+        # Using parameters to ensure the same ordering as exchange
+        local_model_weights = [layer_weights for layer_weights in self.model.parameters()]
+        # Detach the weights to ensure we don't compute gradients with respect to the tensors
+        initial_global_weights = [layer_weights.detach() for layer_weights in self.init_global_model.parameters()]
```
emersodb, line 118: Each time we form the loss (i.e. every batch), we're going to be recomputing this quantity, even though it never changes during a client-side training run. Perhaps we can do this one right after cloning and freezing the global model in `update_before_train`?

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+            config (Config): The config is sent by the FL server to allow for customization in the function if desired.
+        """
+        self.parameter_exchanger.pull_parameters(parameters, self.model, config)
+        self.parameter_exchanger.pull_parameters(parameters, self.global_model, config)
+
+    def update_before_train(self, current_server_round: int) -> None:
+        assert isinstance(self.model, nn.Module)
+        # Clone and freeze the initial weights GLOBAL MODEL. These are used to form the Ditto local
+        # update penalty term.
+        self.init_global_model = self.clone_and_freeze_model(self.global_model)
```
emersodb, line 190: I like doing this here, but perhaps instead of cloning the model here, we can just clone and detach the weights to be stored as in the original Ditto implementation like this 
```
# Saving the initial weights GLOBAL MODEL weights and detaching them so that we don't compute gradients with
# respect to the tensors. These are used to form the Ditto local update penalty term.
self.initial_global_tensors = [
    initial_layer_weights.detach().clone() for initial_layer_weights in self.global_model.parameters()
]
```

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+from fl4health.checkpointing.checkpointer import TorchCheckpointer
+from fl4health.clients.basic_client import BasicClient
+from fl4health.losses.mkmmd_loss import MkMmdLoss
+from fl4health.model_bases.moon_base import MoonModel
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.losses import EvaluationLosses, LossMeterType, TrainingLosses
+from fl4health.utils.metrics import Metric
+
+
+class DittoClient(BasicClient):
```
emersodb, line 20: I like the idea of having a separate file/class for this implementation and I think we should do something sort of similar for the FENDA and MOON implementations. That way we keep the original implementations simpler and have a more complex version. 

However, I have two suggestions here. The first is to name the class something different (maybe just `DittoMkMmdClient`), so that there isn't confusion when instantiating a class. Second, I think we should inherit from `DittoClient` here (rather than just `BasicClient`), as there are a lot of methods that can be reused and the remainder can be override.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        device: torch.device,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        lam: float = 1.0,
+        mkmmd_loss_weight: float = 10.0,
+        beta_update_interval: int = 20,
+        feature_l2_norm: Optional[float] = 0.0,
+    ) -> None:
+        """
+        This client implements the Ditto algorithm from Ditto: Fair and Robust Federated Learning Through
```
emersodb, line 34: The current comment here doesn't discuss the MkMMD part.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        self.mkmmd_loss = MkMmdLoss(device=self.device, minimize_type_two_error=True).to(self.device)
+
+        self.global_model: nn.Module
+        self.init_global_model: nn.Module
+
+    def get_optimizer(self, config: Config) -> Dict[str, Optimizer]:
+        """
+        Returns a dictionary with global and local optimizers with string keys 'global' and 'local' respectively.
+        """
+        raise NotImplementedError
```
emersodb, line 77: I know I didn't do this when I implemented Ditto originally, but I think it would be helpful for us to add a string to the reporting of this error along the lines of
```
NotImplementedError("User Clients must define a function that returns a Dict[str, Optimizer] with keys \"global\" and \"local\" defining separate optimizers for the global and local models of Ditto.")
```

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        Args:
+            parameters (NDArrays): Model parameters to be injected into the client model
+            config (Config): The config is sent by the FL server to allow for customization in the function if desired.
+        """
+        self.parameter_exchanger.pull_parameters(parameters, self.model, config)
+        self.parameter_exchanger.pull_parameters(parameters, self.global_model, config)
+
+    def update_before_train(self, current_server_round: int) -> None:
+        assert isinstance(self.model, nn.Module)
```
emersodb, line 187: I think we want to assert that `self.global_model `is an `nn.Module` here?

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,288 @@
...
+                to compute the Mk-MMD measure. If not provided, a unit-length, random default is constructed. These
+                can be optimized using the functions of this class. Defaults to None.
+            minimize_type_two_error (Optional[bool], optional): Whether we're aiming to minimize the type II error in
+                optimizing the betas or maximize it. The first coincides with trying to minimize feature distance. The
+                second coincides with trying to maximize their feature distance. Defaults to True.
+        """
+        super().__init__()
+        self.device = device
+        if gammas is None:
+            # Note arange is not inclusive, so this ends up being [-3.5, 1] in steps of 0.5
```
emersodb, line 40: I think we want to update this comment to match the steps to 0.25?

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        # Take a step with the local model using the local loss and Ditto constraint
+        losses.backward["backward"].backward()
+        self.optimizers["local"].step()
+
+        # Return dictionary of predictions where key is used to name respective MetricMeters
+        return losses, preds
+
+    def update_after_step(self, step: int) -> None:
+        if step % self.beta_update_interval == 0:
```
emersodb, line 267: Because our step counters start at 0, we're going to update the betas immediately after the first step. I think that's what we want, but just wanted to highlighted it, in case I'm wrong.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+            if self.mkmmd_loss_weight and self.init_global_model:
```
emersodb, line 268: Based on our implementation, these should both not be None. If they are, something has gone wrong and we can throw an exception. So rather than wrapping this up in an if statement. I'd say we just do an 
```
assert self.mkmmd_loss_weight is not None
assert self.init_global_model is not None
```

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+                # Get the feature distribution of the local and init global features with evaluation mode
+                local_distribution, init_global_distribution = self.update_buffers(self.model, self.init_global_model)
```
emersodb, line 270: Maybe I'm missing something down the line, but I doing think we need to compute these distributions if the mkmmd weight is 0? If that's right, then we can tuck these under the if statement below as well to avoid computation when we don't want it to happen.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        """
+        super().__init__(
+            data_path=data_path,
+            metrics=metrics,
+            device=device,
+            loss_meter_type=loss_meter_type,
+            checkpointer=checkpointer,
+        )
+        self.lam = lam
+        self.mkmmd_loss_weight = mkmmd_loss_weight
```
emersodb, line 65: Perhaps we can include a small warning here if the mkmmd weight is 0 that the loss function won't be used and the individual should consider using the vanilla `DittoClient` instead?

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        return super().update_after_step(step)
+
+    def update_buffers(
+        self, local_model: torch.nn.Module, init_global_model: torch.nn.Module
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Update the feature buffer of the local and global features."""
+        assert isinstance(local_model, MoonModel)
+        assert isinstance(init_global_model, MoonModel)
+
+        init_state_local_model = local_model.training
```
emersodb, line 287: A comment here would be useful to understand what we're doing here. Based on the code below, we're using this to store the state that the local model is currently in, so we can set if back if need be.

fl4health/clients/ditto_mkmmd_client.py
emersodb: Below, I see why we clone the initial global model here. So we should definitely leave that, but I think it's still a good idea to save the initial tensors globally, rather than forming them at every step since they are constant.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        # Set local model to evaluation mode
+        local_model.eval()
+
+        assert not local_model.training
+        assert not init_global_model.training
```
emersodb, line 293: A comment here as to why we can assert this would be helpful (i.e. it's cloned and frozen upon creation and should never be in training mode)

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+                input, target = input.to(self.device), target.to(self.device)
+                _, local_features = local_model(input)
+                _, init_global_features = init_global_model(input)
+
+                local_buffer.append(local_features["features"].reshape(len(local_features["features"]), -1))
+                init_global_buffer.append(
+                    init_global_features["features"].reshape(len(init_global_features["features"]), -1)
+                )
+
+        if init_state_local_model:
```
emersodb, line 309: Comment that we're putting it back to its original train/eval mode

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        local_model.eval()
+
+        assert not local_model.training
+        assert not init_global_model.training
+
+        local_buffer = []
+        init_global_buffer = []
+
+        with torch.no_grad():
+            for input, target in self.train_loader:
```
emersodb, line 299: Super minor, but since we're not using target, we can just not store it `for input, _ in self.train_loader:`
Then we don't have to move it to GPU either.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+                input, target = input.to(self.device), target.to(self.device)
+                _, local_features = local_model(input)
+                _, init_global_features = init_global_model(input)
+
+                local_buffer.append(local_features["features"].reshape(len(local_features["features"]), -1))
```
emersodb, line 304: A comment here that we're flattening the features would be useful I think

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+                init_global_buffer.append(
+                    init_global_features["features"].reshape(len(init_global_features["features"]), -1)
+                )
+
+        if init_state_local_model:
+            local_model.train()
+
+        return torch.cat(local_buffer, dim=0), torch.cat(init_global_buffer, dim=0)
```
emersodb, line 312: A comment that the assumption is that the tensors are "batch-first" so our concatenation is in `dim=0` would be helpful I think.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        Returns:
+            Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
+            contains predictions indexed by name and the second element contains intermediate activations
+            index by name.
+
+        Raises:
+            ValueError: Occurs when something other than a tensor or dict of tensors is returned by the model
+            forward.
+        """
+        if isinstance(self.model, MoonModel):
```
emersodb, line 330: A comment here explaining why we're tossing the global model features, would be helpful.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        """
+        Computes the loss and any additional losses given predictions of the model and ground truth data.
+
+        Args:
+            preds (Dict[str, torch.Tensor]): Prediction(s) of the model(s) indexed by name.
+            features (Dict[str, torch.Tensor]): Feature(s) of the model(s) indexed by name.
+            target (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            Tuple[torch.Tensor, Union[Dict[str, torch.Tensor], None]]; A tuple with:
```
emersodb, line 373: This specifies the type for the Dict as being `Union[Dict[str, torch.Tensor], None]` but in the method signature, None isn't an option.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+        additional_losses = {"local_loss": local_loss, "global_loss": global_loss}
+
+        if self.mkmmd_loss_weight != 0:
+            assert "init_global_features" in features
+            assert "features" in features
+            # Compute MK-MMD loss
+            mkmmd_loss = self.mkmmd_loss(features["features"], features["init_global_features"])
+            total_loss += self.mkmmd_loss_weight * mkmmd_loss
+            additional_losses["mkmmd_loss"] = mkmmd_loss
+            if self.feature_l2_norm:
```
emersodb, line 398: I think technically, this component is "independent" of the mkmmd loss weight That is, we don't want to nest this if statement. Rather, even if mkmmd_loss_weight is zero, we could theoretically have an l2 norm loss.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+                feature_l2_norm_loss = torch.linalg.norm(features["features"])
```
emersodb, line 399: Maybe this is a silly idea, but perhaps we should scale this by the batch size? That way it doesn't scale with the size of the batch being considered?

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        Returns:
+            TrainingLosses: an instance of TrainingLosses containing backward loss and
+                additional losses indexed by name. Additional losses includes each loss component and the global model
+                loss tensor.
+        """
+        # Check that both models are in training mode
+        assert self.global_model.training and self.model.training
+
+        total_loss, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
```
emersodb, line 433: If we refactor the Ditto class to use this method and override `self.compute_loss_and_additional_losses` to compute the global and local losses there, we can just drop this from this client after inheriting from Ditto to save code duplication.

fl4health/clients/ditto_mkmmd_client.py
```diff
@@ -0,0 +1,483 @@
...
+            target: (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            EvaluationLosses: an instance of EvaluationLosses containing checkpoint loss and additional losses
+                indexed by name.
+        """
+        # Check that both models are in eval mode
+        assert not self.global_model.training and not self.model.training
+
+        _, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
```
emersodb, line 479: Similar to the train component, we can inherit from Ditto and make this properly inherited.

fl4health/clients/fenda_client.py
```diff
@@ -46,13 +49,17 @@ def __init__(
             Each value associate with one of two contrastive losses in perfcl loss.
             cos_sim_loss_weight: Weight to be used for cosine similarity loss.
             contrastive_loss_weight: Weight to be used for contrastive loss.
+            mkmmd_loss_weights: Weights to be used for mkmmd losses. Fist value is for minimizing the distance
```
emersodb, line 52: Fist -> First 🙂

fl4health/clients/fenda_client.py
```diff
@@ -6,6 +6,8 @@
```
emersodb, line 6: I've put a ticket into our backlog for this, but after we get these calculations ready and merged, we should dedicate a bit of time to trying to simplify this implementation a bit, perhaps by breaking it into pieces etc. As it stands now, the FENDA code has become pretty complex and overloaded. Not necessary here, but in a follow up, I'll try to simplify a few things.

fl4health/clients/fenda_client.py
```diff
@@ -110,8 +124,67 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        """Update the feature buffer of the local, global and aggregated features."""
+        assert isinstance(local_module, torch.nn.Module)
+        assert isinstance(global_module, torch.nn.Module)
+        assert isinstance(aggregated_module, torch.nn.Module)
+
+        local_buffer = []
+        global_buffer = []
+        aggregated_buffer = []
+
+        # Save the state of the local and global modules
```
emersodb, line 156: Maybe include a word on why we need not save the `aggregated_modules` training state?

fl4health/clients/fenda_client.py
```diff
@@ -110,8 +124,67 @@ def update_before_train(self, current_server_round: int) -> None:
...
+
+        with torch.no_grad():
+            for input, target in self.train_loader:
+                input, target = input.to(self.device), target.to(self.device)
+                local_features = local_module.forward(input)
+                global_features = global_module.forward(input)
+                aggregated_features = aggregated_module.forward(input)
+
+                # Local feature are same as old local features
+                local_buffer.append(local_features.reshape(len(local_features), -1))
```
emersodb, line 176: To make the code a bit more readable/re-usable, maybe we can consolidate this flattening into a private function like `def _flatten_features` and add a comment that we're assuming batch first.

fl4health/clients/fenda_client.py
```diff
@@ -172,18 +226,38 @@ def get_perfcl_loss(
...
+        return global_contrastive_loss, local_contrastive_loss
+
+    def get_mkmmd_loss(
+        self,
+        local_features: torch.Tensor,
+        global_features: torch.Tensor,
+        aggregated_global_features: torch.Tensor,
+    ) -> Tuple[Optional[torch.Tensor], Optional[torch.Tensor]]:
+        """
+        MK-MMD loss aims to compute the distance among given two distributions with optimized betas.
```
emersodb, line 249: This comment could be slightly more informative (i.e. the betas are optimized to maximize the statistical power of the measurement.) We're also missing the doc-strings for the args an return values.

fl4health/clients/moon_client.py
```diff
@@ -23,10 +27,13 @@ def __init__(
...
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
-        checkpointer: Optional[TorchCheckpointer] = None,
         temperature: float = 0.5,
-        contrastive_weight: Optional[float] = None,
         len_old_models_buffer: int = 1,
+        beta_update_interval: int = 20,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        contrastive_weight: Optional[float] = None,
+        mkmmd_loss_weights: Optional[Tuple[float, float]] = None,
+        feature_l2_norm: Optional[float] = None,
```
emersodb, line 36: Very minor, but I would suggest renaming this to include `_weight` at the end to be a bit clearer that this isn't actually a "norm"

fl4health/clients/moon_client.py
```diff
@@ -1,10 +1,14 @@
+from logging import INFO
```
emersodb, line 1: For Moon, perhaps we can do something similar to Ditto, where instead of changing the behavior of the default Moon implementation, we can inherit from Moon and strap on the additional MkMMD behavior?

fl4health/clients/moon_client.py
```diff
@@ -108,8 +99,77 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        # Optimizing betas for the MK-MMD loss has not been done yet
+        self.betas_optimized = False
+
         return super().update_before_train(current_server_round)
 
+    def update_after_step(self, step: int) -> None:
+        if step > 0 and step % self.beta_update_interval == 0:
+            if self.mkmmd_loss_weights and len(self.old_models_list) > 0 and self.global_model:
+                # Get the feature distribution of the old, local and global models with evaluation mode
+                local_distribution, global_distribution, old_distribution = self.update_buffers(
```
emersodb, line 111: Maybe worth mentioning in the comment that we only care about the most recent old model for mkmmd?

fl4health/clients/moon_client.py
```diff
@@ -108,8 +99,77 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        assert not old_local_model.training
+
+        with torch.no_grad():
+            for input, target in self.train_loader:
+                input, target = input.to(self.device), target.to(self.device)
+                _, local_features = local_model(input)
+                _, global_features = global_model(input)
+                _, old_local_features = old_local_model(input)
+
+                local_buffer.append(local_features["features"].reshape(len(local_features["features"]), -1))
```
emersodb, line 163: Similar to FENDA, it might be nice to have a helper function for flattening. Maybe we should pull this out into a util function or something to be share throughout?

fl4health/clients/moon_client.py
emersodb: Alternatively, since we're replacing the contrastive loss with the MkMMD loss here, perhaps we just create a separate MOON class for MkMMD.

fl4health/clients/moon_client.py
```diff
@@ -137,13 +197,34 @@ def compute_loss_and_additional_losses(
...
-            contrastive_loss = self.get_contrastive_loss(
-                features["features"], features["global_features"], features["old_features"]
+        if self.contrastive_weight and "old_features" in features:
+            contrastive_loss = self.contrastive_loss(
+                features["features"], features["global_features"].unsqueeze(0), features["old_features"]
             )
             total_loss += self.contrastive_weight * contrastive_loss
             additional_losses["contrastive_loss"] = contrastive_loss
 
+        elif (
```
emersodb, line 207: Perhaps I'm wrong, but I think this can just be a `if`. That is, both losses can technically be applied if the weights are non-zero. Would be overkill, but not necessarily inadmissable?

fl4health/clients/moon_client.py
```diff
@@ -137,13 +197,34 @@ def compute_loss_and_additional_losses(
...
+                total_loss += self.mkmmd_loss_weights[0] * mkmmd_loss_min
+                additional_losses["mkmmd_loss_min"] = mkmmd_loss_min
+
+            if self.mkmmd_loss_weights[1] != 0:
+                mkmmd_loss_max = self.mkmmd_loss_max(features["features"], features["old_features"][-1])
+                total_loss -= self.mkmmd_loss_weights[1] * mkmmd_loss_max
+                additional_losses["mkmmd_loss_max"] = mkmmd_loss_max
+
+            if self.feature_l2_norm:
+                feature_l2_norm_loss = torch.linalg.norm(features["features"])
```
emersodb, line 224: Similar to Ditto, I wonder if we should normalize by the batch size?

fl4health/losses/contrastive_loss.py
```diff
@@ -0,0 +1,61 @@
...
+
+        Returns:
+            torch.Tensor: Contrastive loss value
+        """
+        # TODO: We can extend it to support multiple positive pairs using multi-label classification
+
+        assert self.temperature is not None
+        if len(positive_pairs) != 1:
+            AssertionError(
+                """Each feature can have one positive pair.
```
emersodb, line 47: A few minor points:

1. I think, though not 100% certain, you need to use `raise AssertionError(...)` for this to actually throw?
2. I believe the triple quotes here will mean that any \n values you have will also be logged. So I think it's better to just do 
```
AssertionError(
    "Each feature can have one positive pair. "
    "Thus positive pairs should be a tensor of shape (1, n_samples, n_features) "
    f"rather than {positive_pairs.shape}"
)
```
3. It would likely be useful to the user to also show the shape of the tensor they entered. See above

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,288 @@
...
+    ) -> None:
+        """
+        Compute the multi-kernel maximum mean discrepancy (MK-MMD) between the source and target domains. Also allows
+        for optimization of the coefficients, beta
+
+        Args:
+            device (torch.device): Device onto which tensors should be moved
+            gammas (Optional[torch.Tensor], optional): These are known as the length-scales of the RBF functions used
+                to compute the Mk-MMD distances. The length of this list defines the number of kernels used in the
+                norm measurement. If none, a default of 29 kernels is used. Defaults to None.
```
emersodb, line 29: Is the 29 kernels still correct after shifting the default?

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,288 @@
...
+        else:
+            self.gammas = gammas
+        self.kernel_num = len(self.gammas)
+
+        if betas is None:
+            rand_coefficients = torch.rand((self.kernel_num, 1)).to(self.device)
+            # normalize the coefficients to sum to 1
+            self.betas = (1 / torch.sum(rand_coefficients)) * rand_coefficients
+        else:
+            assert betas.shape == (self.kernel_num, 1)
```
emersodb, line 52: Perhaps we can also sanity check that the entered betas sum to 1?

fl4health/losses/mkmmd_loss.py
```diff
@@ -0,0 +1,288 @@
...
+            try:
+                unnormalized_betas = self.form_and_solve_qp(hat_d_per_kernel, regularized_Q_k)
+            except Exception as e:
+                # If we can't solve the QP due to infeasibility, then we keep the previous betas
+                log(INFO, f"{e} We keep previous betas.")
+                unnormalized_betas = self.betas
+        else:
+            # If we're trying to maximize the type II error, then we are trying to maximize a convex function over a
+            # convex polygon of beta values. So the maximum is found at one of the vertices
+            unnormalized_betas = self.get_best_vertex_for_objective_function(hat_d_per_kernel, regularized_Q_k)
```
emersodb, line 273: In our meeting, we briefly discussed clipping the betas at 0.0 (since we were sometimes seeing small negative betas in spite of the QP constraint.) We can do this to the unnormalized betas so they don't contribute oddly to the normalization 🙂

fl4health/utils/losses.py
```diff
@@ -221,15 +221,16 @@ def aggregate_losses_dict(
...
                 meter type
         """
-        # We don't know the keys of the dict (backward or additional losses) beforehand. We obtain them
-        # from the first entry because we know all of the losses will have the same keys
-        loss_keys = loss_list[0].keys()
-        num_losses = len(loss_list)
+        # We don't know the keys of the dict (backward or additional losses) beforehand. We don't obtain them
+        # from the first entry because losses can have different keys. We get list of all the keys from
+        # all the losses.
+        loss_keys = list(set(key for loss_dict_ in loss_list for key in loss_dict_.keys()))
```
emersodb, line 227: Do we need to convert the `set` here to a `list`? I think we can iterate through the set of loss_keys similar to the way loop through the list as it is.

research/flamby/fed_isic2019/ditto_mkmmd/README.md
```diff
@@ -0,0 +1,24 @@
...
+   path_to_desired_venv/
+```
+
+from the top level directory of the repository
+
+An example is something like
+``` bash
+./research/flamby/fed_isic2019/ditto_mkmmd/run_hp_sweep.sh \
+   research/flamby/fed_isic2019/ditto_mkmmd/config.yaml \
+   research/flamby/fed_isic2019/ditto/ \
```
emersodb, line 19: I think you want /ditto_mkmmd/ for this line as well?

research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,164 @@
...
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
+#   lambda value \
+#   server_address
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/ditto_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/ditto/hp_results/ \
```
emersodb, line 31: ditto_mkmmd here rather than just ditto?

research/flamby/fed_isic2019/ditto_mkmmd/run_hp_sweep.sh
```diff
@@ -0,0 +1,68 @@
...
+#  ./research/flamby/fed_isic2019/ditto_mkmmd/run_hp_sweep.sh \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/
+#
+# Example:
+# ./research/flamby/fed_isic2019/ditto_mkmmd/run_hp_sweep.sh \
+#   research/flamby/fed_isic2019/ditto_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/ditto/ \
```
emersodb, line 15: ditto_mkmmd here I think?

research/flamby/fed_isic2019/fenda_mkmmd/README.md
```diff
@@ -0,0 +1,24 @@
...
+   path_to_desired_venv/
+```
+
+from the top level directory of the repository
+
+An example is something like
+``` bash
+./research/flamby/fed_isic2019/fenda_mkmmd/run_hp_sweep.sh \
+   research/flamby/fed_isic2019/fenda_mkmmd/config.yaml \
+   research/flamby/fed_isic2019/fenda/ \
```
emersodb, line 19: fenda_mkmmd here?

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,162 @@
...
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
+#   server_address
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/fenda_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/fenda/hp_results/ \
```
emersodb, line 30: fenda_mkmmd here?

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,162 @@
...
+CLIENT_LR=$5
+SERVER_ADDRESS=$6
+MU=$7
+GAMMA=$8
+
+# Create the artifact directory
+mkdir "${ARTIFACT_DIR}"
+
+RUN_NAMES=( "Run1" )
+SEEDS=(2021 )
```
emersodb, line 74: As is, this script won't run 5 validations. I think we want to do that by default. So I think this is just an artifact of your experiments?

research/flamby/fed_isic2019/fenda_mkmmd/run_hp_sweep.sh
```diff
@@ -0,0 +1,66 @@
...
+#  ./research/flamby/fed_isic2019/fenda_mkmmd/run_hp_sweep.sh \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/
+#
+# Example:
+# ./research/flamby/fed_isic2019/fenda_mkmmd/run_hp_sweep.sh \
+#   research/flamby/fed_isic2019/fenda_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/fenda/ \
```
emersodb, line 15: fenda_mkmmd here?

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,162 @@
...
+#   client_side_learning_rate_value \
+#   server_address
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/fenda_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/fenda/hp_results/ \
+#   /Users/david/Desktop/FLambyDatasets/fedisic2019/ \
+#   /h/demerson/vector_repositories/fl4health_env/ \
+#   0.0001 \
```
emersodb, line 33: This example is leaving off mu and gamma values as part of the args I think?

research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,164 @@
...
+###############################################
+# Usage:
+#
+# sbatch research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
+#   lambda value \
```
emersodb, line 25: For the experiments here we don't use lambda but rather mu and l2 weights, I think

research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,164 @@
...
+#   server_address
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/ditto_mkmmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/ditto_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/ditto/hp_results/ \
+#   /Users/david/Desktop/FLambyDatasets/fedisic2019/ \
+#   /h/demerson/vector_repositories/fl4health_env/ \
+#   0.0001 \
```
emersodb, line 34: I think we're missing the l2 weight here?

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,162 @@
...
+
+###############################################
+# Usage:
+#
+# sbatch research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
```
emersodb, line 24: this doesn't have mu and gamma listed in the args 🙂

research/flamby/fed_isic2019/moon_mkmmd/client.py
```diff
@@ -0,0 +1,168 @@
...
+        action="store",
+        type=int,
+        help="Seed for the random number generators across python, torch, and numpy",
+        required=False,
+    )
+    parser.add_argument(
+        "--mu",
+        action="store",
+        type=float,
+        help="Weight for the auxiliary losses",
```
emersodb, line 123: These help strings are a bit generic. The might also be in the other clients. Being slightly more specific would be useful I think.

research/flamby/fed_isic2019/moon_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,168 @@
...
+# Usage:
+#
+# sbatch research/flamby/fed_isic2019/moon_mkmmd/run_fold_experiment.slrm \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
+#   server_address\
+#   client_mu
```
emersodb, line 26: I think we're missing gamma here

research/flamby/fed_isic2019/moon_mkmmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,168 @@
...
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/moon_mkmmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/moon_mkmmd/config.yaml \
+#   research/flamby/fed_isic2019/moon/hp_results/ \
+#   /Users/david/Desktop/FLambyDatasets/fedisic2019/ \
+#   /h/demerson/vector_repositories/fl4health_env/ \
+#   0.0001 \
+#   0.0.0.0:8080\
+#   0.1
```
emersodb, line 36: I think we're missing gamma here as well?

research/flamby/fed_isic2019/moon_mkmmd/run_fold_experiment.slrm
emersodb: Also the L2 penalty weight

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: I like the idea so much and I think in fact we should do for all Moon, Fenda and ditto methods. I can do refactor now as it shouldn't take so much time.

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: In new update I basically extend previous ditto_client. so now I have both tensors and init global model in ditto mlmmd client

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: Updated.

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: As we set both local model and global model the same init weights, I expect to tensor be same in the first round. But logs didn't seem so, thanks for pointing this out. I probably should double check this 🤔

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: But I thought of l2 norm as additional component of mkmmd loss, do we need to have it separate than mkmmd loss?

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: Sure that makes sense for me.

fl4health/clients/fenda_client.py
sanaAyrml: I think for now I can just separate extra additions to base fenda under name of fenda_mkmmd_client. Then we can think about other refactoring options. 
Also same thing for moon_client.

fl4health/clients/fenda_client.py
sanaAyrml: Where should we define such a function? in utils or separately for each client?

research/flamby/fed_isic2019/ditto_mkmmd/README.md
sanaAyrml: That is basically how I save my artifacts under ditto model but I can update it to ditto_mkmmd.

fl4health/losses/contrastive_loss.py
sanaAyrml: I agree with all of them. Thanks for recommendations

fl4health/losses/mkmmd_loss.py
sanaAyrml: It is actually 19.

fl4health/clients/moon_client.py
sanaAyrml: I think strapping is better as we are using some functionality of moon such as saving global moon and etc. I will just put up an error message about contrastive loss weight being zero and not being computed.

fl4health/losses/mkmmd_loss.py
sanaAyrml: I figured out in one of tests fails in that assertion error, should I change tests? Also is it better to raise error when preset betas are not sum to 1 or just normalize before setting betas?

sanaAyrml: It also fails when we set random betas as they don't sum up exactly to 1 due to computational approximation.

fl4health/clients/fenda_client.py
```diff
@@ -120,36 +193,17 @@ def get_cosine_similarity_loss(self, local_features: torch.Tensor, global_featur
...
         """
         Compute contrastive over current local features (z_p) with old local features (hat{z_p}) as positive pairs
         and current global features (z_s) as negative pairs.
         """
-        return self.compute_contrastive_loss(
+        return self.contrastive_loss(
             features=local_features,  # (z_p)
-            positive_pairs=old_local_features,  # (\hat{z_p})
-            negative_pairs=global_features,  # (z_s)
+            positive_pairs=old_local_features.unsqueeze(0),  # (\hat{z_p})
```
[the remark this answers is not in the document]
sanaAyrml: In this PR I refactored contrastive loss computation to a class that I use it both in fenda and moon.  In original moon contrastive loss implementation while negative pairs can be multiple we can have only one positive pair. That's why negative pairs has the shape of (n_pairs ,n_samples, n_features). For consistency I made positive pairs follow the same structure of (1 ,n_samples, n_features). That's why we have unsqueezed here and there. I can put a comment to clarify that but I don't think changing it in another place can be good idea as we also use it in cosine similarity function calculation.

fl4health/clients/moon_client.py
```diff
@@ -108,8 +99,77 @@ def update_before_train(self, current_server_round: int) -> None:
         # Save the parameters of the global model
         self.global_model = self.clone_and_freeze_model(self.model)
 
+        # Optimizing betas for the MK-MMD loss has not been done yet
+        self.betas_optimized = False
+
         return super().update_before_train(current_server_round)
 
+    def update_after_step(self, step: int) -> None:
+        if step > 0 and step % self.beta_update_interval == 0:
```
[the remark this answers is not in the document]
sanaAyrml: Good idea

fl4health/clients/moon_client.py
```diff
@@ -137,13 +197,34 @@ def compute_loss_and_additional_losses(
             "loss": loss,
         }
 
-        if self.contrastive_weight:
-            contrastive_loss = self.get_contrastive_loss(
-                features["features"], features["global_features"], features["old_features"]
+        if self.contrastive_weight and "old_features" in features:
```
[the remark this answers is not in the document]
sanaAyrml: The reason I am not checking is that global_features are always there but old_features appear after first iteration.

fl4health/losses/mkmmd_loss.py
emersodb: For the test, let's just set the beta values explicitly after the object is created to get around the assertion without having to re-compute the test? A bit of cheating, but I think that's fine.

For the numerical fluctuations, we can just do an approximate assertion, i.e. $\vert \sum betas - 1 \vert < 0.0001$ or something.

fl4health/clients/ditto_mkmmd_client.py
emersodb: Sounds good!

fl4health/clients/fenda_client.py
emersodb: Sounds good. I don't think we need to do a lot for this PR. Just want to make a note that we can probably make some stuff "prettier" with some additional thought 🙂

fl4health/clients/fenda_client.py
emersodb: Yeah that's a great question. This functionality is used several other classes too. I'd say a utils function for now I guess. I don't love just throwing functions randomly into a "catch-all" folder, but I think that makes the most sense as a quick thing. Maybe put a #TODO above the function to find a better place for it.

fl4health/clients/moon_client.py
emersodb: Okay,  I'll take a look at the implementation.

fl4health/clients/moon_client.py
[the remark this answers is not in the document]
emersodb: Ah I see. Makes sense.

fl4health/clients/ditto_mkmmd_client.py
sanaAyrml: I see what you mean, sure I can decouple it from mkmmd loss, so this scenario wouldn't happen

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
sanaAyrml: oh no sorry this is probably accident.

research/flamby/fed_isic2019/fenda_mkmmd/run_fold_experiment.slrm
sanaAyrml: Good catch

PR #117, opened 2024-03-13
Adding an implementation for FedDG-GA

lotif:
# PR Type
Feature

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8687d5yf4

Implementing FedDG-GA (Federated Domain Generalization with Generalization Adjustment, Zhang et al. 2023).

To make it work, a new parameter called `evaluate_after_fit` has been added to the client. It will perform an evaluation round with the local model against the client's validation data and return the metrics back to the client. It has to be set to `True` when running FedDG-GA, default is `False`

A `FedDGGAStrategy` class has been added to perform the domain generalization adjustment. It will:
- Collect the results at `aggregate_fit` and `aggregate_evaluate` and their metrics
- Calculate an adjustment weight per client based on their metrics
- Update each client's results based on their weight
- Perform the aggregation with the weighted results and update the parameters with the new adjusted parameters so they can be used in the next round.

# Tests Added

- Smoke tests for FedDG-GA using APFL
- Unit tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/VectorInstitute/FL4Health/117)


examples/feddg_ga_example/README.md
```diff
@@ -0,0 +1,44 @@
...
+that aims to produce a model that is better generalizable by implementing a variance reduction regularizer.
+It aims to achieve a tighter generalization with an explicit re-weighted aggregation as opposed to the implicit
+multi-domain data-sharing of conventional domain generalization methods.
+
+FedDG-GA is also a method that can be combined with other FL algorithms to improve their results. In this demo,
+FedDG-GA is applied in conjunction with APFL to an augmented version of the MNIST dataset that is non--IID. The FL
+server expects two clients to be spun up (i.e. it will wait until two clients report in before starting training). Each
+client has a modified version of the MNIST dataset. This modification essentially subsamples a certain number from the
+original training and validation sets of MNIST in order to synthetically induce local variations in the statistical
+properties  of the clients training/validation data. The proportion of labels at each client is determined by dirichlet
```
emersodb, line 13: Super minor, but Dirichlet is normally capitalized 🙂

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+                self.metric_name = metric_type.value
+            if signal == 0.0:
+                self.signal = FairnessMetricType.signal_for_type(metric_type)
+
+
+# Initial value for the adjustment weight of each client model
+INITIAL_ADJUSTMENT_WEIGHT = 1.0 / 3.0
+
+
+class FedDGGAStrategy(FedAvg):
```
emersodb, line 79: Very minor, but in other places, we've used the acronym naming convention of first letter uppercase, remainder lowercase (FENDA -> Fenda). This is a bit of a weird combo acronym. So I think our name would be `FedDgGaStrategy`? Sort of ugly....but follows the "rules"

fl4health/clients/basic_client.py
```diff
@@ -273,6 +273,20 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
         # Update after train round (Used by Scaffold and DP-Scaffold Client to update control variates)
         self.update_after_train(local_steps, loss_dict)
 
+        # Check if we should run an evaluation with validation data after fit (used by FedDGGA)
+        try:
+            evaluate_after_fit = self.narrow_config_type(config, "evaluate_after_fit", bool)
+        except ValueError:
+            evaluate_after_fit = False
```
emersodb, line 280: Rather than doing this try catch here, we can move this functionality into `process_config`?

fl4health/clients/basic_client.py
```diff
@@ -273,6 +273,20 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
...
+
+        if evaluate_after_fit:
+            loss, metric_values = self.validate()
```
emersodb, line 283: perhaps we can pull this, and the below, into a function called `evaluated_after_fit` or something like it. That way, participants could potentially override it to get different behavior as well?

lotif (commented):
> *[Reviewable](https://reviewable.io/reviews/VectorInstitute/FL4Health/117)* status: 0 of 15 files reviewed, 5 unresolved discussions (waiting on @emersodb, @fatemetkl, @jewelltaylor, @sanaAyrml, @yc7z, and @zxj-c)

___
*[`examples/feddg_ga_example/config.yaml` line 8 at r1](https://reviewable.io/reviews/VectorInstitute/FL4Health/117#gh-1539480528:-Nu0T5_kCeCSJ6HX81Is:b-896fix) ([raw file](https://github.com/VectorInstitute/FL4Health/blob/8f5fa01b18cfc0812c712cd757c9a772f993b5ab/examples/feddg_ga_example/config.yaml#L8)):*
<details><summary><i>Previously, emersodb (David Emerson) wrote…</i></summary><blockquote>

Super minor: I'd suggest maybe changing the comment here to be a bit more explicit. Maybe something like "Evaluates model immediately after local training on the validation set (in addition to the training set)"
</blockquote></details>

Done.
___
*[`examples/feddg_ga_example/README.md` line 13 at r1](https://reviewable.io/reviews/VectorInstitute/FL4Health/117#gh-1539429372:-Nu0T3RYBkdBmWNRuGjM:b-896fix) ([raw file](https://github.com/VectorInstitute/FL4Health/blob/8f5fa01b18cfc0812c712cd757c9a772f993b5ab/examples/feddg_ga_example/README.md#L13)):*
<details><summary><i>Previously, emersodb (David Emerson) wrote…</i></summary><blockquote>

Super minor, but Dirichlet is normally capitalized 🙂 
</blockquote></details>

Done.
___
*[`fl4health/clients/basic_client.py` line 280 at r1](https://reviewable.io/reviews/VectorInstitute/FL4Health/117#gh-1539980339:-Nu0T01v-FIClf6IBq5I:b-896fix) ([raw file](https://github.com/VectorInstitute/FL4Health/blob/8f5fa01b18cfc0812c712cd757c9a772f993b5ab/fl4health/clients/basic_client.py#L280)):*
<details><summary><i>Previously, emersodb (David Emerson) wrote…</i></summary><blockquote>

Rather than doing this try catch here, we can move this functionality into `process_config`?
</blockquote></details>

Done.
___
*[`fl4health/clients/basic_client.py` line 283 at r1](https://reviewable.io/reviews/VectorInstitute/FL4Health/117#gh-1539981634:-Nu0T0ZvC_qw7dUzWrGa:b-896fix) ([raw file](https://github.com/VectorInstitute/FL4Health/blob/8f5fa01b18cfc0812c712cd757c9a772f993b5ab/fl4health/clients/basic_client.py#L283)):*
<details><summary><i>Previously, emersodb (David Emerson) wrote…</i></summary><blockquote>

perhaps we can pull this, and the below, into a function called `evaluated_after_fit` or something like it. That way, participants could potentially override it to get different behavior as well?
</blockquote></details>

Done.
___
*[`fl4health/strategies/feddg_ga_strategy.py` line 79 at r1](https://reviewable.io/reviews/VectorInstitute/FL4Health/117#gh-1539535571:-Nu0RCM5AT-q4B4GUcOW:b-a0dm2n) ([raw file](https://github.com/VectorInstitute/FL4Health/blob/8f5fa01b18cfc0812c712cd757c9a772f993b5ab/fl4health/strategies/feddg_ga_strategy.py#L79)):*
<details><summary><i>Previously, emersodb (David Emerson) wrote…</i></summary><blockquote>

Very minor, but in other places, we've used the acronym naming convention of first letter uppercase, remainder lowercase (FENDA -> Fenda). This is a bit of a weird combo acronym. So I think our name would be `FedDgGaStrategy`? Sort of ugly....but follows the "rules"
</blockquote></details>

Yeah, pretty ugly :/ I'll do it but I prefer the other way even if it break the rules a bit.

fl4health/strategies/feddg_ga_strategy.py
lotif: Yeah, pretty ugly :/ I'll do it but I prefer the other way even if it break the rules a bit.

fl4health/clients/basic_client.py
lotif: Moved, please check.

fl4health/clients/basic_client.py
lotif: Done, please check.

examples/feddg_ga_example/server.py
```diff
@@ -0,0 +1,99 @@
...
+        on_fit_config_fn=fit_config_fn,
+        # We use the same fit config function, as nothing changes for eval
+        on_evaluate_config_fn=fit_config_fn,
+        fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+        evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        initial_parameters=get_all_model_parameters(initial_model),
+    )
+
+    client_manager = FixedSamplingClientManager()
+    server = FlServer(strategy=strategy, client_manager=client_manager)
```
emersodb, line 65: One thing we might want to think about here is either:
1. We put a check in the FlServer class to ensure that when the strategy being used is FedDgGaStrategy, then the client manager must have type `FixedSamplingClientManager`.
2. We create a server class on top of FlServer that enforces that explicitly.

That way someone doesn't accidentally use a different client manager and therefore get weird but hard to debug results?

emersodb: Ah I see, you actually do this assertion in the strategy. That's even better. Feel free to ignore.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+
+        Returns:
+            (List[Tuple[ClientProxy, FitIns]]) the input for the clients' fit function.
+        """
+        client_fit_ins = super().configure_fit(server_round, parameters, client_manager)
+
+        assert isinstance(
+            client_manager, FixedSamplingClientManager
+        ), f"Client manager is not of type FixedSamplingClientManager: {type(client_manager)}"
+
```
emersodb, line 202: Since you have access to the `on_fit_config` here, we probably also want to confirm that `evaluate_after_fit` is true here and provide an informative error message if it is not?

fl4health/client_managers/fixed_sampling_client_manager.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def sample(
+        self,
+        num_clients: int,
+        min_num_clients: Optional[int] = None,
+        criterion: Optional[Criterion] = None,
+    ) -> List[ClientProxy]:
+        """
+        Return a new client sample for the first time it runs. For subsequent runs,
+        it will return the sample sampling until self.reset_sampling() is called.
```
emersodb, line 27: I think maybe you mean to write "...it will return the **same** sampling until..." here, but I'm not certain 😂

fl4health/clients/basic_client.py
```diff
@@ -273,6 +278,10 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
         # Update after train round (Used by Scaffold and DP-Scaffold Client to update control variates)
         self.update_after_train(local_steps, loss_dict)
 
+        # Check if we should run an evaluation with validation data after fit (used by FedDGGA)
```
emersodb, line 281: Very minor, but I'd just rephrase the parenthetical to (for example, this is used by FedDGGA). The only reason I suggest that, is I actually plan to use this functionality to expand our options for checkpointing as well 🙂

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+        Args:
+            fairness_metric_type: (FairnessMetricType) the fairness metric type.
+
+        Returns: (float) -1.0 if FairnessMetricType.ACCURACY, 1.0 if FairnessMetricType.LOSS and
+            0 if FairnessMetricType.CUSTOM.
+        """
+        if fairness_metric_type == FairnessMetricType.ACCURACY:
+            return -1.0
+        if fairness_metric_type == FairnessMetricType.LOSS:
+            return 1
```
emersodb, line 37: super minor, but once in a while libraries like pytorch will be cranky if a float is specified as 1 instead of 1.0 because types are inferred...could we return 1.0 and 0.0 for these guys 😂

fatemetkl (commented):
Thanks Marcelo for this PR. I think this is a very nice addition. I just had a few comments/questions.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+    ):
+        """
+        Instantiates a fairness metric with a type and optional metric name and
+            signal if one wants to override them.
+
+        Args:
+            metric_type: (FairnessMetricType) the fairness metric type. If CUSTOM, the metric_name and
+                signal should be provided.
+            metric_name: (str, optional) the name of the metric to be used as fairness metric.
+                Optional, default is metric_type.value.
```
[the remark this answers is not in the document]
emersodb: Ah, I see what's happening below. Rather than using a real value for these, we could declare them as
```
metric_name: Optional[str] = None,
signal: Optional[float] = None,
```
and below we can just do 

```
if metric_type is FairnessMetricType.CUSTOM:
    assert metric_name is not None and signal is not None
else:
    if metric_name is None:
        self.metric_name = metric_type.value
    if signal is None:
        self.signal = FairnessMetricType.signal_for_type(metric_type)
```

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            on_fit_config_fn=on_fit_config_fn,
+            on_evaluate_config_fn=on_evaluate_config_fn,
+            accept_failures=accept_failures,
+            initial_parameters=initial_parameters,
+            fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+            evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        )
+
+        if fairness_metric is None:
+            self.fairness_metric = FairnessMetric(FairnessMetricType.LOSS)
```
emersodb, line 164: We can make the `fairness_metric` input default to this fairness metric by default in the arguments above instead of None and then just do `self.fairness_metric = fairness_metric` here I think?

emersodb: hmmm...actually, because its a class that's a bad idea 😂

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            server_round: (int) the current server round.
+            results: (List[Tuple[ClientProxy, FitRes]]) The clients' fit results.
+            failures: (List[Union[Tuple[ClientProxy, FitRes], BaseException]]) the clients' fit failures.
+
+        Returns:
+            (Tuple[Optional[Parameters], Dict[str, Scalar]]) A tuple containing the aggregated parameters
+                and the aggregated fit metrics.
+        """
+
+        _, metrics_aggregated = super().aggregate_fit(server_round, results, failures)
```
emersodb, line 234: A comment as to why we're "discarding" the parameters here would be useful, I think.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+    def aggregate_evaluate(
+        self,
+        server_round: int,
+        results: List[Tuple[ClientProxy, EvaluateRes]],
+        failures: List[Union[Tuple[ClientProxy, EvaluateRes], BaseException]],
+    ) -> Tuple[Optional[float], Dict[str, Scalar]]:
+        """
+        Aggregate evaluation losses using weighted average.
+
+        Collects the evaluation metrics and update the adjustment weights, which will be used
```
emersodb, line 253: "and **updates** the adjustment weights" 🙂

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        loss_aggregated, metrics_aggregated = super().aggregate_evaluate(server_round, results, failures)
+
+        self.evaluation_metrics = {}
+        for result in results:
+            cid = result[0].cid
+            self.evaluation_metrics[cid] = result[1].metrics
+            # adding the loss to the metrics
+            val_loss_key = FairnessMetricType.LOSS.value
+            self.evaluation_metrics[cid][val_loss_key] = loss_aggregated  # type: ignore
```
emersodb, line 273: Out of curiosity, why do we need the type ignore here?

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            failures: (List[Union[Tuple[ClientProxy, FitRes], BaseException]]) the clients' evaluate failures.
+
+        Returns:
+            (Tuple[Optional[float], Dict[str, Scalar]]) A tuple containing the aggregated evaluation loss
+                and the aggregated evaluation metrics.
+        """
+        loss_aggregated, metrics_aggregated = super().aggregate_evaluate(server_round, results, failures)
+
+        self.evaluation_metrics = {}
+        for result in results:
```
emersodb, line 268: Super minor, but since results is a tuple, it might be slightly easier to read if we unpack the tuple here explicitly. That is, 
```
for client_proxy, eval_res in results:
....
```

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+
+        Args:
+            results: (List[Tuple[ClientProxy, FitRes]]) The clients' fit results.
+
+        Returns:
+            (NDArrays) the weighted and aggregated results.
+        """
+
+        aggregated_results: Optional[NDArrays] = None
+        for result in results:
```
emersodb, line 293: Similar to the above, perhaps we can explicitly unpack this tuple

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            cid = result[0].cid
+
+            # initializing adjustment weights for this client if they don't exist yet
+            if cid not in self.adjustment_weights:
+                self.adjustment_weights[cid] = INITIAL_ADJUSTMENT_WEIGHT
```
emersodb, line 298: It we don't have an adjustment weight for a client (i.e. it joins the fitting round later on or something), then the sum of the adjustment weights will no longer be 1.0. If we add in a client later on, should we not normalize the weight adjustment coefficients to preserve this property?

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+    def update_weights_by_ga(self, server_round: int, cids: List[str]) -> None:
+        """
+        Update the self.adjustment_weights dictionary by calculating the new weights
+        based on the current server round, fit and evaluation metrics.
+
+        Args:
+            server_round: (int) the current server round.
+            cids: (List[str]) the list of client ids that participated in this round.
+        """
+        value_list = []
```
emersodb, line 325: Value list is a bit generic, I think what you're calculating here is the average generalization gap (mu)

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+        step size and total number of rounds.
+
+        Args:
+            server_round: (int) the current server round
+
+        Returns: (float) the current value for the weight step size.
+        """
+        assert self.num_rounds is not None
+        weight_step_size_decay = self.weight_step_size / self.num_rounds
+        weight_step_size_for_round = self.weight_step_size - ((server_round - 1) * weight_step_size_decay)
```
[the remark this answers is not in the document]
emersodb: I agree with Fatemeh, I think the way this is written my not be algebraically equivalent actually, if I'm reading the variables correctly.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+            evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        )
+
+        if fairness_metric is None:
+            self.fairness_metric = FairnessMetric(FairnessMetricType.LOSS)
+        else:
+            self.fairness_metric = fairness_metric
+
+        self.weight_step_size = weight_step_size
```
emersodb, line 168: I believe we want the `weight_step_size` to be > 0.0 and < 1.0. Perhaps we can add an assertion to that effect here?

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+        step_size = (1.0 / 3.0) * self.get_current_weight_step_size(server_round)
+
+        # updating weights
+        new_total_weight = 0.0
+        for i in range(len(cids)):
+            cid = cids[i]
+            self.adjustment_weights[cid] += self.fairness_metric.signal * norm_gap_list[i] * step_size
+
+            # weight clip
+            clipped_weight = np.clip(self.adjustment_weights[cid], 0.0, 1.0)
```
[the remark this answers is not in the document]
emersodb: Seconding Fatemeh's suggestion. The normalization will get you back to the unit ball. If they do this clipping in practice in their code implementation we should put a comment here stating so.

fl4health/client_managers/fixed_sampling_client_manager.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class FixedSamplingClientManager(SimpleClientManager):
+    """Keeps sampling fixed until it's reset"""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.current_sample: Optional[List[ClientProxy]] = None
+
+    def reset_sample(self) -> None:
```
emersodb, line 15: It doesn't look like this is called anywhere in the code? I think maybe you're using this to maintain the same clients sampled between evaluate and fit rounds. However, since this isn't going to be called by the server anywhere, I think the effect is that we sample at the beginning of FL and never resample again?

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            fairness_metric_type: (FairnessMetricType) the fairness metric type.
+
+        Returns: (float) -1.0 if FairnessMetricType.ACCURACY, 1.0 if FairnessMetricType.LOSS and
+            0 if FairnessMetricType.CUSTOM.
+        """
+        if fairness_metric_type == FairnessMetricType.ACCURACY:
+            return -1.0
+        if fairness_metric_type == FairnessMetricType.LOSS:
+            return 1
+        return 0
```
emersodb, line 38: I think returning a 0 here is sort of "hiding" an issue. If I understand how the signal value works below, a signal value of 0 essentially means that we're nulling any updates to the GA weights. Perhaps we can through an error (or warning) here instead stating the resulting signal will be zero, which is very likely not what the user wants.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+        else:
+            norm_gap_list = value_list_ndarray / max_value
+
+        step_size = (1.0 / 3.0) * self.get_current_weight_step_size(server_round)
+
+        # updating weights
+        new_total_weight = 0.0
+        for i in range(len(cids)):
+            cid = cids[i]
+            self.adjustment_weights[cid] += self.fairness_metric.signal * norm_gap_list[i] * step_size
```
emersodb, line 358: Perhaps a note here on the signal value is worthwhile. Something like
```
For loss values, large and **positive** gaps imply worse generalization of global weights to local models. Therefore, we want to **increase** weight for these model parameters to improve generalization. So signal is positive. For accuracy, large **negative** gaps imply worse generalization. So the signal is -1.0, to increase weights for the associated model parameters.
```

examples/feddg_ga_example/server.py
```diff
@@ -0,0 +1,99 @@
...
+        config["batch_size"],
+        config["n_server_rounds"],
+        local_epochs=config.get("local_epochs"),
+        local_steps=config.get("local_steps"),
+        evaluate_after_fit=config.get("evaluate_after_fit", False),
+    )
+
+    initial_model = ApflModule(MnistNetWithBnAndFrozen())
+
+    # Server performs simple FedAveraging as its server-side optimization strategy
```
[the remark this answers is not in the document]
lotif: That's a good idea. David had some other ides to test this but I'll keep this in mind if I need to test further.

fl4health/clients/basic_client.py
lotif: Makes sense, and it's interesting you have other uses for this!

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
lotif: Makes sense, will do!

fl4health/strategies/feddg_ga_strategy.py
lotif: Yep, makes sense. Will change to raise an exception.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            assert metric_name is not None and metric_name != "" and signal is not None
+        else:
+            if metric_name == "":
+                self.metric_name = metric_type.value
+            if signal == 0.0:
+                self.signal = FairnessMetricType.signal_for_type(metric_type)
+
+
+# Initial value for the adjustment weight of each client model
+INITIAL_ADJUSTMENT_WEIGHT = 1.0 / 3.0
```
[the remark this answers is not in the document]
lotif: Yes, I now realize what that means. I saw it in the implemented code and was slightly confused but wrongfully didn't question further why that number was chosen.

fl4health/strategies/feddg_ga_strategy.py
lotif: Yep! The most annoying python bug/gotcha they insist on not fixing!!!

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+
+        Returns:
+            (Tuple[Optional[Parameters], Dict[str, Scalar]]) A tuple containing the aggregated parameters
+                and the aggregated fit metrics.
+        """
+
+        _, metrics_aggregated = super().aggregate_fit(server_round, results, failures)
+
+        self.train_metrics = {}
+        for result in results:
```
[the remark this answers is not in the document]
lotif: Good suggestion, I didn't know we could do that.

fl4health/strategies/feddg_ga_strategy.py
lotif: Because `loss_aggregated` can apparently be None (type is `Optional[float]`) and mypy does't like that being assigned to `Scalar`.

fl4health/strategies/feddg_ga_strategy.py
lotif: Yes, Fatemeh commented on it, an overlook on my part. I'm changing it to 1/number-of-clients.

fl4health/strategies/feddg_ga_strategy.py
emersodb: Ah right, I see. If the value is, in fact, None, does other stuff break? That is, should we just type narrow by asserting that `loss_aggregated is not None` here?

fl4health/strategies/feddg_ga_strategy.py
lotif: About adding a client, I am not sure what to do. Right now, it's easier in the code to recalculate the weight adjustment at the beginning of every fit_round and maybe that is a good behaviour unless the amount of clients changes dramatically between rounds.

fl4health/strategies/feddg_ga_strategy.py
lotif: I'm adding this comment to `FairnessMetricType.signal_for_type` as well.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            log(
+                WARNING,
+                "Max value in metric diff list is 0. Adjustment weights will remain the same. "
+                + f"Value list: {value_list}",
+            )
+            norm_gap_list = np.zeros(len(value_list))
+        else:
+            norm_gap_list = value_list_ndarray / max_value
+
+        step_size = (1.0 / 3.0) * self.get_current_weight_step_size(server_round)
```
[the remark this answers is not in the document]
lotif: That is also copied from the implementation. Do you think they meant to use the same calculation ad the default adjustment weight here (1/M)?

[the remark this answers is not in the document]
emersodb: Very weird. On initial thought, I imagine that we should go with 1/M here. However, this damping for $d^r$ is not mentioned in the paper. So I think we want to do two things.  
1. Confirm that this is, in fact, 1/3 in the code base specifically for scaling $d^r$ and not tangled in with another calculation somewhere. 
2. If it is truly a damping parameter, we should put a comment in the code here pointing to this scaling in the code base and mention that it differs from the paper.

fl4health/strategies/feddg_ga_strategy.py
emersodb: Yeah I think that's fine. In practice, at least based on my reading, this method isn't really well defined for a situation where you are adding and losing clients in each round to be honest.

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
lotif: Again, pulled from the implementation:

https://github.com/MediaBrain-SJTU/FedDG-GA/blob/master/algorithms/fedavg/train_pacs_GA.py#L83

I guess next time I should rely more on the paper instead  😅

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
lotif: This is the code: 

https://github.com/MediaBrain-SJTU/FedDG-GA/blob/master/utils/weight_adjust.py#L18

the multiplication by 1/3 is done after $d^r$ is calculated. Will add a comment.

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
lotif: They do that in the code, I guess just as an extra guarantee?

https://github.com/MediaBrain-SJTU/FedDG-GA/blob/master/utils/weight_adjust.py#L31

Will add the comment.

fl4health/strategies/feddg_ga_strategy.py
lotif: I was assuming if loss is None we will probably have bigger problems but I guess it doesn't hurt to add an assertion here.

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
emersodb: So I'm honestly not sure what they are doing here, as their round index starts at 0. So subtracting 1 from it makes no sense. However, in the paper, their first local round of training starts at zero. So for them $(1-r/R)\cdot d=d$ for the first local round. If we start at `server_round = 1` (which I think we do), then doing $d\cdot (1-(r-1)/R)$ as you have coded gets us back to that expression correctly. 
1.  So if you confirm that our server rounds start at 1, then your formulation is correct and we should add a comment that this differs from the paper because our rounds start at 1 not 0.
2. If our server rounds start at 0 (I don't think so but not 100% sure), then we need to use the paper formulation
3. There implementation is dumb either way 😂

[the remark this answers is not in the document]
lotif: Our server round starts at 1. I'm confused now. In our case, I should keep this code as is?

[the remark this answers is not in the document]
lotif: makes sense, thanks!!!

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
emersodb: Oh man...there code is awful. I see why you didn't have the mean gap adjustments calculations etc. in the weight refinement calculations. They deserve to have to rescind their paper for not following their own algorithm. So my vote is that we don't include this at all. It doesn't fit with their stated algorithm in their paper at all. We can have a comment that they had some weird scaler in their code, but I don't think we want to keep it around.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,382 @@
...
+            assert (
+                cid in self.train_metrics and cid in self.evaluation_metrics
+            ), f"{cid} not in {self.train_metrics.keys()} or {self.evaluation_metrics.keys()}"
+
+            global_model_metric_value = self.evaluation_metrics[cid][self.fairness_metric.metric_name]
+            local_model_metric_value = self.train_metrics[cid][self.fairness_metric.metric_name]
+            assert isinstance(global_model_metric_value, float) and isinstance(local_model_metric_value, float)
+
+            value_list.append(global_model_metric_value - local_model_metric_value)
+
```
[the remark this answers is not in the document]
lotif: Fixed the implementation of this part to be the same as the paper. Tested it with a 5 test clients for 5 rounds and seems to be working well. Please check again when you have a minute.

lotif:
Thank @emersodb and @fatemetkl for the thorough review and great comments! It is ready to review again, please check when you have a minute.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,412 @@
...
+            local_model_metric_value = self.train_metrics[cid][self.fairness_metric.metric_name]
+            assert isinstance(global_model_metric_value, float) and isinstance(local_model_metric_value, float)
+
+            generalization_gaps.append(global_model_metric_value - local_model_metric_value)
+
+        # Calculating the normalized the generalization gaps
+        generalization_gaps_ndarray = np.array(generalization_gaps)
+        mean_generalization_gap = np.mean(generalization_gaps_ndarray)
+        var_generalization_gaps = generalization_gaps_ndarray - mean_generalization_gap
+        max_var_generalization_gap = np.max(var_generalization_gaps)
```
emersodb, line 357: As a small modification, here that they did have in their code and I think makes sense is to take the absolute value of the `var_generalization_gaps` before grabbing the max.

fl4health/strategies/feddg_ga_strategy.py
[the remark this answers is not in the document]
lotif: Ok, makes sense, will remove. And yeah, it was a bit of a challenge to adapt the code, but for me the lesson learned for next time is to not trust the implementation!

fl4health/strategies/feddg_ga_strategy.py
lotif: Done.

fatemetkl (commented):
LGTM. Thanks @lotif for addressing all the comments!  
For that 1/3 constant maybe we can later create an issue on their repo.

emersodb:
> LGTM. Thanks @lotif for addressing all the comments! For that 1/3 constant maybe we can later create an issue on their repo.

In their paper, most of their examples have 4 clients. If you're doing leave-one-out eval, that means there are only 3 clients as part of the training. So it feels related to the number of clients, but makes no sense where they put it. They also never compute the mean gaps ($\mu$) anywhere in their code, which is crazy, given its a central part of their weight adaptation process

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -322,45 +335,54 @@ def update_weights_by_ga(self, server_round: int, cids: List[str]) -> None:
...
             local_model_metric_value = self.train_metrics[cid][self.fairness_metric.metric_name]
             assert isinstance(global_model_metric_value, float) and isinstance(local_model_metric_value, float)
 
-            value_list.append(global_model_metric_value - local_model_metric_value)
+            generalization_gaps.append(global_model_metric_value - local_model_metric_value)
 
-        # calculating norm gap
-        value_list_ndarray = np.array(value_list)
-        max_value = np.max(np.abs(value_list_ndarray))
+        # Calculating the normalized the generalization gaps
```
emersodb, line 353: Super minor: Just a typo `Calculating the normalized generalization gaps`

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -322,45 +335,54 @@ def update_weights_by_ga(self, server_round: int, cids: List[str]) -> None:
...
+        if max_var_generalization_gap == 0:
             log(
                 WARNING,
-                "Max value in metric diff list is 0. Adjustment weights will remain the same. "
-                + f"Value list: {value_list}",
+                "Max variance in generalization gap is 0. Adjustment weights will remain the same. "
+                + f"Generalization gaps: {generalization_gaps}",
             )
-            norm_gap_list = np.zeros(len(value_list))
+            normalized_generalization_gaps = np.zeros(len(generalization_gaps))
```
emersodb, line 365: Super minor, but rather than doing `np.zeros(len(generalization_gaps)` you can simplify it as `np.zeros_like(generalization_gaps)`

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,411 @@
...
+        # For loss values, large and **positive** gaps imply worse generalization of global
+        # weights to local models. Therefore, we want to **increase** weight for these model
+        # parameters to improve generalization. So signal is positive. For accuracy, large
+        # **negative** gaps imply worse generalization. So the signal is -1.0, to increase
+        # weights for the associated model parameters.
+        if fairness_metric_type == FairnessMetricType.ACCURACY:
+            return -1.0
+        if fairness_metric_type == FairnessMetricType.LOSS:
+            return 1.0
+        raise Exception("This function should not be called with CUSTOM type.")
```
emersodb, line 44: Super minor, but perhaps we use a specific exception class rather than a bare one?

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,411 @@
...
+                The metric to evaluate the local model of each client against the global model in order to
+                determine their adjustment weight for aggregation. Can be set to any default metric in
+                FairnessMetricType or set to use a custom metric. Optional, default is
+                FairnessMetric(FairnessMetricType.LOSS).
+            weight_step_size : float
+                The step size to determine the magnitude of change for the adjustment weight. It has to be
+                0 < weight_step_size < 1. Optional, default is 0.2.
+        """
+        super().__init__(
+            fraction_fit=fraction_fit,
```
emersodb, line 151: For this method, one thing I've been thinking about is that it's sort of "ill-defined" for situations where you only have some of the clients participating in any given round. The paper doesn't really contemplate this anywhere that I can find. Because of that, I think it would be good practice to warn the user that the algorithms convergence and behavior are not guaranteed if `fraction_fit` and `fraction_evaluate` are not 1.0, even though it is technically supported? No big changes needed, just logging a warning is sufficient.

fl4health/strategies/feddg_ga_strategy.py
```diff
@@ -0,0 +1,411 @@
...
+        loss_aggregated, metrics_aggregated = super().aggregate_evaluate(server_round, results, failures)
+
+        self.evaluation_metrics = {}
+        for client_proxy, eval_res in results:
+            cid = client_proxy.cid
+            self.evaluation_metrics[cid] = eval_res.metrics
+            # adding the loss to the metrics
+            val_loss_key = FairnessMetricType.LOSS.value
+            assert loss_aggregated is not None
+            self.evaluation_metrics[cid][val_loss_key] = loss_aggregated
```
emersodb, line 285: I'm a little bit confused as to why we're adding in the aggregated loss here rather than the individual loss values on each client after weight aggregation? That is, I believe we should be storing $G_{\hat{D}_i}(\theta^r)$ here for each individual $i$, rather than their weighted average of these values.

emersodb: Perhaps I'm missing something here though.

emersodb: The refactor would be as simple as using `eval_res.loss` here instead of `loss_aggregated` I think

lotif: That's right, thanks for catching it.

lotif:
@emersodb I've addressed your comments, please check again when you have time.

emersodb (approved):
Changes look good to me! Let's fix up the tests and get it merged 🎉

PR #119, opened 2024-03-18
Fixing an issue with Opacus modified models and optimizers.

emersodb:
# PR Type
Fix

# Short Description

Addressing an issue that can arise when opacus replaces layers that are not DP compliant and the parameters have misalignment in the optimizer. Also adding in an example of "safely" setting the server initial parameters..

# Tests Added

Added a few tests to ensure that the functionality is indeed replacing the desired layers and that the new layers are also being replaced in the optimizer itself.


examples/dp_fed_examples/instance_level_dp/server.py
```diff
@@ -52,6 +53,14 @@ def fit_config(
     )
 
 
+def get_model_and_validate() -> nn.Module:
```
emersodb, line 56: Note: This model wasn't causing any problems, but I wanted the examples to reflect a situation where the server should pass the appropriate model parameters **after** validation and correction to ensure that all the model weights line up properly

examples/dp_scaffold_example/server.py
```diff
@@ -12,13 +13,15 @@
...
+from fl4health.utils.functions import privacy_validate_and_fix_modules
 from fl4health.utils.metric_aggregation import evaluate_metrics_aggregation_fn, fit_metrics_aggregation_fn
 
 
 def get_initial_model_information() -> Tuple[Parameters, Parameters]:
     # Initializing the model parameters on the server side.
     # Currently uses the Pytorch default initialization for the model parameters.
-    initial_model = MnistNet()
+    initial_model: nn.Module = MnistNet()
+    initial_model, _ = privacy_validate_and_fix_modules(initial_model)
```
emersodb, line 24: Note: This model wasn't causing any problems, but I wanted the examples to reflect a situation where the server should pass the appropriate model parameters **after** validation and correction to ensure that all the model weights line up properly

zxj-c (approved):

fl4health/utils/functions.py
```diff
@@ -1,8 +1,30 @@
...
+    # Validate that the model layers are compatible with privacy mechanisms in Opacus and try to replace the layers
+    # with compatible ones if necessary.
+
+    errors = ModuleValidator.validate(model, strict=False)
+    reinitialize_optimizer = len(errors) > 0
+    # Due to a bug in Opacus, it's possible that we need to run multiple rounds fo module validator fix for
+    # complex nested models to fully replace all layers within a model (for example, in the Fed-IXI model)
+    while len(errors) != 0:
+        for error in errors:
+            log(WARNING, f"Opacus error: {error}")
```
jewelltaylor, line 26: Not sure if this error mention that the layers are being replaced, but its probably a good idea to surface to the user if not

fl4health/utils/functions.py
```diff
@@ -1,8 +1,30 @@
...
 from flwr.common.typing import Parameters
+from opacus.validators import ModuleValidator
 
 
 def get_all_model_parameters(model: nn.Module) -> Parameters:
     # Extracting all model parameters and converting to Parameters object
     return ndarrays_to_parameters([val.cpu().numpy() for _, val in model.state_dict().items()])
+
+
+def privacy_validate_and_fix_modules(model: nn.Module) -> Tuple[nn.Module, bool]:
```
jewelltaylor, line 16: nit: Not a huge deal at all, but I don't love the naming on this file (functions.py). I think we should avoid having catch alls where we accumulate vaguely related functions. Its not super intuitive where to find things.

fl4health/clients/instance_level_privacy_client.py
```diff
@@ -47,16 +45,17 @@ def setup_client(self, config: Config) -> None:
         super().setup_client(config)
 
         # Configure DP training
-        self.setup_opacus_objects()
+        self.setup_opacus_objects(config)
 
-    def setup_opacus_objects(self) -> None:
+    def setup_opacus_objects(self, config: Config) -> None:
```
jewelltaylor, line 50: Since we are already having to do this interval validation on the client, should we also move the validation internally on the server side. That way to the user doesn't need to know anything about the validation function you wrote and its a more seamless experience. I may be overlooking some details though so let me know

emersodb: The only reason I didn't do that is because then we'd have to pass the models to the strategies for manipulation. This is sort of already available in SCAFFOLD strategies, but it isn't for BasicFedAvg, which is used by this client. If we did that, I'd want to create my own strategies, so that it isn't hardcoded into SCAFFOLD or BasicFedAvg. That's because Opacus doesn't play nice with some of the more sophisticated models (like Fed-IXI's U-net).

emersodb: Maybe we can put this in the backlog to consider?

fl4health/utils/functions.py
emersodb: Lol...that's fair. I'll put something in the backlog to change this, since it also means we probably want to move `get_all_model_parameters`

fl4health/clients/instance_level_privacy_client.py
jewelltaylor: sounds good!

jewelltaylor (approved):
LGTM!

fl4health/utils/functions.py
emersodb: That's a good point. I'll add this to the logs.

fl4health/clients/instance_level_privacy_client.py
emersodb: Added to the backlog.

fl4health/utils/functions.py
emersodb: Added to the backlog

fl4health/utils/functions.py
emersodb: Added more descriptive logging and tested that it prints the warnings

PR #120, opened 2024-03-20
Update torch, flwr and other requirements

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Work on upgrading Flower version from 1.4 to the most recent version 1.7](https://app.clickup.com/t/8687wyq8b)

It has been a while since we have updated some of the core dependencies in FL4Health. Since then, there has been a major release of pytorch and several minor releases for flwr that have functionality that is helpful for our use cases. To that end, I have ventured to upgrade torch and flwr. In the process a lot of other packages needed to be updated. Thankfully there were only a few minor instances where I had to change anything. The results are of the smoke tests were very similar but the changes fell slight outside of the tolerance on the equality operation so I had to update the numbers. 

Also, since flwr 1.7 requires cryptography 41.0.7 but this leads to failing pip audit failing, I have temporarily disabled the pi p audit for just just instances that were violated because of the aformentioned. As Marcelo suggesting, I will open an issue with flwr because this may be a serious issue.

# Tests Added
- NA


pyproject.toml
```diff
@@ -15,9 +15,9 @@ classifiers = [
 python = ">=3.9.0,<3.11"
 numpy = "^1.24"
 pandas = "^2.0"
-flwr = "1.4.0"
+flwr = "1.7.0"
 opacus = "^1.3.0"
-torch = "^1.12.1"
+torch = ">=2.0.0, !=2.0.1, !=2.1.0"
```
lotif, line 20: This is risky, if torch 3 is launched it can get upgraded by accident. Also, those exceptions seem pretty complex. How about not the following:
```toml
torch = "^2.1.1"
```

jewelltaylor: I originally tried something similar but due to a bug in poetry, I had to resort to a rather convoluted constraint: https://stackoverflow.com/questions/76327419/valueerror-libcublas-so-0-9-not-found-in-the-system-path

But on second glance, upgrading to 2.1.1 or higher should be fine. So I will try what you suggested

lotif: Huh that's a curious problem, I've never seen it before.

pyproject.toml
```diff
@@ -15,9 +15,9 @@ classifiers = [
 python = ">=3.9.0,<3.11"
 numpy = "^1.24"
 pandas = "^2.0"
-flwr = "1.4.0"
+flwr = "1.7.0"
 opacus = "^1.3.0"
-torch = "^1.12.1"
+torch = "2.0.0"
```
emersodb, line 20: Did we want 2.1.1 here rather than 2.0.0?

emersodb (approved):
I think this looks good to me.

pyproject.toml
jewelltaylor: I run into some reproducibility issues when I upgrade to torch 2.1.1 that I ran into. Instead of pinning to ^2.1.1, do you think its okay to use 2.0.0. Thats what I have been able to get working in terms of consistency between my local machine and the github server

jewelltaylor: Hashing out with Marcelo now

PR #122, opened 2024-03-22
Displaying all the smoke tests metrics errors at once

lotif:
# PR Type
Fix

# Short Description

Clickup Ticket(s): NA

Fixing smoke tests metric errors is a pain because they are displayed one by one. Changing it here to display all of them at once to avoid multiple iterations of fixing them.

# Tests Added
NA

### DIFF
--- a/tests/smoke_tests/run_smoke_test.py
+++ b/tests/smoke_tests/run_smoke_test.py
@@ -4,7 +4,7 @@ import json
 import logging
 from enum import Enum
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


emersodb (approved):
I imagine this fix is because of how much you enjoyed fixing smoke test issues on your FedDG-GA PR 😂. Looks good to me.

lotif:
@emersodb lol exactly, I ended that day thinking "there's gotta be a better way of doing this" 😂

PR #124, opened 2024-03-27
AG News Experiments and Basic Client Refactor

yc7z:
# PR Type
[Feature | Fix ]

# Short Description

1. A refactor of the `BasicClient` class to support the model taking in a dictionary of tensors as its input. This is useful, for example, if one wants to train a language model on text data. In this case, in addition to the tensors that represent the tokenized sentence, we may also need to pass in the attention masks.
2. An extension of the label-based samplers to support a dictionary-based dataset.
3. A refactor of the dynamic layer exchange example to use hugging face models and datasets. The new example also shows how to use collate functions to dynamically pad the batches as opposed to padding everything to the maximum length, which is what we did before.
4. The code for the same AG news experiment using the sparse tensor exchanger.
5. Moved the original dynamic layer exchanger example to the research folder and replaced it with a typical example on non i.i.d. CIFAR-10.
6. Some small fixes here and there.

# Tests Added

N/A


tests/clients/test_basic_client.py
```diff
@@ -3,12 +3,15 @@
...
 from unittest.mock import MagicMock
 
+import freezegun
 import torch
 from flwr.common import Scalar
 from freezegun import freeze_time
 
 from fl4health.clients.basic_client import BasicClient
 
+freezegun.configure(extend_ignore_list=["transformers"])  # type: ignore
```
yc7z, line 13: I had to add this to prevent `freeze_time` from failing. More details: 

https://github.com/huggingface/transformers/issues/24545

examples/dynamic_layer_exchange_example/client.py
```diff
@@ -0,0 +1,75 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="FL Client Main")
+    parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
+    args = parser.parse_args()
+
+    DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    data_path = Path(args.dataset_path)
+    client = CifarDynamicLayerClient(data_path, [Accuracy("accuracy")], DEVICE, store_initial_model=True)
+    fl.client.start_client(server_address="0.0.0.0:8080", client=client.to_client())
```
yc7z, line 74: I got a warning flower that `start_numpy_client` is deprecated and will be removed in the future, and using `start_client` while ensuring `client.to_client()` is called whenever `client` is an instance of `NumpyClient` is advised. We should probably replace all our current calls of `start_numpy_client` in this way.

fl4health/utils/load_data.py
```diff
@@ -79,8 +78,8 @@ def load_cifar10_data(
             transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5)),
         ]
     )
-    training_set = CIFAR10(str(data_dir), train=True, download=True, transform=transform)
```
yc7z, line 82: The `CIFAR10` that comes with `torchvision` actually doesn't work with our samplers, hence the changes here.

examples/dynamic_layer_exchange_example/client.py
```diff
@@ -0,0 +1,75 @@
...
+            norm_threshold, exchange_percentage, normalize, select_drift_more
+        )
+
+        layer_selection_function = (
+            layer_selection_function_constructor.select_by_percentage()
+            if filter_by_percentage
+            else layer_selection_function_constructor.select_by_threshold()
+        )
+
+        # The user may call layer_selection_function_constructor.select_by_threshold()
```
emersodb, line 59: I think maybe this comment is outdated given your if statements above?

examples/dynamic_layer_exchange_example/client.py
emersodb: Sounds good. I'll put a ticket in the backlog to do this.

fl4health/clients/basic_client.py
```diff
@@ -24,6 +24,10 @@
 from fl4health.utils.metrics import Metric, MetricManager
 
 T = TypeVar("T")
+BatchType = TypeVar(
```
emersodb, line 27: Is this being used anywhere? I don't see it anywhere on the branch, but I may be missing something?

yc7z: Good catch. This is from a previous version and is not currently used anywhere. I'll remove it.

fl4health/clients/basic_client.py
```diff
@@ -629,10 +652,20 @@ def predict(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[st
...
-        output = self.model(input)
+        if isinstance(input, torch.Tensor):
+            output = self.model(input)
+        elif isinstance(input, dict):
+            # If input is a dictionary, then we unpack it before computing the forward pass.
+            # Note that this assumes the keys of the input match (exactly) the keyword args
+            # of self.model.forward().
+            output = self.model(**input)
+        else:
+            raise TypeError(""""input" must be of type torch.Tensor or Dict[str, torch.Tensor].""")
```
emersodb, line 668: I don't think the triple quotes are necessary here. I think you can do `"\"input\""`, but I'm not 100% sure. Normally triple quotes are reserved for line allowing line breaks into the string.

examples/fedopt_example/client.py
```diff
@@ -68,9 +68,12 @@ def setup_client(self, config: Config) -> None:
                 metric.setup(self.label_encoder)
         super().setup_client(config)
 
-    def predict(self, input: torch.Tensor) -> Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]:
+    def predict(
+        self,
+        input: TorchInputType,
```
emersodb, line 73: If we're going to be moving to this type alias, I'd suggest we have method documentation discussing the type and functionality. That is, this is just a union of `torch.Tensor `and `Dict[str, torch.Tensor]`, since these are supposed to be examples for people to follow and maybe they'll have less intimate knowledge of the library itself.

emersodb: I think the type alias is worthwhile though!

fl4health/clients/basic_client.py
```diff
@@ -387,14 +388,32 @@ def _handle_reporting(
         reporting_dict.update(metric_dict)
         self.wandb_reporter.report_metrics(reporting_dict)
 
-    def train_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[TrainingLosses, Dict[str, torch.Tensor]]:
+    def _move_data_to_device(self, data: TorchInputType) -> TorchInputType:
```
emersodb, line 391: For the new methods here, can we add some docstrings to explain their use?

fl4health/clients/basic_client.py
```diff
@@ -387,14 +388,32 @@ def _handle_reporting(
...
+        if isinstance(data, torch.Tensor):
+            return data.to(self.device)
+        elif isinstance(data, dict):
```
emersodb, line 394: I believe we can be a little more specific with the type of dict here. That is, `Dict[str, torch.Tensor]`

fl4health/clients/basic_client.py
```diff
@@ -387,14 +388,32 @@ def _handle_reporting(
...
+            return {key: value.to(self.device) for key, value in data.items()}
+        else:
+            raise TypeError("data must be of type Tensor or Dict[str, Tensor].")
+
+    def is_empty_batch(self, input: Union[torch.Tensor, Dict[str, torch.Tensor]]) -> bool:
+        if isinstance(input, torch.Tensor):
+            return len(input) == 0
+        elif isinstance(input, dict):
```
emersodb, line 402: Similar to the above, dict can be a bit more specific.

fl4health/clients/basic_client.py
```diff
@@ -387,14 +388,32 @@ def _handle_reporting(
...
+            return all(len(val) == 0 for _, val in input.items())
```
emersodb, line 403: Perhaps we can be a bit more sophisticate here. If the length of the tensors are different sizes we also have a problem. So perhaps we first check that they are all the same size and then return true if that size is zero?

fl4health/clients/basic_client.py
emersodb, line 391: Since this method is restricted to `TorchInputType`, perhaps we call this function, `_move_input_data_to_device` to be specific

fl4health/strategies/fedavg_dynamic_layer.py
```diff
@@ -49,6 +51,8 @@ def __init__(
         Args:
             fraction_fit (float, optional): Fraction of clients used during training. Defaults to 1.0. Defaults to 1.0.
             fraction_evaluate (float, optional): Fraction of clients used during validation. Defaults to 1.0.
+            min_fit_clients (int, optional): _description_. Defaults to 2.
```
emersodb, line 54: __description__ here probably needs to be filled in 🙂

fl4health/utils/dataset.py
```diff
@@ -53,3 +53,31 @@ def __init__(
...
+    def __init__(
+        self,
+        data_path: Path,
+        train: bool,
+        transform: Union[None, Callable] = None,
+        target_transform: Union[None, Callable] = None,
+    ):
+        cifar10_dataset = CIFAR10(data_path, train=train, download=True)
+        self.data = torch.from_numpy(cifar10_dataset.data)
+        self.targets = torch.Tensor(cifar10_dataset.targets).long()
```
emersodb, line 68: Interesting that this long conversion is necessary for Cifar but not Mnist 🤔

fl4health/clients/basic_client.py
yc7z: `isinstance(data, Dict[str, torch.Tensor])` actually doesn't work because `isinstance` can't take in Generic types. I know using `dict` is not ideal but I don't know of a better way...

emersodb: Right...I forgot about that part. Using dict here is sufficient then

examples/fedopt_example/client.py
yc7z: Good suggestion!

fl4health/clients/basic_client.py
yc7z: In the case of attention masks, it makes sense that all tensors in the dictionary should have the same length. But in general, could there be a case when we might want to pass in multiple arguments of different sizes into a model?

emersodb: hmmm I suppose that's possible, but we're essentially checking on batch size here and I think we'd want there to be information per batch element (even if it's duplicated). Maybe that's more strict than we want, but for now I'd vote we enforce equality here and loosen it if we need to later on?

research/ag_news/client_data.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def collate_fn_with_padding(
+    tokenizer: PreTrainedTokenizer, batch: List[Tuple[Dict[str, List[torch.Tensor]], torch.Tensor]]
+) -> Tuple[Dict[str, List[torch.Tensor]], torch.Tensor]:
+    """
+    Pad the sequences within a batch to the same length.
+    """
+    input_list, target_list = [], []
+    for pair in batch:
```
emersodb, line 22: Rather than indexing into pair here, it's a little more readable to unpack them in the for with something like 
```
for input_dict, target in batch:
....
```

research/ag_news/client_data.py
```diff
@@ -0,0 +1,80 @@
...
+
+    val_dataset = tokenized_ag_news["test"]
+    val_dataset.set_format(type="torch", columns=all_column_names)
+    val_dataset = create_text_classification_dataset(val_dataset, data_column_names, "label")
+    val_dataset = sampler.subsample(val_dataset)
+
+    # The collate function is used to dynamically pad the sequences within every batch to the same length.
+    # This is more efficient than padding all sequences to the maximum length.
+    train_loader = DataLoader(
+        train_dataset, batch_size=batch_size, shuffle=True, collate_fn=partial(collate_fn_with_padding, bert_tokenizer)
```
emersodb, line 75: Very nicely done

research/ag_news/dynamic_layer_exchange/README.md
```diff
@@ -0,0 +1,74 @@
...
+You can customize how tensors are selected and how many tensors to exchange, but more on this later.
+
+This example fine-tunes a pre-trained BERT encoder to perform text classification on the AG News dataset in a federated setting.
+Each client performs Dirichlet sampling on its training set so we get heterogeneous data distributions across clients.
+
+## Running the Example
+In order to run the example, first ensure you have [installed the dependencies in your virtual environment according to the main README](/README.md#development-requirements) and it has been activated.
+
+## Information about Dataset
+client_data.py handles all data preprosessiong steps and is responsible for constructing the dataloaders for training and evaluation.
```
emersodb, line 16: The spelling of preprocessing appears to have gone a little sideways here 🙂

research/ag_news/sparse_tensor_exchange/run_hp_sweep.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+###############################################
+# Usage:
+#
+#  ./research/ag_news/dynamic_layer_exchange/run_hp_sweep.sh \
```
emersodb, line 6: I think you want to update the various example paths in this file to `sparse_tensor_exchange` instead of `dynamic_layer_exchange`

research/ag_news/sparse_tensor_exchange/server.py
```diff
@@ -0,0 +1,115 @@
...
+        # Server waits for min_available_clients before starting FL rounds
+        min_available_clients=config["n_clients"],
+        on_fit_config_fn=fit_config_fn,
+        on_evaluate_config_fn=fit_config_fn,
+        fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
+        evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        initial_parameters=get_all_model_parameters(initial_model),
+    )
+
+    client_manager = PoissonSamplingClientManager()
```
emersodb, line 83: Given that we're not subsampling the clients, I don't think this matters, but you can likely just use a `SimpleClientManager` here.

research/ag_news/dynamic_layer_exchange/server.py
```diff
@@ -64,87 +58,38 @@ def fit_config(
...
     # Since clients can send different tensors to the server, we perform weighted averaging separately on each tensor.
     strategy = FedAvgDynamicLayer(
         # Server waits for min_available_clients before starting FL rounds
         min_available_clients=config["n_clients"],
         on_fit_config_fn=fit_config_fn,
-        on_evaluate_config_fn=eval_config_fn,
+        on_evaluate_config_fn=fit_config_fn,
         fit_metrics_aggregation_fn=fit_metrics_aggregation_fn,
         evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
         initial_parameters=get_all_model_parameters(initial_model),
```
emersodb, line 95: Not a big deal, but similar to the sparse tensor exchanger, I think you can just use a `SimpleClientManager` here instead of a `PoissonSamplingClientManager`

tests/clients/test_basic_client.py
emersodb: Interesting 🤔

fl4health/clients/basic_client.py
yc7z: Sounds good to me!

fl4health/utils/dataset.py
yc7z: Actually the long conversion is not necessary, but unlike Mnist, we do need to convert the targets to tensors for Cifar.

research/ag_news/dynamic_layer_exchange/README.md
yc7z: 😂

research/ag_news/sparse_tensor_exchange/server.py
yc7z: Yes good point

fl4health/utils/dataset.py
yc7z: My bad... Turns out we do need to convert it to long

fl4health/clients/ensemble_client.py
```diff
@@ -82,13 +84,16 @@ def train_step(self, input: torch.Tensor, target: torch.Tensor) -> Tuple[Trainin
...
+            input (TorchInputType): The input to be fed into the model.
+            TorchInputType is simply an alias for the union of torch.Tensor and
+            Dict[str, torch.Tensor].
             target (torch.Tensor): The target corresponding to the input.
 
         Returns:
             Tuple[TrainingLosses, Dict[str, torch.Tensor]]: The losses object from the train step along with
                 a dictionary of any predictions produced by the model.
         """
+        assert isinstance(input, torch.Tensor)
```
emersodb, line 96: Any reason that we have to do this asset here? The basic_client predict method should be able to handle a `Dict[str, torch.Tensor]` I think?

fl4health/clients/fenda_client.py
```diff
@@ -64,19 +64,21 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
+            input (TorchInputType): Inputs to be fed into the model. TorchInputType is simply an alias
+            for the union of torch.Tensor and Dict[str, torch.Tensor].
 
         Returns:
             Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
             contains predictions indexed by name and the second element contains intermediate activations
             index by name. Specificaly the features of the model, features of the global model and features of
             the old model are returned. All predictions included in dictionary will be used to compute metrics.
         """
+        assert isinstance(input, torch.Tensor)
```
emersodb, line 81: Similar to the ensemble, do we need this assertion?

fl4health/clients/moon_client.py
```diff
@@ -45,19 +45,21 @@ def __init__(
...
+            input (TorchInputType): Inputs to be fed into the model. TorchInputType is simply an alias
+            for the union of torch.Tensor and Dict[str, torch.Tensor].
 
         Returns:
             Tuple[Dict[str, torch.Tensor], Dict[str, torch.Tensor]]: A tuple in which the first element
             contains predictions indexed by name and the second element contains intermediate activations
             index by name. Specifically the features of the model, features of the global model and features of
             the old model are returned. All predictions included in dictionary will be used to compute metrics.
         """
+        assert isinstance(input, torch.Tensor)
```
emersodb, line 62: Similar to the ditto predict, perhaps we can do `self.model(**input)`?

fl4health/clients/ensemble_client.py
yc7z: I only did this assert because the `forward()` method of `EnsembleModel` only takes in `torch.Tensor` as its input, and the model of `ensemble_client` has type `EnsembleModel`. For the same reason I also added an assert for Fenda and Moon. Do you think it would be better to do `self.model(**input)` instead?

fl4health/clients/fenda_client.py
yc7z: See https://github.com/VectorInstitute/FL4Health/pull/124#discussion_r1566168462

yc7z: Also just for clarity, I did `self.model(**input)` for ditto because the model there has a general `nn.Module` type, unlike the ensemble/moon/fenda clients.

fl4health/clients/ensemble_client.py
emersodb: Ah I see. No, let's leave that for now. We can refactor when necessary.

emersodb (approved):
Good to go from me

PR #125, opened 2024-04-09
Add MR-MTL Method

sanaAyrml:
# PR Type
[Feature]

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/86884xap4)

This is the implementation of  [MR-MTL: On Privacy and Personalization in Cross-Silo Federated Learning](https://arxiv.org/pdf/2206.07902.pdf). 
The method is very related to FedProx and Ditto. Essentially, at the start of each client training round, we do not update local model weights directly with the global model weights, but instead we constrain the model weights to be close to the initial global model weights at the start of each client training round. Initial global model weights are computed by averaging the model weights at the end of the previous round. Such a mean-regularized training is done by adding a penalty term to the loss function that constrains the local model weights to be close to the initial global model weights.

# Tests Added

Three tests are added to `tests/clients/test_mr_mtl_client.py` regarding setting global weights (as we don't set them for local model but just save them for computing mr loss), forming mr loss and computing loss.


examples/ditto_example/README.md
```diff
@@ -1,4 +1,4 @@
-# FedProx Federated Learning Example
+# Ditto Federated Learning Example
```
sanaAyrml, line 1: I think this left from copy past so I fixed it.

examples/mr_mtl_example/client.py
```diff
@@ -0,0 +1,72 @@
...
+    DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    data_path = Path(args.dataset_path)
+    log(INFO, f"Device to be used: {DEVICE}")
+    log(INFO, f"Server Address: {args.server_address}")
+
+    # Set the random seed for reproducibility
+    set_all_random_seeds(args.seed)
+
+    client = MnistMrMtlClient(data_path, [Accuracy()], DEVICE, lam=0.1)
+    fl.client.start_numpy_client(server_address=args.server_address, client=client)
```
emersodb, line 67: Apparently, this `start_numpy_client` method has been deprecated in the new Flower versions (will be removed in later versions). I have a ticket to migrate our existing examples away from this, but for net new examples, lets use 
```fl.client.start_client(server_address=args.server_address, client=client.to_client())```
as advised by flower

File: examples/mr_mtl_example/client.py [hunk 3]
```diff
@@ -0,0 +1,72 @@
...
+
+    # Shutdown the client gracefully
+    client.shutdown()
+
+    client.metrics_reporter.dump()
```
emersodb, line 72: @lotif: Do you think we should just add this `client.metrics_report.dump()` to the shutdown so that it just always happens?

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+from flwr.common.typing import Config, NDArrays, Scalar
+
+from fl4health.checkpointing.checkpointer import TorchCheckpointer
+from fl4health.clients.basic_client import BasicClient
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.losses import LossMeterType, TrainingLosses
+from fl4health.utils.metrics import Metric
+
+
+class MrMtlClient(BasicClient):
```
emersodb, line 17: This is the correct capitalization scheme I think, but it keeps making me think we are writing "Mr. Montreal" in shorthand 😂

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self,
+        data_path: Path,
+        metrics: Sequence[Metric],
+        device: torch.device,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        checkpointer: Optional[TorchCheckpointer] = None,
+        lam: float = 1.0,
```
emersodb, line 25: This isn't part of the original implementation (and shouldn't be in this PR), but it would be interesting to see what effect adapting this value based on the adaptive FedProx approach or the generalization gap of FedDG-GA would have on the approach. A similar idea could be said for the Ditto parameter.

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+    ) -> None:
+        """
+        This client implements the MR-MTL algorithm from MR-MTL: On Privacy and Personalization in Cross-Silo
+        Federated Learning. The idea is that we want to train personalized versions of the global model for each
+        client. However despite Ditto we don't solve a separate solver for global model. We update initial global model
```
emersodb, line 30: Perhaps ```However, instead of using a separate solver for the global model, as in Ditto, we update the initial global model...```?

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+    def get_parameters(self, config: Config) -> NDArrays:
+        """
+        For MR-MTL, we transfer the LOCAL model weights to the server to be aggregated and set as INITIAL GLOBAL model
+        weightson client side.
+
+        Args:
+            config (Config): The config is sent by the FL server to allow for customization in the function if desired.
+
+        Returns:
+            NDArrays: GLOBAL model weights to be sent to the server for aggregation
```
emersodb, line 106: I think we want to change `GLOBAL` to `LOCAL` here?

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+        return self.parameter_exchanger.push_parameters(self.model, config=config)
+
+    def set_parameters(self, parameters: NDArrays, config: Config, fitting_round: bool) -> None:
+        """
+        The parameters being pass are to be routed to the initial global model to be used in a penalty term in
+        training the local model.
+
+        Args:
+            parameters (NDArrays): Parameters have information about model state to be added to the relevant client
+                model (global model for all but the first step of MR-MTL)
```
emersodb, line 118: So it looks like, based on the implementation below, we actually only ever set the global model (even if its the first step). Based on algorithm A.1 from the paper, I think that's fine. That is, each client starts with a random initial model but the same global model. However, we should talk about that in the description for this method if that's how we're doing it. That is, make that behavior explicit in the comments.

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+        assert self.init_global_model is not None and self.model is not None
+        assert self.parameter_exchanger is not None and isinstance(self.parameter_exchanger, FullParameterExchanger)
+
+        # Route the parameters to the GLOBAL model in MR-MTL
+        assert self.parameter_exchanger is not None
+        log(INFO, "Setting the global model weights")
+        self.parameter_exchanger.pull_parameters(parameters, self.init_global_model, config)
+
+    def update_before_train(self, current_server_round: int) -> None:
+        assert isinstance(self.model, nn.Module)
```
emersodb, line 136: Based on the fact that you're manipulating `self.init_global_model` below, I think this assert is meant to be on `self.init_global_model` instead?

fl4health/clients/mr_mtl_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns:
+            TrainingLosses: an instance of TrainingLosses containing backward loss and
+                additional losses indexed by name. Additional losses includes each loss component of the total loss.
+        """
+        # Check that both models are in training mode
+        assert not self.init_global_model.training and self.model.training
+
+        total_loss, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
+        if additional_losses is None:
```
emersodb, line 170: TrainingLosses actually handles the setting where `additional_losses` is None. So no need to do this step 🙂

tests/clients/test_mr_mtl_client.py
```diff
@@ -0,0 +1,79 @@
...
+
+    params = [val.cpu().numpy() + 0.1 for _, val in mr_mtl_client.model.state_dict().items()]
+    mr_mtl_client.set_parameters(params, config, fitting_round=True)
+
+    mr_loss = mr_mtl_client.get_mr_drift_loss()
+
+    assert pytest.approx(mr_loss.detach().item(), abs=0.02) == (mr_mtl_client.lam / 2.0) * (
+        1.5 + 0.06 + 24.0 + 81.92 + 0.16 + 0.32
+    )
+
```
emersodb, line 51: Perhaps we can add somewhere in these tests a sanity check after explicitly calling `update_before_train` that the global model is in eval mode and that it's parameters `require_grad` are all false, but the self.model remains in train mode and has `requires_grad=True` throughout.

fl4health/clients/mr_mtl_client.py
sanaAyrml: Lol 😂

fl4health/clients/mr_mtl_client.py
sanaAyrml: I am confused by this comment, which part do you mean exactly?

fl4health/clients/mr_mtl_client.py
sanaAyrml: Correct, it should be local

fl4health/clients/mr_mtl_client.py
sanaAyrml: Good idea, I added a comment about this to the description.

fl4health/clients/mr_mtl_client.py
emersodb: Apologies, this was just me recording a thought about something to investigate in the future. That is, would MR-MTL or Ditto benefit from an adaptive implementation of the FedProx-like parameter $\lambda$. It could be adjusted as they suggested in the original paper, or perhaps we could adjust it based on a similar measure to that proposed in FedDG-GA (which is a generalization gap).

TL;DR: No need to do anything. Was just thinking about potential future experimentation.

PR #126, opened 2024-04-18
Updating the start numpy client code.

emersodb:
**Don't be too scared of the number of files changed. The changes are very small.  🙂 **

# PR Type
Other

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/868821mk3

This PR aims to do two things. The first is simply to migrate the clients to use the recommended `start_client` functionality instead of `start_numpy_client` which has now been deprecated. This also requires a call of `client.to_client()` when doing so.

The second is just to address some spelling issues in comments and code throughout the repository. These changes are all, hopefully very minor and do not affect functionality in any way.

# Tests Added

N/A the changes are covered by existing tests.


fl4health/feature_alignment/tabular_feature.py
```diff
@@ -10,7 +10,11 @@
 
 class TabularFeature:
     def __init__(
-        self, feature_name: str, feature_type: TabularType, fill_value: Optional[Scalar], metadata: MetaData = []
+        self,
+        feature_name: str,
+        feature_type: TabularType,
+        fill_value: Optional[Scalar],
+        metadata: Optional[MetaData] = None,
```
emersodb, line 17: Not that this is to avoid a common ["gotcha" in python for mutable collection defaults](https://docs.python-guide.org/writing/gotchas/). We just default this to None instead of an empty list and set it explicitly in the constructor. Python is dumb sometimes...

fatemetkl (approved):
LGTM!

.github/workflows/static_code_checks.yaml
```diff
@@ -48,3 +48,4 @@ jobs:
             GHSA-3ww4-gg4f-jr7f
             GHSA-9v9h-cgj8-h64p
             GHSA-6vqw-3v5j-54x4
+            GHSA-7gpw-8wmc-pm8g
```
lotif, line 51: Looks like we can just upgrade `aiohttp` to latest and remove this from the ignore list.

```
poetry update aiohttp
```

https://github.com/advisories/GHSA-7gpw-8wmc-pm8g

lotif (commented):
That's a massive grammar and typo fix 😂 :clap:

fl4health/feature_alignment/tabular_feature.py
yc7z: Great catch! I didn't know about this at all 😂

.github/workflows/static_code_checks.yaml
emersodb: Yeah will do that before merging!

yc7z (approved):

lotif (approved):

PR #128, opened 2024-04-23
Adding the ability to specify a checkpointing for models **before** aggregation.

emersodb:
**Note**: There are a scary number of files changed because this affects the functionality of BasicClient in a way that requires slight modifications of many of the inheriting classes. Most of the changes should be quite minor in those downstream classes.

# PR Type
Feature

Clickup Ticket(s): 
1. https://app.clickup.com/t/86880z10x
2. https://app.clickup.com/t/860r0e3a0

Refactoring the client-side checkpointing functionality to allow for pre- and post-aggregation (server-side) checkpointing. That is, the user can specify a checkpointer for models after the weights have been aggregated by the server (still supported, but was the only default option), or prior to this happening (i.e. right after training). If either of the checkpointers are not specified, then checkpointing at that time is skipped. Note, this allows us to mimic "fine-tuning" of models. That is, the models final training is exclusively local. 

For example, with FedProx, the model checkpointed would actually be a "personal" model for each client (potentially after several preceding rounds of aggregation).

Also allowing for more generic checkpointing functionality. That is, given a loss value and a dictionary of metrics, users can define an arbitrary scoring function on those objects to produce a checkpoint. The best loss checkpointer is a specific instantiation of this type of checkpointer.

# Tests Added

Added some new tests to cover the new ClientSideCheckpointModule functionality and also the new functionality associated with the TorchCheckpointer child classes.


fl4health/checkpointing/checkpointer.py
```diff
@@ -2,71 +2,157 @@
...
+        Args:
+            checkpoint_dir (str): Directory to which the model is saved. This directory should already exist. The
+                checkpointer will not create it if it does not.
+            checkpoint_name (str): Name of the checkpoint to be saved.
+        """
+        super().__init__(
+            checkpoint_dir, checkpoint_name, checkpoint_score_function=loss_score_function, maximize=False
+        )
+
+    def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: Dict[str, Scalar]) -> None:
```
emersodb, line 140: Note, I'm overriding this method to replace the logging with something more specific. If anyone has any better ideas on how to do this, let me know.

fl4health/clients/basic_client.py
```diff
@@ -631,8 +657,6 @@ def validate(self) -> Tuple[float, Dict[str, Scalar]]:
         metrics = self.val_metric_manager.compute()
         self._handle_logging(loss_dict, metrics, is_validation=True)
 
-        # Checkpoint based on loss which is output of user defined compute_loss method
-        self._maybe_checkpoint(loss_dict["checkpoint"])
```
emersodb, line 635: Note that I moved this into the evaluate function rather than the validate function, as we may not want to checkpoint every time we validate etc.

fatemetkl (approved):
LGTM! Checkpoint score functions, and pre-aggregation vs post-aggregation checkpointing are nice additions towards flexibility.

PR #129, opened 2024-04-24
Creating an Opacus specific Checkpointer

emersodb:
# PR Type
Feature/Fix

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/8688075u1

This PR addresses a checkpointing issue that may arise when using Opacus `make_private` functionality. When using Opacus for dp-sgd training, the library wraps the `nn.Module` objects as `GradSampleModule`s. In the ideal case, nothing about checkpointing changes for this operation. However, checkpointing via `torch.save`, which uses pickle dump under the hood can fail depending on how Opacus decides to perform the wrapping. For complex models, the odds of this happening increases (i.e. FedISIC EfficientNet models), but it can be induced even for simple layers by forcing the wrapping via "functorch" as done in the tests. To address this, rather than using torch.save, the Opacus checkpointers simply pickle the state dictionaries and load them via `load_state_dict`. 

Using these checkpointers is demonstrated in the instance level dp example.

# Tests Added

A number of tests have been implemented in the tests folder for the new checkpointers.


fl4health/checkpointing/opacus_checkpointer.py
```diff
@@ -0,0 +1,167 @@
...
+        instead of using the torch.save workflow.
+
+        Args:
+            model (nn.Module): Model to be potentially saved (should be an Opacus wrapped model)
+            loss (float): Loss value associated with the model to be used in checkpointing decisions.
+            metrics (Dict[str, Scalar]): Metrics associated with the model to be used in checkpointing decisions.
+        """
+        assert isinstance(
+            model, GradSampleModule
+        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
```
emersodb, line 32: Used to force the use of this checkpointer only for Opacus based models.

fl4health/utils/functions.py
```diff
@@ -1,36 +1,8 @@
...
-from flwr.common.logger import log
 from flwr.common.parameter import ndarrays_to_parameters
 from flwr.common.typing import Parameters
-from opacus.validators import ModuleValidator
 
 
 def get_all_model_parameters(model: nn.Module) -> Parameters:
     # Extracting all model parameters and converting to Parameters object
     return ndarrays_to_parameters([val.cpu().numpy() for _, val in model.state_dict().items()])
-
```
emersodb, line 14: Moved this functionality into its own file (fl4health/utils/privacy_utilities.py)

examples/dp_fed_examples/instance_level_dp/client.py
```diff
@@ -37,10 +41,18 @@ def get_criterion(self, config: Config) -> _Loss:
     parser.add_argument("--dataset_path", action="store", type=str, help="Path to the local dataset")
     args = parser.parse_args()
 
+    client_name = "".join(choices(string.ascii_uppercase, k=5))
```
lotif, line 44: Interesting way to generate a random ID!

fl4health/checkpointing/opacus_checkpointer.py
lotif: There is no way of changing the type in the method signature?

fl4health/checkpointing/opacus_checkpointer.py
```diff
@@ -0,0 +1,167 @@
...
+            log(
+                INFO,
+                f"Not saving Opacus model state: Current score ({comparison_score}) is not "
+                f"{self.comparison_str} Best score ({self.best_score})",
+            )
+
+    def _process_state_dict_keys(self, opacus_state_dict: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        # State dictionary keys for Opacus modules will be prefixed with an _module. So we remove these when loading
+        # the state information into a standard torch model.
```
lotif, line 53: nit: unnecessary `#` at the beginning of those lines.

fl4health/checkpointing/opacus_checkpointer.py
```diff
@@ -0,0 +1,167 @@
...
+            model_state_dict = pickle.load(handle)
+            # If the target is just a plain PyTorch module, we remove the _module key prefix that Opacus inserts into
+            # its GradSampleModules.
+            if not target_is_grad_sample_module:
+                model_state_dict = self._process_state_dict_keys(model_state_dict)
+            target_model.load_state_dict(model_state_dict, strict=True)
+
+
+def latest_score_function(loss: float, _: Dict[str, Scalar]) -> float:
+    return 0.0
```
lotif, line 105: What is the point of this function?

fl4health/checkpointing/opacus_checkpointer.py
emersodb: You are correct, we actually can change the signature here. So I did that. However, I still want to only allow this checkpointer to be used for Opacus models. If the library user isn't using mypy, they'll still be able to call maybe_checkpoint. So I left the assertions in there, but changed the signature.

lotif (approved):

PR #131, opened 2024-04-27
Consolidate Weight Drift Losses

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link(s) if applicable.](https://app.clickup.com/t/86886ggzv)

The goal for this ticket is to consolidate the weight drift losses that were identical and duplicated within the FedProx, Ditto, and MR-MTL clients into a single loss function that they could all use. This involved a bit of refactoring of the tests. I also fixed a few silly test choices that I made in the FedProx tests. The behavior was working as expected, but the tests weren't doing what I wanted them to do.

# Tests Added

Updated the tests for each of the clients to work with the new consolidated weight drift loss function.


fl4health/clients/ditto_client.py
```diff
@@ -153,12 +132,15 @@ def set_parameters(self, parameters: NDArrays, config: Config, fitting_round: bo
             log(INFO, "Setting the global model weights")
             self.parameter_exchanger.pull_parameters(parameters, self.global_model, config)
 
+    def update_before_train(self, current_server_round: int) -> None:
```
emersodb, line 135: Moving the setting of the initial_global_tensors to the `update_before_train` function so that it isn't done during validation. It wasn't causing any issues, but it's unnecessary and this function conveniently isolates both the `set_parameters` functionality and  ensures that the tensors are saved only once before local training starts

fl4health/clients/fed_prox_client.py
```diff
@@ -93,12 +80,15 @@ def set_parameters(self, parameters: NDArrays, config: Config, fitting_round: bo
 
         super().set_parameters(server_model_state, config, fitting_round)
 
+    def update_before_train(self, current_server_round: int) -> None:
```
emersodb, line 83: Moving the setting of the `initial_tensors` to the `update_before_train` function so that it isn't done during validation. It wasn't causing any issues, but it's unnecessary and this function conveniently isolates both the `set_parameters` functionality and  ensures that the tensors are saved only once before local training starts

fl4health/clients/mr_mtl_client.py
```diff
@@ -113,33 +90,38 @@ def set_parameters(self, parameters: NDArrays, config: Config, fitting_round: bo
...
             fitting_round (bool): Boolean that indicates whether the current federated learning
-                round is a fitting round or an evaluation round.
-                This is used to help determine which parameter exchange should be used for pulling parameters.
-                A full parameter exchanger is only used if the current federated learning round is the very
-                first fitting round.
+                round is a fitting round or an evaluation round. Not used here.
         """
         # Make sure that the proper components exist.
-        assert self.init_global_model is not None and self.model is not None
+        assert self.initial_global_model is not None
```
emersodb, line 104: We don't use self.model in this function. So I removed the check.

fl4health/losses/weight_drift_loss.py
```diff
@@ -0,0 +1,34 @@
...
+        self,
+        device: torch.device,
+    ) -> None:
+        super().__init__()
+        self.device = device
+
+    def _compute_weight_difference_inner_product(self, x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+        return torch.pow(torch.linalg.norm(x - y), 2.0)
+
+    def forward(self, target_model: nn.Module, constraint_tensors: List[torch.Tensor], weight: float) -> torch.Tensor:
```
emersodb, line 18: Note, the weight here has to be provided in the forward to be compatible with FedProx adaptive weight functionality. That is, setting the weight in the __init__ wouldn't work the way we wanted because FedProx (or other methods), might modify it.

sanaAyrml (approved):

PR #132, opened 2024-05-03
Enable pre-commit ci and add default config

amrit110:
# PR Type
CI

# Short Description
- Adds default config for using pre-commit ci (https://pre-commit.ci/).
- If this PR is approved, I will give permissions to the CI to submit PRs for fixing pre-commit errors and also updating the hooks

# Tests Added


amrit110:
Currently failing due to pip-audut check. Posted an update which would fix that #133

emersodb (approved):
I think this is good with me, but worth @lotif also double checking to see if there are any implications I might be missing.

.pre-commit-config.yaml
```diff
@@ -45,3 +45,15 @@ repos:
...
+ci:
+    autofix_commit_msg: |
+        [pre-commit.ci] Add auto fixes from pre-commit.com hooks
+
+        for more information, see https://pre-commit.ci
+    autofix_prs: true
+    autoupdate_branch: ''
+    autoupdate_commit_msg: '[pre-commit.ci] pre-commit autoupdate'
+    autoupdate_schedule: weekly
+    skip: [mypy]
```
lotif, line 58: Should we add tests to the skip list too? I think the goal of this is to perform the auto-corrections from black, isort and such, right? Please correct me if I'm wrong.

We should definitely skip the smoke tests, though, or this will take a long time to finish (and also because of that flakiness that is still happening sometimes, albeit a bit rarely).

lotif: Oh I see, there are no tests in the pre-commit for this repo. Approving :)

lotif (approved):

.pre-commit-config.yaml
emersodb: Do you think we should have tests in the pre-commit (would be a different PR)? I didn't include them, because I didn't necessarily want to preclude people from intermediate (but test breaking) commits. However, I could be convinced that was not a good idea 😂.

lotif: I think it's a good idea if the tests are fast to run. The Florist unit tests are super fast (maybe because there are not that many yet) so I included them, but left the integration tests out.

PR #147, opened 2024-05-09
Address mypy issues for DPScaffoldClient

FreddieNeverLeft:
# PR Type
Fix

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8685apf16)

Referenced in this comment by David: https://github.com/VectorInstitute/FL4Health/pull/48#discussion_r1293855966
Mypy appears to have some trouble sorting out typing when the client inherits from multiple base classes with complex functionality. 

Side-stepped this issue by ScaffoldClient initializing using a generic torch optimizer and then asserting it to SGD during setup_client()

# Tests Added

N/A.


jewelltaylor (approved):
LGTM! It is definitely better than just ignoring type checking for this class

PR #151, opened 2024-05-15
part 1 changes for adding training set capabilities for client side evaluation 

FreddieNeverLeft:
# PR Type
**[Feature]**

# Short Description

[Clickup Ticket(s): Link](https://app.clickup.com/t/8685pr77r)
[Clickup Subticket: Link](https://app.clickup.com/t/8688fby08)

As detailed in the ticket's subtasks, this branch will focus on subtask1:

Task 1: add test dataset naively, assuming the user will ensure the val and test set are proportionally distributed to each client

# Tests Added
updated test_metrics_reporter_evaluate() in test_basic_client.py, for test dataset metrics.


fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
...
+    ) -> Tuple[float, Dict[str, Scalar]]:
         """
-        Validate the current model on the entire validation dataset.
+        Evaluate the model on the given validation or test dataset.
+
+        Args:
+            loader (DataLoader): The data loader for the dataset (validation or test).
+            loss_meter (LossMeter): The meter to track the losses.
+            metric_manager (MetricManager): The manager to track the metrics.
+            is_validation (bool): Whether this evaluation is for validation or test. Default is True for validation.
```
emersodb, line 661: If we create the LoggingMode enum, we can pass that here as well and it will simplify a bit of the code below.

fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
...
-            for input, target in self.val_loader:
+            for input, target in loader:
                 input, target = self._move_input_data_to_device(input), target.to(self.device)
                 losses, preds = self.val_step(input, target)
-                self.val_loss_meter.update(losses)
-                self.val_metric_manager.update(preds, target)
+                loss_meter.update(losses)
+                metric_manager.update(preds, target)
 
-        # Compute losses and metrics over validation set
```
emersodb, line 655: Very minor, but we lost this comment in the refactor. Maybe just bring it over?

fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
...
+        if is_validation:
+            self._handle_logging(loss_dict, metrics, is_validation=True)
+        else:
+            self._handle_logging(loss_dict, metrics, is_testing=True)
 
         return loss_dict["checkpoint"], metrics
 
+    def validate(self) -> Tuple[float, Dict[str, Scalar]]:
+        """
+        Validate the current model on the entire validation or test dataset.
```
emersodb, line 687: Very minor, but we're evaluating the current model over the entire validation set _and potentially an entire test dataset if it has been defined_

fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
...
+
+        Returns:
+            Tuple[float, Dict[str, Scalar]]: The validation loss and a dictionary of metrics
+                from validation (and test if present).
+        """
+        val_loss, val_metrics = self._val_or_test(self.val_loader, self.val_loss_meter, self.val_metric_manager)
+        if hasattr(self, "test_loader") and self.test_loader:
```
emersodb, line 694: Since this is the base class for our clients, I don't think we need to test whether the attribute exists. However, perhaps I'm missing something that you saw that required this check?

fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
...
+            test_loss, test_metrics = self._val_or_test(
+                self.test_loader, self.test_loss_meter, self.test_metric_manager, is_validation=False
+            )
+            val_metrics["test - loss"] = test_loss
+            val_metrics.update(test_metrics)
```
emersodb, line 699: A comment here would be useful to explain why there won't be any risk of keys clashing here (due to the naming convention associated with the metric managers.

fl4health/clients/basic_client.py
```diff
@@ -874,6 +917,20 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, ...]:
         """
         raise NotImplementedError
 
+    def get_test_data_loader(self, config: Config) -> Optional[DataLoader]:
+        """
+        User defined method that returns a PyTorch Test DataLoader
```
emersodb, line 922: I would expand on this comment a bit to say that by default, this function returns None, assuming that there is no test dataset to be used. If the user would like to load and evaluate a dataset, they need only override this function in their client class.

fl4health/clients/basic_client.py
```diff
@@ -874,6 +917,20 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, ...]:
...
+
+        Args:
+            config (Config): The config from the server.
+
+        Returns:
+            Optional[Tuple[DataLoader, ...]]: Tuple of length 2. The optinal client test loader.
```
emersodb, line 928: I presume the type and description here is a copy paste error? Also optional is spelled wrong :-)

fl4health/clients/basic_client.py
```diff
@@ -874,6 +917,20 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, ...]:
...
+                Defaults to None.
```
emersodb, line 929: It doesn't technically "default" to none. The base implementation always returns none

tests/clients/test_basic_client.py
```diff
@@ -72,14 +73,16 @@ def test_metrics_reporter_evaluate() -> None:
...
+                "evaluate_metrics": test_metrics_final,
             },
         },
     }
 
 
 def test_evaluate_after_fit_enabled() -> None:
     fl_client = MockBasicClient()
+    fl_client.validate = MagicMock()  # type: ignore
+    fl_client.validate.return_value = fl_client.mock_loss, fl_client.mock_metrics
```
emersodb, line 85: Are these values actually required, since you're not asserting that they are equal to anything?

tests/clients/test_basic_client.py
```diff
@@ -88,6 +91,8 @@ def test_evaluate_after_fit_enabled() -> None:
 
 def test_evaluate_after_fit_disabled() -> None:
     fl_client = MockBasicClient()
+    fl_client.validate = MagicMock()  # type: ignore
+    fl_client.validate.return_value = fl_client.mock_loss, fl_client.mock_metrics
```
emersodb, line 95: Are these values actually required for this test, since you're not asserting that they are equal to anything?

tests/clients/test_basic_client.py
```diff
@@ -61,6 +61,7 @@ def test_metrics_reporter_evaluate() -> None:
     test_current_server_round = 2
     test_loss = 123.123
     test_metrics: Dict[str, Union[bool, bytes, float, int, str]] = {"test_metric": 1234}
+    test_metrics_final: Dict[str, Union[bool, bytes, float, int, str]] = {"test_metric": 1234, "test - loss": 123.123}
```
emersodb, line 64: Two small recommendations:
(1) I think you can just redefine `test_metrics` here to be your new dictionary.
(2) I know you didn't write this, but I think that Union[bool, bytes, float, int, str] is equivalent to the Flower Scalar type. So you can just use that rather than the long union.

tests/clients/test_basic_client.py
emersodb, line 1: I'd recommend adding an explicit test of the new functionality of validate where you actually test that the merge of the two metrics dictionaries flows appropriately from the definitions of the metrics managers. You can use something like the mocks in this file and/or the fake dataloaders of `tests/clients/test_setup_client_basic.py`

emersodb (commented):
Overall changes look pretty good. Added some fairly small comments to make the code slightly cleaner etc.

fl4health/clients/basic_client.py
```diff
@@ -635,30 +644,62 @@ def train_by_steps(
 
         return loss_dict, metrics
 
-    def validate(self) -> Tuple[float, Dict[str, Scalar]]:
+    def _val_or_test(
```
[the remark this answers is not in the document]
FreddieNeverLeft: Thanks for the suggestion! Changed this in the latest commit.

fl4health/clients/basic_client.py
```diff
@@ -320,15 +323,15 @@ def evaluate_after_fit(self) -> Tuple[float, Dict[str, Scalar]]:
 
     def evaluate(self, parameters: NDArrays, config: Config) -> Tuple[float, int, Dict[str, Scalar]]:
         """
-        Evaluates the model on the validation set.
+        Evaluates the model on the validation set, and test set (if possible)
```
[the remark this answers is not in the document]
FreddieNeverLeft: Ok, changed this!

fl4health/clients/basic_client.py
FreddieNeverLeft: I added Enum LooggingMode Class on the top of baseclient

tests/clients/test_basic_client.py
FreddieNeverLeft: these are unfortunately required for the test to run properly

tests/clients/test_basic_client.py
FreddieNeverLeft: i added some extended functionality to the existing evaluate test, let me know if that is enough

fl4health/clients/basic_client.py
```diff
@@ -387,6 +390,7 @@ def _handle_logging(
         current_round: Optional[int] = None,
         current_epoch: Optional[int] = None,
         is_validation: bool = False,
+        is_testing: bool = False,
```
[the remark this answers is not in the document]
FreddieNeverLeft: added!

tests/clients/test_basic_client.py
FreddieNeverLeft: they are unfortunate required or else this test will fail

fl4health/clients/basic_client.py
```diff
@@ -635,30 +646,64 @@ def train_by_steps(
...
+        loss_dict = loss_meter.compute().as_dict()
+        metrics = metric_manager.compute()
+        self._handle_logging(loss_dict, metrics, logging_mode=logging_mode)
 
         return loss_dict["checkpoint"], metrics
 
+    def validate(self) -> Tuple[float, Dict[str, Scalar]]:
+        """
+        Validate the current model on the entire validation
+            or potentially an entire test dataset if it has been defined.
```
emersodb, line 690: The or here sort of implies that we only ever do one or the other. I would suggest changing the "or" to an "and"

fl4health/clients/basic_client.py
```diff
@@ -689,6 +734,8 @@ def setup_client(self, config: Config) -> None:
         train_loader, val_loader = self.get_data_loaders(config)
         self.train_loader = train_loader
         self.val_loader = val_loader
+        test_loader = self.get_test_data_loader(config)
```
emersodb, line 737: Any reason to do this in two stages instead of just doing `self.test_loader = ...`?

emersodb (commented):
Changes look good. The only additional thing I would do is add a test set to our basic example. That way you can make sure everything logs and runs as expected and it serves as an example of the new functionality.

tests/clients/test_basic_client.py
```diff
@@ -60,9 +60,11 @@ def test_metrics_reporter_fit() -> None:
 def test_metrics_reporter_evaluate() -> None:
     test_current_server_round = 2
     test_loss = 123.123
-    test_metrics: Dict[str, Union[bool, bytes, float, int, str]] = {"test_metric": 1234}
+    test_metrics: Dict[str, Scalar] = {"test_metric": 1234}
+    test_metrics_testing: Dict[str, Scalar] = {"testing_metric": 1234}
+    test_metrics = {"test_metric": 1234, "testing_metric": 1234, "test - loss": 123.123}
```
emersodb, line 65: It looks like you're immediately redefining test metrics here from above? I think you can at least delete that.

FreddieNeverLeft: I made a mistake with this change, for some reason test_metrics is modified in place, so I can only add a new one at the end to compare with, ill change the name in the next commit

fl4health/clients/basic_client.py
FreddieNeverLeft: fixed!

fl4health/clients/basic_client.py
FreddieNeverLeft: done!

FreddieNeverLeft:
Not entirely sure what could've caused the static code checks to fail @emersodb

emersodb:
> Not entirely sure what could've caused the static code checks to fail @emersodb

This is due to a pip-audit failure. This PR ([here](https://github.com/VectorInstitute/FL4Health/pull/154)) should fix your issues once it is merged in.

emersodb (approved):
Looks good to go.

One final question: Did you actually run the basic example and verify test metrics were produced?

FreddieNeverLeft:
yes I ran it, the test metrics were in the output after each round

PR #152, opened 2024-05-16
Part 1: Extracting a small code refactor from Sana's other work

emersodb:
# PR Type
Other

# Short Description

Small name and type changes. Incorporating an important change that Sana has made in her MkMMD work that could be separated out into a small PR.

# Tests Added

None, just a refactor.


fl4health/clients/ditto_client.py
```diff
@@ -345,15 +369,4 @@ def compute_evaluation_loss(
...
-        global_loss = self.criterion(preds["global"], target)
-
-        # Compute local model vanilla loss
-        assert "local" in preds
-        local_loss = self.criterion(preds["local"], target)
-
-        additional_losses = {"local_loss": local_loss, "global_loss": global_loss}
-
-        return EvaluationLosses(checkpoint=local_loss, additional_losses=additional_losses)
+        return super().compute_evaluation_loss(preds, features, target)
```
lotif, line 372: I think `super().compute_evaluation_loss` here will end up calling `compute_loss_and_additional_losses` from the super class instead of the one in this class? Is this right, and if so, is it the intended behaviour?

emersodb: That's a great question. I verified that the correct losses were being calculated when running the example. So it seems to be working as intended, which is to **call the Ditto `compute_loss_and_additional_losses` function**

However, just to verify that the behavior isn't an accident, I mocked up a test script with that behavior.
```
class A:
    def do_something(self) -> None:
        print("Class A!")

    def something_else(self) -> None:
        print("Class A: Something Else!")
        self.do_something()

class B(A):
    def do_something(self) -> None:
        print("Class B!")

    def something_else(self) -> None:
        print("Class B: Something Else!")
        super().something_else()

class_b = B()
class_b.something_else()
```
Output:
```
Class B: Something Else!
Class A: Something Else!
Class B!
```
So it is being routed the way we want.

sanaAyrml (approved):
All of changes look great for me.

fl4health/clients/ditto_client.py
lotif: Great, thanks for coding that up! Seems confusing but I'm glad it works the way it's supposed to

lotif (approved):

PR #160, opened 2024-05-30
Part 3: Migrating Moon Clients to use new losses

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8687xebzz)

This PR is Part 2 of several PRs aiming to simplify and decouple our implementations of FENDA and PerFCL. In this second part, we're simply migrating the existing MOON client implementation to use the new loss functions that have been created. The tests are also being migrated accordingly and expanded. Finally, fixing a typo (Extactor -> Extractor) in a few places.

# Tests Added

Migrating the MOON client tests and expanding them to cover more code.


fl4health/clients/moon_client.py
```diff
@@ -25,33 +22,58 @@ def __init__(
...
             loss_meter_type=loss_meter_type,
             checkpointer=checkpointer,
         )
-        self.cos_sim = torch.nn.CosineSimilarity(dim=-1).to(self.device)
-        self.ce_criterion = torch.nn.CrossEntropyLoss().to(self.device)
-        self.contrastive_weight = contrastive_weight
         self.temperature = temperature
+        self.contrastive_weight = contrastive_weight
+        if self.contrastive_weight == 0:
+            log(INFO, "Contrastive loss weight is set to 0, thus Contrastive loss will not be computed.")
```
lotif, line 59: I think the level here should be `WARNING`.

sanaAyrml (approved):
Everything is well.

fl4health/clients/moon_client.py
```diff
@@ -25,33 +22,58 @@ def __init__(
         loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
         checkpointer: Optional[ClientCheckpointModule] = None,
         temperature: float = 0.5,
-        contrastive_weight: Optional[float] = None,
+        contrastive_weight: float = 1.0,
         len_old_models_buffer: int = 1,
     ) -> None:
+        """
+        This client implements the MOON algorithm from Model-Contrastive Federated Learning. The key idea of MOON
+        is to utilize the similarity between model representations to correct the local training of individual parties,
```
jewelltaylor, line 30: nit: The key idea of MOON is to enforce similarity between representations from the global and local model through a contrastive loss to correct the local training of individual parties. Or something along those lines just to make it clear what the different models mean in this context

fl4health/clients/moon_client.py
```diff
@@ -25,33 +22,58 @@ def __init__(
...
         """
-        Computes the prediction(s) and features of the model(s) given the input.
+        Computes the prediction(s) and features of the model(s) given the input. This function also produces the
+        necessary features from the global_model (aggregated model from server) and old_models (previous client-side
+        optimized models) in order to be able to compute the appropriate contrastive loss.
 
         Args:
             input (TorchInputType): Inputs to be fed into the model. TorchInputType is simply an alias
-            for the union of torch.Tensor and Dict[str, torch.Tensor].
+            for the union of torch.Tensor and Dict[str, torch.Tensor]. Here, the Moon models require input to
```
jewelltaylor, line 75: nit: In the documentation, we have been going with the MOON as opposed to Moon it seems like. Just wanted to flag that

fl4health/clients/moon_client.py
```diff
@@ -137,9 +161,9 @@ def compute_loss_and_additional_losses(
             "loss": loss,
         }
 
-        if self.contrastive_weight:
-            contrastive_loss = self.get_contrastive_loss(
-                features["features"], features["global_features"], features["old_features"]
+        if "old_features" in features and "global_features" in features:
```
jewelltaylor, line 164: Maybe a quick comment to mention why we are unsqueezing 0th dimension of global features (ie only one positive pair expected)

jewelltaylor: Should we also enforce that these keys are in the features dict? Is there any cases where we use MOON and not want to use the contrastive loss

fl4health/clients/moon_client.py
emersodb: 👍

fl4health/clients/moon_client.py
emersodb: Yup, will make it consistent. I'll also do a lookup to see if we have that inconsistency elsewhere.

fl4health/clients/moon_client.py
emersodb: So these features will actually not be present for the server round. For example, there are no "old_features" in the first round, because there is no "previous round model" yet. I thought about using the server round instead of the presence of these features, but we don't currently pass the server round to this function. So thought I would do this for now and we can revisit this at a later stage.

jewelltaylor: Ahhh yes, that is right. Makes sense to me.

jewelltaylor (approved):
LGTM!

PR #161, opened 2024-05-30
Part 4: Addition of a loss config and FLamby client simplification

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8687xebzz)

This PR is Part 3 of several PRs aiming to simplify and decouple our implementations of FENDA and PerFCL. In this third part, we're adding in FENDA loss config for future use and simplifying the FLamby client scripts as they won't be able to use the additional loss weights  in future iterations.

# Tests Added

No tests in this one


lotif (approved):

sanaAyrml (approved):
Changes look amazing.

fl4health/losses/fenda_loss_config.py
```diff
@@ -0,0 +1,87 @@
...
+        local_feature_contrastive_loss_weight: float,
+        global_feature_loss_temperature: float = 0.5,
+        local_feature_loss_temperature: float = 0.5,
+    ) -> None:
+        self.global_feature_contrastive_loss_weight = global_feature_contrastive_loss_weight
+        self.local_feature_contrastive_loss_weight = local_feature_contrastive_loss_weight
+        self.perfcl_loss_function = PerFclLoss(device, global_feature_loss_temperature, local_feature_loss_temperature)
+
+
+class CosSimLossConfig:
```
jewelltaylor, line 24: nit: Mine as well expand this out to match format of corresponding loss like we have for the others. CosineSimilarityLossConfig

fl4health/losses/fenda_loss_config.py
```diff
@@ -0,0 +1,87 @@
...
+        self.cos_sim_loss_function = CosineSimilarityLoss(device)
+
+
+class ContrastiveLossConfig:
+    def __init__(self, device: torch.device, contrastive_loss_weight: float, temperature: float = 0.5) -> None:
+        self.contrastive_loss_weight = contrastive_loss_weight
+        self.contrastive_loss_function = ContrastiveLoss(device, temperature)
+
+
+class ConstrainedFendaLossConfig:
```
jewelltaylor, line 36: I may be missing the motivation and context on why you did this, but it seems a bit weird to pack in functionality to compute losses in attributes of an object called a LossConfig. It seems more natural to pass a loss config (or parameters that make make up a config) to a loss class to instantiate the loss and have the loss be callable or something

emersodb: I think that's a good point and perhaps suggests that I've named these classes incorrectly. I'm using them like loss function containers rather than configurations. For context, these classes are used to implement the ConstrainedFendaClient in Part 6 (I should have named this one Part 5 😂)

I'm a little hesitant to define loss classes with applies etc. just because loss functions sort of have a specific structure in pytorch. That is, they are of type `nn.Module` with a forward as their apply. Perhaps I'm thinking about it too much and should do that.

jewelltaylor: Yeah, I think a rename would be fine! I will leave it up to you if you want to go as far as defining a PyTorch loss. I think the functionality is great but perhaps the naming was just confusing when initially going through the code

emersodb: Sounds good. I'll do a "container" rename for now and rethink this down the road.

jewelltaylor (approved):
LGTM!

PR #164, opened 2024-06-05
FLASH iml

Justinyangjy:
# PR Type

# Short Description

[Clickup Ticket(s): Link(s).](https://app.clickup.com/t/8687wwu1z)

Added FLASH server side aggregation + FLASH client that performs early stopping

# Tests Added

Added Example and tests to ensure numerical correctness


Justinyangjy:
changed to draft as I wanted to add a better example than the generic one now

Justinyangjy:
the tests as well as my local tests give the same error rn, and its showing errors for basic_client example @emersodb

emersodb:
> the tests as well as my local tests give the same error rn, locally the basic example is also giving this error @emersodb

It isn't failing on the main branch. So it feels like this is probably related to the FLASH changes. Let's talk about it on Monday!

Justinyangjy:
its not passing locally but im glad its working here! basic cifar 10 example currently

examples/flash_example/README.md
```diff
@@ -0,0 +1,30 @@
+# FLASH Federated Learning Example
+
+In this example, we implement the FLASH algorithm from the paper [FLASH: Fast and Robust Federated Learning Through Approximate Second-Order Method](https://proceedings.mlr.press/v202/panchal23a/panchal23a.pdf). This method enhances federated learning by addressing both statistical heterogeneity and concept drift issues.
```
emersodb, line 3: Perhaps I'm missing something, but it looks like the title (at least from the linked PDF) is "Flash: Concept Drift Adaptation in Federated Learning" rather than FLASH: Fast and Robust Federated Learning Through Approximate Second-Order Method?

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+    Parameters,
+    Scalar,
+    ndarrays_to_parameters,
+    parameters_to_ndarrays,
+)
+from flwr.server.client_proxy import ClientProxy
+from flwr.server.strategy import FedOpt
+
+
+class FLASH(FedOpt):
```
emersodb, line 17: The standard naming convention that we've adopted is that acronyms (like FLASH) when part of classes are camel cased. So FLASH would be Flash or PerFCL would be PerFcl.

examples/flash_example/server.py
```diff
@@ -0,0 +1,85 @@
...
+        config["n_server_rounds"],
+        config["downsampling_ratio"],
+        local_epochs=config.get("local_epochs"),
+        local_steps=config.get("local_steps"),
+    )
+
+    initial_model = Net()
+
+    # Server performs simple FedAveraging as its server-side optimization strategy
+    strategy = FedAvg(
```
emersodb, line 49: Perhaps I'm missing something here, but this doesn't look like it's using the FLASH strategy that you created?

fl4health/clients/flash_client.py
```diff
@@ -0,0 +1,139 @@
...
+                self.update_after_step(local_step)
+                self.total_steps += 1
+                local_step += 1
+
+            metrics = self.train_metric_manager.compute()
+            loss_dict = self.train_loss_meter.compute().as_dict()
+            current_loss = loss_dict.get("backward", 0.0)
+
+            # Early stopping check
+            if self.gamma is not None and abs(previous_loss - current_loss) < self.gamma / (local_step + 1):
```
emersodb, line 77: I don't think we're quite computing the right early stopping criterion here, since we want to make sure that the current loss is strictly less than the previous loss. As it's currently written, consider the setting
```
previous_loss = 0.15
current_loss = 0.3
gamma = 0.06
local_step = 0
```
loss has actually increased, but since we're computing the absolute value, we're not going to early stop.

emersodb: Based on the definition in the FLASH paper, we don't want to use `local_step` here in the denominator for $\gamma/e$, as e represents the **epoch** we're on, rather than the step. Correct me if I'm wrong however.

emersodb, line 77: Finally, it looks like we're using the training loss here. In the paper, they use the validation loss to determine whether to perform early stopping. You should be able to get this loss value simply by calling `validate`.

fl4health/clients/flash_client.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Args:
+            data_path (Path): Path to the data directory.
+            metrics (Sequence[Metric]): List of metrics to be used for evaluation.
+            device (torch.device): Device to be used for training.
+            loss_meter_type (LossMeterType, optional):
+                Type of loss meter to be used. Defaults to LossMeterType.AVERAGE.
+            checkpointer (Optional[ClientCheckpointModule], optional): Checkpointer module defining when and how to do
+                checkpointing during client-side training. No checkpointing is done if not provided. Defaults to None.
+            gamma (float, optional):
```
emersodb, line 38: gamma isn't actually an argument to this client. So I'd take this out of the doc string.

fl4health/clients/flash_client.py
```diff
@@ -0,0 +1,139 @@
...
+        metrics = self.train_metric_manager.compute()
+
+        self._handle_logging(loss_dict, metrics, current_round=current_round)
+        self._handle_reporting(loss_dict, metrics, current_round=current_round)
+
+        return loss_dict, metrics
+
+    def setup_client(self, config: Config) -> None:
+        super().setup_client(config)
+        self.gamma = self.narrow_config_type(config, "gamma", float) if "gamma" in config else 0.04
```
emersodb, line 139: This will imply that the gamma always has a value (i.e. the user doesn't have the option to turn off early stopping. I'd say that we could just do
```
if "gamma" in config:
    self.gamma = self.narrow_config_type(config, "gamma", float)
```
and leave it none if the user has not specified the threshold.

fl4health/clients/flash_client.py
```diff
@@ -0,0 +1,139 @@
...
+            self.train_loss_meter.update(losses)
+            self.train_metric_manager.update(preds, target)
+            self.update_after_step(step)
+            self.total_steps += 1
+
+            loss_dict = self.train_loss_meter.compute().as_dict()
+            current_loss = loss_dict.get("backward", 0.0)
+
+            # Early stopping check
+            if self.gamma is not None and abs(previous_loss - current_loss) < self.gamma / (step + 1):
```
emersodb, line 123: In the FLASH paper, they don't actually have early stopping defined for "step-wise training." My guess is that this is for two reasons
1. Loss stochasticity between steps is a lot larger than over an entire epoch, meaning you'll be early stopping more often than you probably should
2. They measure validation loss and if you do that every step, that's very expensive. So my suggestion would be that we don't allow training by steps for FLASH clients for now. 

I think the easiest way to invalidate step-wise training would be to override `process_config` in the following way.
```
def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, None], int, bool]:
    local_epochs, local_steps, current_server_round, evaluate_after_fit = super().process_config(config)
    if local_steps  is not None:
        raise ValueError("Training by steps is not applicable for FLASH clients. Please define 'local_epochs' in your config instead")
    return local_epochs, local_steps, current_server_round, evaluate_after_fit
```

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+    """FLASH - Fast and Robust Federated Learning Through Approximate Second-Order Method.
```
emersodb, line 18: Similar comment to the readme file above, the title doesn't seem to match that in the linked PDF

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        Server-side learning rate. Defaults to 1e-1.
+    eta_l : float, optional
+        Client-side learning rate. Defaults to 1e-1.
+    beta_1 : float, optional
+        Momentum parameter. Defaults to 0.9.
+    beta_2 : float, optional
+        Second moment parameter. Defaults to 0.99.
+    tau : float, optional
+        Controls the algorithm's degree of adaptability. Defaults to 1e-9.
+    d_t : NDArrays, optional
```
emersodb, line 58: This isn't an argument to the constructor. So I would move your documentation out of this docstring and to where the properties is declared.

emersodb: It would also be helpful to state what the initial value is. For this algorithm, d_0 is all zeros.

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        """Aggregate fit results using the FLASH method."""
+        fedavg_parameters_aggregated, metrics_aggregated = super().aggregate_fit(
+            server_round=server_round, results=results, failures=failures
+        )
+        if fedavg_parameters_aggregated is None:
+            return None, {}
+
+        fedavg_weights_aggregate = parameters_to_ndarrays(fedavg_parameters_aggregated)
+
+        # FLASH
```
emersodb, line 131: I'd say we can drop this comment, or elaborate a bit more, as `delta_t` isn't specific to FLASH

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        return rep
+
+    def aggregate_fit(
+        self,
+        server_round: int,
+        results: List[Tuple[ClientProxy, FitRes]],
+        failures: List[Union[Tuple[ClientProxy, FitRes], BaseException]],
+    ) -> Tuple[Optional[Parameters], Dict[str, Scalar]]:
+        """Aggregate fit results using the FLASH method."""
+        fedavg_parameters_aggregated, metrics_aggregated = super().aggregate_fit(
```
emersodb, line 123: Because your FLASH strategy inherits from FedOpt, this `aggregate_fit` method is actually doing weighted parameter aggregation. In the FLASH algorithm, it looks like they are strictly doing uniform averaging of the parameter updates. I'm alright with having this be an option, but by default we want to follow the FLASH implementation. You can do uniform or weighted aggregation of both the weights and metrics by inheriting from `BasicFedAvg` in `fl4health/strategies/basic_fedavg.py`

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        if not self.v_t:
+            self.v_t = [np.zeros_like(x) for x in delta_t]
+        self.v_t = [self.beta_2 * x + (1 - self.beta_2) * np.multiply(y, y) for x, y in zip(self.v_t, delta_t)]
+
+        # d_t
+        if self.d_t is None:
+            self.d_t = [np.zeros_like(x) for x in delta_t]
+        beta_3 = [
+            np.linalg.norm(v_prev) / (np.linalg.norm((delta**2) - v_prev) + np.linalg.norm(v_prev))
+            for delta, v_prev in zip(delta_t, self.v_t)
```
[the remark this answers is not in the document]
emersodb: Another issue here is that you have already updated `self.v_t` here so that it is actually $v^{(r)}$ not $v^{(r-1)}$ as in the equation for $\beta_{3j}$. This calculation either needs to happen before this update, or you need to store the previous values.

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        ]
+        self.d_t = [
+            b3 * d_prev + (1 - b3) * ((delta**2) - v_prev)
+            for b3, d_prev, delta, v_prev in zip(beta_3, self.d_t, delta_t, self.v_t)
```
emersodb, line 153: Calling the components of `self.v_t` `v_prev` here is a bit misleading. At this point you've updated `self.v_t` such that it represents $v^{(r)}$, which is correct. So the `prev` part isn't quite accurate.

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,164 @@
...
+        ]
+
+        # Update global weights
+        new_weights = [
+            x + self.eta * y / (np.sqrt(z) - d + self.tau)
+            for x, y, z, d in zip(self.current_weights, self.m_t, self.v_t, self.d_t)
```
emersodb, line 159: `x,y,z`, and `d` here are a bit short and could be more descriptive.

emersodb: I do note that these are the variables used in the FedAdam implementation, but we strive to be a little more verbose 🙂

tests/strategies/test_flash.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # First layer weighted aggregate should be all -0.00352075
+    weighted_target_1 = np.ones((3, 3)) * (-0.00352075)
+    # Second layer weighted aggregate should be all -0.00337517
+    weighted_target_2 = np.ones((4, 4)) * (-0.00337517)
+
+    assert parameters is not None
+    aggregated_ndarrays = parameters_to_ndarrays(parameters)
+    assert np.allclose(weighted_target_1, aggregated_ndarrays[0])
+    assert np.allclose(weighted_target_2, aggregated_ndarrays[1])
```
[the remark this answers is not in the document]
emersodb: It might also make your life easier to break up the pieces of these updates into "testable" modules. That is, something like 
```
def update_m_t(...) ...
def update_v_t(...) ...
def compute_beta_3(...)...
def update_d_t(...)...
```
etc.

examples/flash_example/README.md
Justinyangjy: Changed, I have no idea where I got this name from 🤔

fl4health/strategies/flash.py
Justinyangjy: got it, updated and will do this for new methods

examples/flash_example/server.py
Justinyangjy: good catch! corrected this.

fl4health/strategies/flash.py
Justinyangjy: corrected!

fl4health/strategies/flash.py
Justinyangjy: removed

fl4health/strategies/flash.py
emersodb: Sounds good! There is a bit of ambiguity in the style guides about how to case acronyms. We've just selected this convention.

tests/strategies/test_flash.py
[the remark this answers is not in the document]
Justinyangjy: changed in the latest commit!

fl4health/strategies/flash.py
Justinyangjy: done, I will change this  for the remaining functions

fl4health/strategies/flash.py
Justinyangjy: corrected in the latest commit

fl4health/strategies/flash.py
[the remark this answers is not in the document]
Justinyangjy: corrected to element wise and stored v_prev, will have to test for correctness

fl4health/strategies/flash.py
Justinyangjy: changed to BasicFedAvg parent class

fl4health/strategies/flash.py
Justinyangjy: moved to init, removing d_t as its not a parameter

fl4health/clients/flash_client.py
Justinyangjy: done in latest commit!

fl4health/clients/flash_client.py
Justinyangjy: done!

fl4health/clients/flash_client.py
Justinyangjy: I think this makes sense, disabled local steps for now

fl4health/clients/flash_client.py
Justinyangjy: done in the latest commit

Justinyangjy: i agree, I copy pasted the wrong variable, using local_epochs now

tests/strategies/test_flash.py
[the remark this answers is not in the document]
Justinyangjy: test done!

Justinyangjy:
Had to use numpy element-wise operations as multiple zip()'s was too slow for smoke tests

fl4health/clients/flash_client.py
```diff
@@ -0,0 +1,100 @@
...
+            loss_meter_type=loss_meter_type,
+            checkpointer=checkpointer,
+        )
+        # gamma: Threshold for early stopping based on the change in validation loss.
+        self.gamma: Optional[float] = None
+
+    def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, None], int, bool]:
+        local_epochs, local_steps, current_server_round, evaluate_after_fit = super().process_config(config)
+        if local_steps is not None:
+            raise ValueError(
```
jewelltaylor, line 52: It is too bad that the method is not applicable for train by steps since one of the benchmarks we have been using (FLamby) uses train by steps in the experimental setup. I guess the main issue is that there would be too much variation in the per step losses? Or is there also other factors that do not make it applicable? Just curious for my understanding :)

Justinyangjy: That is quite unfortunate :( one thing is as you mentioned the variation in the per step losses, another thing David mentioned was that FLASH uses validation for these early stopping criteria, which is quite computationally expensive , so it is disabled for now.

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,175 @@
...
+            # Update m_t
+            self.m_t[i] = self.beta_1 * m + (1 - self.beta_1) * delta
+
+            # Update v_t
+            self.v_t[i] = self.beta_2 * v + (1 - self.beta_2) * np.square(delta)
+
+            # Compute beta_3
+            norm_v_prev = np.abs(v)
+            norm_diff = np.abs(np.square(delta) - self.v_t[i])
+            beta_3_matrix = norm_v_prev / (norm_diff + norm_v_prev)
```
jewelltaylor, line 140: Should these be euclidean norms as per line 19 in Algorithm 1 in the paper? I could be missing something though I see you and David went back and forth about similar part of the code earlier

fl4health/strategies/flash.py
```diff
@@ -0,0 +1,175 @@
...
+        fedavg_parameters_aggregated, metrics_aggregated = super().aggregate_fit(
+            server_round=server_round, results=results, failures=failures
+        )
+        if fedavg_parameters_aggregated is None:
+            return None, {}
+
+        fedavg_weights_aggregate = parameters_to_ndarrays(fedavg_parameters_aggregated)
+
+        delta_t: NDArrays = [x - y for x, y in zip(fedavg_weights_aggregate, self.current_weights)]
+        for attr in ["m_t", "v_t", "d_t"]:
```
jewelltaylor, line 161: Not a big deal at all but can these be initialized in the constructor instead? Or am i missing something?

fl4health/clients/flash_client.py
```diff
@@ -71,69 +79,22 @@ def train_by_epochs(
...
-
-        self._handle_logging(loss_dict, metrics, current_round=current_round)
-        self._handle_reporting(loss_dict, metrics, current_round=current_round)
-
         return loss_dict, metrics
 
     def setup_client(self, config: Config) -> None:
         super().setup_client(config)
-        self.gamma = self.narrow_config_type(config, "gamma", float) if "gamma" in config else 0.04
+        if "gamma" in config:
```
emersodb, line 99: I think you can add an else statement here if gamma isn't in the config that just logs that early stopping is disabled. Something else
```
if "gamma" in config:
    self.gamma = self.narrow_config_type(config, "gamma", float)
else:
    log(INFO, "Gamma not present in config. Early stopping is disabled.")

File: fl4health/clients/flash_client.py [hunk 20]
```diff
@@ -71,69 +79,22 @@ def train_by_epochs(
...
-
-            loss_dict = self.train_loss_meter.compute().as_dict()
-            current_loss = loss_dict.get("backward", 0.0)
-
-            # Early stopping check
-            if self.gamma is not None and abs(previous_loss - current_loss) < self.gamma / (step + 1):
-                log(INFO, f"Early stopping at step {step} with loss change {abs(previous_loss - current_loss)}")
+            if self.gamma is not None and previous_loss - current_loss < self.gamma / (local_epoch + 1):
+                log(
+                    INFO, f"Early stopping at epoch {local_epoch} with loss change {abs(previous_loss - current_loss)}"
```
emersodb, line 89: I think it's worth logging what gamma is in this message just for visual comparison.

fl4health/strategies/flash.py
emersodb: I think it's worth setting these values in the `__init__`. Let us know if we're missing a reason that's not possible though, Justin.

fl4health/strategies/flash.py
```diff
@@ -100,26 +105,50 @@ def __init__(
...
         """Compute a string representation of the strategy."""
-        rep = f"FLASH(accept_failures={self.accept_failures})"
+        rep = f"Flash(accept_failures={self.accept_failures})"
         return rep
 
+    def _update_parameters(self, delta_t: NDArrays) -> None:
+        """Update m_t, v_t, beta_3, and d_t per-element"""
+        assert self.v_t is not None and self.m_t is not None and self.d_t is not None
+
+        for i, (delta, m, v, d_prev) in enumerate(zip(delta_t, self.m_t, self.v_t, self.d_t)):
```
emersodb, line 130: Super minor, but I'd attach the `_prev` prefix to `m` and `v` here too just to be clear that they are the "previous values"

fl4health/strategies/flash.py
emersodb: In line 19 of the algorithm, the best we can tell is that the operations are being carried out element-wise (using the index j). The l2-norm of a single element is just the absolute value ($\Vert a \Vert_2 = \sqrt{a^2} = \vert a \vert$).

fl4health/strategies/flash.py
```diff
@@ -100,26 +105,50 @@ def __init__(
...
+            # Update m_t
+            self.m_t[i] = self.beta_1 * m + (1 - self.beta_1) * delta
+
+            # Update v_t
+            self.v_t[i] = self.beta_2 * v + (1 - self.beta_2) * np.square(delta)
```
emersodb, line 135: Since we're computing `np.square(delta)` three times here, it's probably worth precomputing it and then reusing that value?

fl4health/strategies/flash.py
jewelltaylor: Ahh, I assumed j was an index over layers or something but I see from the discussion from before we have reason to believe it is not. Seems pretty weird to have the l2-norm in the algorithm since its elementwise though haha

emersodb: Yeah I agree. There isn't any indication that $\Delta^{(5)}$ or any of the other parameters are structured element-wise that I could find. If they didn't have a broken link to their code, we could verify, but it's not there 🤦

emersodb:
I really like all the changes you made. The tests you implemented are fantastic! I added a few very small comments. Once those are in, I think we're ready to merge!

fl4health/strategies/flash.py
Justinyangjy: unfortunately the paper was super unclear when it came to the updating, and its never explicitly stated what `n` stands for, the only logical conclusion was is the update happpens per-element, which meant the norms of each element is equivalent to the abs for the matrix

emersodb (approved):
I think this looks good to go to me!

jewelltaylor (approved):
LGTM!

PR #155, opened 2024-05-21
nnUNet PICAI

jewelltaylor:
# PR Type
[**Feature** | Fix | Documentation | Other ]

# Short Description

Clickup Ticket(s): [Initial nnUNet Integration for PICAI research](https://app.clickup.com/t/8687xfxp6)

This PR does some initial work to leverage the nnUNet framework in FL4Health. This includes:
- Moving preprocessing into the repo so that we have the flexibility to produce preprocessed datasets in the format that nnUNet expects.
- Run preprocessing on raw dataset on cluster to produce desired dataset on cluster. Remove old datasets.
- Update the existing FL and central examples to use the new dataset format. 
- Add scripts to train nnUNet on the Whole Gland Segmentation task. One script to train from scratch on PICAI dataset. Another script to finetune a model already trained on a private dataset in the Haider Lab. 
- Update scripts to requeue if they are still running when they timeout (previously a huge issue since training always takes longer than max job time)
- Update documentation in the picai research folder to include the new functionality

In following tickets I will: 
- Do some initial experiments to establish some baselines.
- More sophisticated data augmentation pipeline, possibly borrowed from nnUNet, to apply to current central and FL examples.
- Integrate architecture and training pipeline configuration from nnUNet to the federated setting. 

# Tests Added
Since a lot of this is research code and subject to change, I have not included tests. However, let me know if you think there are any tests that would be useful.

PR #165, opened 2024-06-06
Add tests for picai transforms and other test fixes

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Add Picai Tranforms Tests](https://app.clickup.com/t/8688q3ah7)

In #155, I added a preprocessing transformation pipeline for PICAI. In this PR, I simply have added tests for those transforms and also fixed a small issue in loss tests from previous PRs when using GPU.

# Tests Added

- Tests for PICAI transforms
- Updated Loss tests

### DIFF
--- a/research/picai/preprocessing_transforms.py
+++ b/research/picai/preprocessing_transforms.py
@@ -33,7 +33,7 @@ def resample_img(
     original_size = image.GetSize()
 
     # convert PICAI z, y, x (Depth x Height x Width) convention to SimpleITK's convention x, y, z
-    out_spacing = list(reversed(list(spacing)))
+    out_spacing = (spacing[2], spacing[1], spacing[0])
 
     if size is None:
         # calculate output size in voxels
--- a/research/picai/preprocessing_transforms.py
+++ b/research/picai/preprocessing_transforms.py
@@ -6,7 +6,7 @@ import SimpleITK as sitk
 
 def resample_img(
     image: sitk.Image,
-    spacing: Tuple[float, float, float] = (2.0, 2.0, 2.0),
+    spacing: Tuple[float, float, float],
     size: Optional[Tuple[int, int, int]] = None,
     is_label: bool = False,
     pad_value: Optional[Union[float, int]] = 0.0,
@@ -33,7 +33,7 @@ def resample_img(
     original_size = image.GetSize()
 
     # convert PICAI z, y, x (Depth x Height x Width) convention to SimpleITK's convention x, y, z
-    out_spacing = list(reversed(spacing))
+    out_spacing = list(reversed(list(spacing)))
 
     if size is None:
         # calculate output size in voxels
@@ -135,7 +135,7 @@ def input_verification_crop_or_pad(
 
 def crop_or_pad(
     image: sitk.Image,
-    size: Tuple[int, int, int] = (20, 256, 256),
+    size: Tuple[int, int, int],
     physical_size: Optional[Tuple[float, float, float]] = None,
     crop_only: bool = False,
 ) -> sitk.Image:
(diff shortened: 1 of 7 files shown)


research/picai/preprocessing_transforms.py
```diff
@@ -33,7 +33,7 @@ def resample_img(
     original_size = image.GetSize()
 
     # convert PICAI z, y, x (Depth x Height x Width) convention to SimpleITK's convention x, y, z
-    out_spacing = list(reversed(spacing))
+    out_spacing = list(reversed(list(spacing)))
```
emersodb, line 36: I know reversed was my suggestion. However, now that we've moved from sequences to tuples, maybe we just do the following?
```
out_spacing = (spacing[2], spacing[1], spacing[0])
```

research/picai/preprocessing_transforms.py
```diff
@@ -157,13 +157,14 @@ def crop_or_pad(
...
     shape, size = input_verification_crop_or_pad(image, size, physical_size)
 
-    # set identity operations for cropping and padding
-    rank = len(size)
-    padding = [[0, 0]] * rank
-    slicer = [slice(None)] * rank
+    # Since the subarrays are being set in the below for loop
+    # We have to ensure that they are seperate lists
+    # and not the same reference (ie [[0, 0]] * rank)
+    padding = [[0, 0] for _ in range(len(size))]
```
emersodb, line 163: Ah interesting. I didn't realize this tied them through reference like that. It doesn't work that way for things like `ints`. Good to know for the future.

research/picai/preprocessing_transforms.py
```diff
@@ -177,11 +178,12 @@ def crop_or_pad(
             idx_end = idx_start + size[i]
             slicer[i] = slice(idx_start, idx_end)
 
+    new_image = image[tuple(slicer)]
+
     # crop and/or pad image
-    if isinstance(image, sitk.Image):
+    if any(pad[0] > 0 or pad[1] > 0 for pad in padding):
```
emersodb, line 184: Any reason we're "dropping" support for non-sitk.Image types?

research/picai/preprocessing_transforms.py
jewelltaylor: Yeah I did not realize it either, I think it is only for mutable types

tests/losses/test_contrastive_loss.py
```diff
@@ -18,21 +18,21 @@ def test_contrastive_loss() -> None:
         features=local_features.reshape(len(local_features), -1),
         positive_pairs=global_features.reshape(1, len(global_features), -1),
         negative_pairs=previous_local_features.reshape(1, len(previous_local_features), -1),
-    )
+    ).cpu()
```
emersodb, line 21: I'm not sure if you think this is better or not, but `.item()` automatically moves the value to cpu and produces a plain python float.

research/picai/preprocessing_transforms.py
jewelltaylor: I definitely want to support a wider variety of image types and for some its pretty trivial, but at this point just to ensure everything matches how its currently typed which doesn't include numpy arrays

tests/losses/test_contrastive_loss.py
jewelltaylor: yeah sure I can do that

research/picai/preprocessing_transforms.py
emersodb: Sounds good. I'm good with doing it. Was just curious.

emersodb (approved):
Outside of the few small comments I left, this is good to go. Thanks for fixing the unit tests!

PR #170, opened 2024-06-07
PR4: Add deep_mmd_loss files

sanaAyrml:
# PR Type
[Feature]

# Short Description

This is a tentative implementation for deep mmd loss.

# Tests Added

No tests added yet.


fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+        Args:
+            data_path (Path): path to the data to be used to load the data for client-side training
+            metrics (Sequence[Metric]): Metrics to be computed based on the labels and predictions of the client model
+            device (torch.device): Device indicator for where to send the model, batches, labels etc. Often 'cpu' or
+                'cuda'
+            loss_meter_type (LossMeterType, optional): Type of meter used to track and compute the losses over
+                each batch. Defaults to LossMeterType.AVERAGE.
+            checkpointer (Optional[TorchCheckpointer], optional): Checkpointer to be used for client-side
+                checkpointing. Defaults to None.
+            metrics_reporter (Optional[MetricsReporter], optional): A metrics reporter instance to record the metrics
```
emersodb, line 43: I think this is probably an issue with the DittoClient docstring as well, but the `metrics_reporter` is technically not an arg in this implementation.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+                during the execution. Defaults to an instance of MetricsReporter with default init parameters.
+            lam (float, optional): weight applied to the Ditto drift loss. Defaults to 1.0.
+            deep_mmd_loss_weight (float, optional): weight applied to the DEEP-MMD loss. Defaults to 10.0.
+            flatten_feature_extraction_layers (Dict[str, bool], optional): Dictionary of layers to extract features
+                from them what is the flattened feature size. Defaults to {}. If it is -1 then the layer is not
+                flattened.
```
emersodb, line 49: It looks like `size_feature_extraction_layers` is missing from the args documentation here?

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+        self,
+        data_path: Path,
+        metrics: Sequence[Metric],
+        device: torch.device,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        checkpointer: Optional[ClientCheckpointModule] = None,
+        lam: float = 1.0,
+        deep_mmd_loss_weight: float = 10.0,
+        flatten_feature_extraction_layers: Dict[str, bool] = {},
+        size_feature_extraction_layers: Dict[str, int] = {},
```
emersodb, line 30: We noted that we don't want mutable defaults in the previous PR. Just wanted to put this here as a reminder to change these around too 🙂

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+            forward.
+        """
+
+        # We use features from init_global_model to compute the DEEP-MMD loss not the global_model
+        global_preds = self.global_model(input)
+        local_preds = self.model(input)
+        features = self.local_feature_extractor.get_extracted_features()
+        if self.deep_mmd_loss_weight != 0:
+            # Compute the features of the init_global_model
+            _ = self.init_global_model(input)
```
emersodb, line 129: I don't think you need to catch this with `_` you're not going to store it anyway.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+            preds (Dict[str, torch.Tensor]): Prediction(s) of the model(s) indexed by name. Anything stored
+                in preds will be used to compute metrics.
+            features: (Dict[str, torch.Tensor]): Feature(s) of the model(s) indexed by name.
+            target: (torch.Tensor): Ground truth data to evaluate predictions against.
+
+        Returns:
+            EvaluationLosses: an instance of EvaluationLosses containing checkpoint loss and additional losses
+                indexed by name.
+        """
+        for layer in self.flatten_feature_extraction_layers.keys():
```
emersodb, line 161: If you're going to be indexing into `self.deep_mmd_losses` anyway, could we simply do 
```
for layer_loss_module in self.deep_mmd_losses.values():
    layer_loss_module.training = False
```
Correct me if I'm wrong, but you likely also want to do something like `layer_loss_module.requires_grad=False as well here?

emersodb: I think it's still worth overriding `compute_evaluation_loss` and `compute_training_loss` and asserting that all `layer_loss_module.training == False` or vice versa though to be safe 🙂

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+                - A dictionary with `local_loss`, `global_loss`, `total_loss` and, based on client attributes set
+                from server config, also `deep_mmd_loss`, `feature_l2_norm_loss` keys and their respective calculated
+                values.
+        """
+        total_loss, additional_losses = super().compute_loss_and_additional_losses(preds, features, target)
+
+        if self.deep_mmd_loss_weight != 0:
+            # Compute DEEP-MMD loss
+            total_deep_mmd_loss = torch.tensor(0.0, device=self.device)
+            for layer in self.flatten_feature_extraction_layers.keys():
```
emersodb, line 191: As above, since we're not accessing components of `flatten_feature_extraction_layers`, it might be more straightforward to do:
```
for layer, layer_loss_module in self.deep_mmd_losses.items():
    layer_deep_mmd_loss = layer_loss_module(
        features[layer], features[" ".join(["init_global", layer])]
    )
    additional_losses["_".join(["deep_mmd_loss", layer])] = layer_deep_mmd_loss
    total_deep_mmd_loss += layer_deep_mmd_loss
```

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+import numpy as np
+import torch
+
+
+class ModelLatentF(torch.nn.Module):
+    """Latent space for both domains."""
+
+    def __init__(self, x_in: int, H: int, x_out: int):
+        """Init latent features."""
+        super(ModelLatentF, self).__init__()
```
emersodb, line 12: I know this isn't all your code, but this can be changed to `super().__init__()`

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        self.restored = False
```
emersodb, line 13: Is restored used anywhere else?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
+from typing import Optional, Tuple
+
+import numpy as np
+import torch
+
+
+class ModelLatentF(torch.nn.Module):
+    """Latent space for both domains."""
+
+    def __init__(self, x_in: int, H: int, x_out: int):
```
emersodb, line 10: I'm not sure if `H` here has a specific meaning elsewhere, but it essentially looks like the hidden dimension of this kernel network. So maybe we can just call it `hidden_dimension`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            torch.nn.Softplus(),
+            torch.nn.Linear(H, H, bias=True),
+            torch.nn.Softplus(),
+            torch.nn.Linear(H, x_out, bias=True),
+        )
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        """Forward the LeNet."""
+        fealant = self.latent(input)
+        return fealant
```
emersodb, line 27: lol...weird word

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        lr: float = 0.001,
+        hidden_size: int = 10,
+        output_size: int = 50,
+        layer_name: Optional[str] = None,
+        training: bool = True,
+    ) -> None:
+        super().__init__()
+        self.device = device
+        self.lr = lr
+        self.layer_name = layer_name
```
emersodb, line 44: Any utility to the `layer_name` here? It seems like it's not used, but I may be missing something.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        # Initialize parameters
+        self.epsilonOPT: torch.Tensor = torch.log(torch.from_numpy(np.random.rand(1) * 10 ** (-10)).to(self.device))
+        self.epsilonOPT.requires_grad = self.training
+        self.sigmaOPT: torch.Tensor = torch.sqrt(torch.tensor(2 * 32 * 32, dtype=torch.float).to(self.device))
+        self.sigmaOPT.requires_grad = self.training
+        self.sigma0OPT: torch.Tensor = torch.sqrt(torch.tensor(0.005, dtype=torch.float).to(self.device))
+        self.sigma0OPT.requires_grad = self.training
+
+        # Initialize optimizers
+        self.optimizer_F = torch.optim.Adam(
```
emersodb, line 59: Maybe we can use AdamW here, just for the decay correction?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        self.device = device
+        self.lr = lr
+        self.layer_name = layer_name
+        self.training = training
+
+        # Initialize the model
+        self.featurizer = ModelLatentF(input_size, hidden_size, output_size).to(self.device)
+
+        # Initialize parameters
+        self.epsilonOPT: torch.Tensor = torch.log(torch.from_numpy(np.random.rand(1) * 10 ** (-10)).to(self.device))
```
emersodb, line 51: Just to go with our naming conventions, maybe we can do *_opt for all of these variables with OPT suffixes?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        self.sigmaOPT.requires_grad = self.training
+        self.sigma0OPT: torch.Tensor = torch.sqrt(torch.tensor(0.005, dtype=torch.float).to(self.device))
+        self.sigma0OPT.requires_grad = self.training
+
+        # Initialize optimizers
+        self.optimizer_F = torch.optim.Adam(
+            list(self.featurizer.parameters()) + [self.epsilonOPT] + [self.sigmaOPT] + [self.sigma0OPT], lr=self.lr
+        )
+
+    def Pdist2(self, x: torch.Tensor, y: Optional[torch.Tensor]) -> torch.Tensor:
```
emersodb, line 63: maybe expand this to `pairwise_distiance_squared`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        """compute the paired distance between x and y."""
+        x_norm = (x**2).sum(1).view(-1, 1)
+        if y is not None:
+            y_norm = (y**2).sum(1).view(1, -1)
+        else:
+            y = x
+            y_norm = x_norm.view(1, -1)
+        Pdist = x_norm + y_norm - 2.0 * torch.mm(x, torch.transpose(y, 0, 1))
```
emersodb, line 71: same with this, maybe just `paired_distance`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        mmd_std_temp = torch.sqrt(mmd_var_temp + 10 ** (-8))
+        STAT_u = torch.div(-1 * mmd_value_temp, mmd_std_temp)
+        # Compute gradient
+        STAT_u.backward()
+        # Update weights using gradient descent
+        self.optimizer_F.step()
+        return
+
+    def compute_kernel(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor:
+        """Train the kernel."""
```
emersodb, line 193: This is actually not training the kernel, but rather doing pure inference I think?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            len_s=X.shape[0],
+            Fea_org=features.view(features.shape[0], -1),
+            sigma=sigma,
+            sigma0=sigma0_u,
+            epsilon=ep,
+        )
+
+        return mmd_value_temp
+
+    def forward(self, Xs: torch.Tensor, Xt: torch.Tensor) -> torch.Tensor:
```
emersodb, line 220: So based on our implementation, am I correct in thinking that during training, we'll also be updating the kernel values at each step based on the batches? If this doesn't quite work, another thing we could do is update the deep kernels periodically, like we do with MKMMD where we "optimize" the kernels for a frozen set of models.

fl4health/losses/deep_mmd_loss.py
emersodb: maybe `feature_latent_map`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            return mmd2, None, Kxyxy
+        hh = Kx + Ky - Kxy - Kxy.transpose(0, 1)
+        V1 = torch.dot(hh.sum(1) / ny, hh.sum(1) / ny) / ny
+        V2 = (hh).sum() / (nx) / nx
+        varEst = 4 * (V1 - V2**2)
+        return mmd2, varEst, Kxyxy
+
+    def MMDu(
+        self,
+        Fea: torch.Tensor,
```
emersodb, line 113: maybe rename to `features`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        self.sigmaOPT.requires_grad = False
+        self.sigma0OPT.requires_grad = False
+        self.epsilonOPT.requires_grad = False
+
+        features = torch.cat([X, Y], 0)
+
+        # Compute output of deep network
+        model_output = self.featurizer(features)
+        # Compute epsilon, sigma and sigma_0
+        ep = torch.exp(self.epsilonOPT) / (1 + torch.exp(self.epsilonOPT))
```
emersodb, line 205: rename `epsilon` and note that it is the epsilon in $\kappa_w(x, y)$ in the paper

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        V1 = torch.dot(hh.sum(1) / ny, hh.sum(1) / ny) / ny
+        V2 = (hh).sum() / (nx) / nx
+        varEst = 4 * (V1 - V2**2)
+        return mmd2, varEst, Kxyxy
+
+    def MMDu(
+        self,
+        Fea: torch.Tensor,
+        len_s: int,
+        Fea_org: torch.Tensor,
```
emersodb, line 115: rename `original_features` for clarity.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        #  Train deep network for MMD-D
+        # ------------------------------
+        # Initialize optimizer
+        self.optimizer_F.zero_grad()
+        # Compute output of deep network
+        model_output = self.featurizer(features)
+        # Compute epsilon, sigma and sigma_0
+        ep = torch.exp(self.epsilonOPT) / (1 + torch.exp(self.epsilonOPT))
+        sigma = self.sigmaOPT**2
+        sigma0_u = self.sigma0OPT**2
```
emersodb, line 172: based on the implementation of `MMDu` I would suggest renaming `sigma0` to `sigma_phi`, `sigma0OPT` to `sigma_phi_opt` and `sigma0_u` to `sigma_phi` (since there doesn't seem to be any reason to have `_u` in there anyway

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            else:
+                xy = torch.div(torch.sum(Kxy), (nx * ny))
+            mmd2 = xx - 2 * xy + yy
+        if not is_var_computed:
+            return mmd2, None, Kxyxy
+        hh = Kx + Ky - Kxy - Kxy.transpose(0, 1)
+        V1 = torch.dot(hh.sum(1) / ny, hh.sum(1) / ny) / ny
+        V2 = (hh).sum() / (nx) / nx
+        varEst = 4 * (V1 - V2**2)
+        return mmd2, varEst, Kxyxy
```
emersodb, line 109: Correct me if I'm wrong, but in all instances where this function is called, we don't make use of `Kxyxy`. Can we just drop it from the return values?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+    def h1_mean_var_gram(
+        self, Kx: torch.Tensor, Ky: torch.Tensor, Kxy: torch.Tensor, is_var_computed: bool, use_1sample_U: bool = True
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor], torch.Tensor]:
+        """compute value of MMD and std of MMD using kernel matrix."""
+        Kxxy = torch.cat((Kx, Kxy), 1)
+        Kyxy = torch.cat((Kxy.transpose(0, 1), Ky), 1)
+        Kxyxy = torch.cat((Kxxy, Kyxy), 0)
+        nx = Kx.shape[0]
+        ny = Ky.shape[0]
+        is_unbiased = True
```
emersodb, line 84: Perhaps, rather than hardcoding this here, we can turn it into a class variable and default it to True (i.e. `self.is_unbiased`? That way it can be pulled out of this function.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        sigma = self.sigmaOPT**2
+        sigma0_u = self.sigma0OPT**2
+        # Compute Compute J (STAT_u)
+        mmd_value_temp, _, _ = self.MMDu(
+            Fea=model_output,
+            len_s=X.shape[0],
+            Fea_org=features.view(features.shape[0], -1),
+            sigma=sigma,
+            sigma0=sigma0_u,
+            epsilon=ep,
```
emersodb, line 215: I think we want to set `is_var_computed = False` in this call to save some calculation

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        sigma = self.sigmaOPT**2
+        sigma0_u = self.sigma0OPT**2
+        # Compute Compute J (STAT_u)
+        mmd_value_temp, mmd_var_temp, _ = self.MMDu(
+            Fea=model_output,
+            len_s=X.shape[0],
+            Fea_org=features.view(features.shape[0], -1),
+            sigma=sigma,
+            sigma0=sigma0_u,
+            epsilon=ep,
```
emersodb, line 180: Maybe explicitly set `is_var_computed=True` in this call rather than leaving it to the default?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        """compute value of deep-kernel MMD and std of deep-kernel MMD using merged data."""
+        X = Fea[0:len_s, :]  # fetch the sample 1 (features of deep networks)
+        Y = Fea[len_s:, :]  # fetch the sample 2 (features of deep networks)
+        X_org = Fea_org[0:len_s, :]  # fetch the original sample 1
+        Y_org = Fea_org[len_s:, :]  # fetch the original sample 2
+        L = 1  # generalized Gaussian (if L>1)
+        Dxx = self.Pdist2(X, X)
+        Dyy = self.Pdist2(Y, Y)
+        Dxy = self.Pdist2(X, Y)
+        Dxx_org = self.Pdist2(X_org, X_org)
```
emersodb, line 132: I think we can put these inside the `if is_smooth` block, since they only need to be computed if that is true. Otherwise they aren't used.

emersodb: Maybe also rename these to expand `org` to `original` just to be explicit

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        epsilon: torch.Tensor,
+        is_smooth: bool = True,
+        is_var_computed: bool = True,
+        use_1sample_U: bool = True,
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor], torch.Tensor]:
+        """compute value of deep-kernel MMD and std of deep-kernel MMD using merged data."""
+        X = Fea[0:len_s, :]  # fetch the sample 1 (features of deep networks)
+        Y = Fea[len_s:, :]  # fetch the sample 2 (features of deep networks)
+        X_org = Fea_org[0:len_s, :]  # fetch the original sample 1
+        Y_org = Fea_org[len_s:, :]  # fetch the original sample 2
```
emersodb, line 127: Maybe rename these to `X_original` and `Y_original`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        L = 1  # generalized Gaussian (if L>1)
```
emersodb, line 128: Rather than hardcoding here, we can make this a class variable that defaults to 1?

fl4health/losses/deep_mmd_loss.py
emersodb: It looks like we don't leverage the fact that y can be none to get the distances of x with itself. Maybe we just drop that option and require y to be passed to simplify this function.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
+from typing import Optional, Tuple
```
emersodb, line 1: I think it's worth creating some tests for this loss function before doing any of the refactors I suggest below just to make sure we don't inadvertently break anything in the calculations. It could be as simple as generating a random X, Y and recording the outputs of each function. That way we know we did something if those outputs change with our refactors.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+                -Dyy_org / sigma
+            )
+            Kxy = (1 - epsilon) * torch.exp(-((Dxy / sigma0) ** L) - Dxy_org / sigma) + epsilon * torch.exp(
+                -Dxy_org / sigma
+            )
+        else:
+            Kx = torch.exp(-Dxx / sigma0)
+            Ky = torch.exp(-Dyy / sigma0)
+            Kxy = torch.exp(-Dxy / sigma0)
+
```
emersodb, line 149: Maybe here just add a comment noting what Kx, Ky, Kxy represent. That is, they are $k_w(x_i, x_j)$, $k_w(y_i, y_j)$ and $k_w(x_i,y_j)$ for all i, j in the samples of X and Y

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
emersodb: I also might be missing this, but I don't see where we set `layer_loss_module.training` to `True` in the client. Based on the loss code, this would mean that we won't run training of the deep kernels after the first server round, which I think we want to keep doing?

fl4health/losses/deep_mmd_loss.py
emersodb: Perhaps we can also name these to be a bit more verbose. That is, `Kx` -> `kernel_xx` `K_y` -> `kernel_yy` and `K_xy` -> `kernel_xy`. We could do the same for `Dx`, `Dy`, `Dxy` -> `distance_xx`, `distance_yy`, `distance_xy`

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        Pdist[Pdist < 0] = 0
+        return Pdist
+
+    def h1_mean_var_gram(
+        self, Kx: torch.Tensor, Ky: torch.Tensor, Kxy: torch.Tensor, is_var_computed: bool, use_1sample_U: bool = True
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor], torch.Tensor]:
+        """compute value of MMD and std of MMD using kernel matrix."""
+        Kxxy = torch.cat((Kx, Kxy), 1)
+        Kyxy = torch.cat((Kxy.transpose(0, 1), Ky), 1)
+        Kxyxy = torch.cat((Kxxy, Kyxy), 0)
```
emersodb, line 81: It's worth a comment that this final matrix (`Kxyxy`) is composed of the blocks [[`Kx`, `Kxy`], [`Kxy^T`, `Ky`]]. Just to help any reader understand what's actually being constructed here.

emersodb: Actually, it seems like this tensor isn't used anywhere, which would make its formation unnecessary. Maybe I'm missing something though?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        nx = Kx.shape[0]
+        ny = Ky.shape[0]
+        is_unbiased = True
+        if is_unbiased:
+            xx = torch.div((torch.sum(Kx) - torch.sum(torch.diag(Kx))), (nx * (nx - 1)))
```
emersodb, line 86: Maybe note that these calculations correspond to the calculation of $\hat{\text{MMD}}_^2$ in equation 2 of the paper?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            torch.nn.Linear(H, x_out, bias=True),
+        )
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        """Forward the LeNet."""
+        fealant = self.latent(input)
+        return fealant
+
+
+class DeepMmdLoss(torch.nn.Module):
```
emersodb, line 30: I think it's worth recording the paper that this is being taken from somewhere here 🙂

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        nx = Kx.shape[0]
+        ny = Ky.shape[0]
+        is_unbiased = True
+        if is_unbiased:
+            xx = torch.div((torch.sum(Kx) - torch.sum(torch.diag(Kx))), (nx * (nx - 1)))
+            yy = torch.div((torch.sum(Ky) - torch.sum(torch.diag(Ky))), (ny * (ny - 1)))
+            # one-sample U-statistic.
+            if use_1sample_U:
+                xy = torch.div((torch.sum(Kxy) - torch.sum(torch.diag(Kxy))), (nx * (ny - 1)))
+            else:
```
emersodb, line 91: I'm having trouble figuring out where the calculation below is described in their paper. It doesn't seem like it's a well defined calculation to be honest. I'm tempted to suggest we remove it and only offer the `is_unbiased` boolean to switch between the estimates $\hat{\text{MMD}}_u^2$ and $\hat{\text{MMD}}_b^2$ (which is what is computed below)

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+                xy = torch.div(torch.sum(Kxy), (nx * ny))
+            mmd2 = xx - 2 * xy + yy
```
emersodb, line 93: This calculation can be moved outside the if and else blocks rather than being repeated in both blocks.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            yy = torch.div((torch.sum(Ky)), (ny * ny))
+            # one-sample U-statistic.
+            if use_1sample_U:
+                xy = torch.div((torch.sum(Kxy)), (nx * ny))
+            else:
+                xy = torch.div(torch.sum(Kxy), (nx * ny))
+            mmd2 = xx - 2 * xy + yy
+        if not is_var_computed:
+            return mmd2, None, Kxyxy
+        hh = Kx + Ky - Kxy - Kxy.transpose(0, 1)
```
emersodb, line 105: To be more consistent with the paper notation, I'd rename `hh` to `h_ij`?

emersodb: I'd add a comment above that we're computing the estimate in Equation (5) from the paper.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        V1 = torch.dot(hh.sum(1) / ny, hh.sum(1) / ny) / ny
```
emersodb, line 106: This is a weird way to write this. My suggestion, to make it look more like Equation (5) from the paper is to do
```
V1 = (4.0/ny**3)*(torch.dot(hh.sum(1), hh.sum(1)))
V2 = (4.0/nx**4)*(hh.sum()**2)
variance_estimate = V1 - V2

File: fl4health/losses/deep_mmd_loss.py [hunk 38]
```diff
@@ -0,0 +1,224 @@
...
+        self.sigmaOPT.requires_grad = True
+        self.sigma0OPT.requires_grad = True
+        self.epsilonOPT.requires_grad = True
+
+        features = torch.cat([X, Y], 0)
+
+        # ------------------------------
+        #  Train deep network for MMD-D
+        # ------------------------------
+        # Initialize optimizer
```
emersodb, line 165: This isn't really initializing the optimizer. It's just clearing the gradients.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            len_s=X.shape[0],
+            Fea_org=features.view(features.shape[0], -1),
+            sigma=sigma,
+            sigma0=sigma0_u,
+            epsilon=ep,
+        )
+        if mmd_var_temp is None:
+            raise AssertionError("Error: Variance of MMD is not computed. Please set is_var_computed=True.")
+        mmd_std_temp = torch.sqrt(mmd_var_temp + 10 ** (-8))
+        STAT_u = torch.div(-1 * mmd_value_temp, mmd_std_temp)
```
emersodb, line 185: I'm okay with calling this `stat_u`, but I would kill the all caps 🙂.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        # ------------------------------
+        # Initialize optimizer
+        self.optimizer_F.zero_grad()
+        # Compute output of deep network
+        model_output = self.featurizer(features)
+        # Compute epsilon, sigma and sigma_0
+        ep = torch.exp(self.epsilonOPT) / (1 + torch.exp(self.epsilonOPT))
+        sigma = self.sigmaOPT**2
+        sigma0_u = self.sigma0OPT**2
+        # Compute Compute J (STAT_u)
```
emersodb, line 173: I'd add to this comment that what we're forming is the ratio in Equation (4) from the paper in `stat_u` which is referred to as J in therein.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        mmd_value_temp, mmd_var_temp, _ = self.MMDu(
```
emersodb, line 174: I'm not sure why they are suffixing with `_temp` here. Maybe we just call it `..._estimate`?

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+            Fea=model_output,
+            len_s=X.shape[0],
+            Fea_org=features.view(features.shape[0], -1),
+            sigma=sigma,
+            sigma0=sigma0_u,
+            epsilon=ep,
+        )
+        if mmd_var_temp is None:
+            raise AssertionError("Error: Variance of MMD is not computed. Please set is_var_computed=True.")
+        mmd_std_temp = torch.sqrt(mmd_var_temp + 10 ** (-8))
```
emersodb, line 184: I'm good with doing the sqrt here, since its how we form $\hat{J}_{\lambda}$. However, I'd move the lambda shift here into the formation of the estimate to collocate it better with the estimate calculation. So this would become something like 
```
mmd_std_estimate = torch.sqrt(mmd_var_estimate)
```

fl4health/losses/deep_mmd_loss.py
emersodb: Discussed in a comment below, but I would move the addition of the lambda shift into this calculation to better align it with Equation (5) from the paper. That is
```
variance_estimate = V1 - V2 + (10 ** (-8))
```

fl4health/losses/deep_mmd_loss.py
emersodb: I would add a comment that we're trying to maximize the ratio (mmd_value_stimate/mmd_std_estimate), so we multiply by -1 to be compatible with gradient descent.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        STAT_u = torch.div(-1 * mmd_value_temp, mmd_std_temp)
+        # Compute gradient
+        STAT_u.backward()
+        # Update weights using gradient descent
+        self.optimizer_F.step()
+        return
```
emersodb, line 190: Return is unnecessary here.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        features = torch.cat([X, Y], 0)
+
+        # Compute output of deep network
+        model_output = self.featurizer(features)
+        # Compute epsilon, sigma and sigma_0
+        ep = torch.exp(self.epsilonOPT) / (1 + torch.exp(self.epsilonOPT))
+        sigma = self.sigmaOPT**2
+        sigma0_u = self.sigma0OPT**2
+        # Compute Compute J (STAT_u)
+        mmd_value_temp, _, _ = self.MMDu(
```
emersodb, line 209: As with the training function, I'd suggest renaming this to `mmd_value_estimate` from `_temp`

research/flamby/fed_isic2019/ditto_deep_mmd/client.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        super().__init__(
+            data_path=data_path,
+            metrics=metrics,
+            device=device,
+            loss_meter_type=loss_meter_type,
+            checkpointer=checkpointer,
+            lam=lam,
+            deep_mmd_loss_weight=deep_mmd_loss_weight,
+            flatten_feature_extraction_layers={key: True for key in size_feature_extraction_layers},
```
emersodb, line 57: Maybe also form this dictionary above the super call as you're doing with `size_feature_extraction_layers`?

research/flamby/fed_isic2019/ditto_deep_mmd/client.py
```diff
@@ -0,0 +1,200 @@
...
+        action="store",
+        type=int,
+        help="Seed for the random number generators across python, torch, and numpy",
+        required=False,
+    )
+    parser.add_argument(
+        "--mu",
+        action="store",
+        type=float,
+        help="Weight for the mkmmd losses",
```
emersodb, line 141: Deep MMD losses 🙂

research/flamby/fed_isic2019/ditto_deep_mmd/client.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    parser.add_argument(
+        "--deep_mmd_loss_depth",
+        action="store",
+        type=int,
+        help="Depth of applying the deep mmd loss",
+        required=False,
+        default=1,
+    )
+    parser.add_argument(
```
emersodb, line 159: Similarly, I don't think this arg is used anywhere either

research/flamby/fed_isic2019/ditto_deep_mmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,169 @@
...
+#
+# sbatch research/flamby/fed_isic2019/ditto_deep_mmd/run_fold_experiment.slrm \
+#   path_to_config.yaml \
+#   path_to_folder_for_artifacts/ \
+#   path_to_folder_for_dataset/ \
+#   path_to_desired_venv/ \
+#   client_side_learning_rate_value \
+#   lambda value \
+#   mu value \
+#   depp_mmd_loss_depth value \
```
emersodb, line 27: Johnny Depp MMD Loss!

research/flamby/fed_isic2019/ditto_deep_mmd/run_fold_experiment.slrm
```diff
@@ -0,0 +1,169 @@
...
+#
+# Example:
+# sbatch research/flamby/fed_isic2019/ditto_deep_mmd/run_fold_experiment.slrm \
+#   research/flamby/fed_isic2019/ditto_deep_mmd/config.yaml \
+#   research/flamby/fed_isic2019/ditto_deep_mmd/hp_results/ \
+#   /Users/david/Desktop/FLambyDatasets/fedisic2019/ \
+#   /h/demerson/vector_repositories/fl4health_env/ \
+#   0.0001 \
+#   0.01 \
+#   0.1 \
```
emersodb, line 38: I think the depth argument is missing in this example?

fl4health/losses/deep_mmd_loss.py
sanaAyrml: coming back to this after two months, I see the weirdness much more 🤣 looool

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -0,0 +1,201 @@
...
+        device: torch.device,
+        loss_meter_type: LossMeterType = LossMeterType.AVERAGE,
+        checkpointer: Optional[ClientCheckpointModule] = None,
+        lam: float = 1.0,
+        deep_mmd_loss_weight: float = 10.0,
+        flatten_feature_extraction_layers: Dict[str, bool] = {},
+        size_feature_extraction_layers: Dict[str, int] = {},
+    ) -> None:
+        """
+        This client implements the DEEP-MMD loss function in the Ditto framework.
```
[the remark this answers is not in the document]
sanaAyrml: I changed it to Deep MMD.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: I ended up omitting one of the parameters since we decided to consistently use flattened features for MK-MMD losses. Now, in the init function, I only have `size_feature_extraction_layers` and set `self.flatten_feature_extraction_layers`, with flatten set to True for all layers.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: Good catch! The `True` setting was indeed missing, so I added it to the `update_before_train` function. Following your suggestion, I moved the `False` setting to the `validate` function. I kept the assertions in both `compute_evaluation_loss` and `compute_training_loss` functions for consistency.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: This was the original one but sure I can change if you think it will be useful.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: Yes exactly that's what we are doing. We first update kernel for each batch then compute value.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: I see I see, indeed it looks completely weird and useless as we do not use it originally.

sanaAyrml: I agree that it is better to drop as it is not useful and it does not make any sense for me too.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: Done

fl4health/losses/deep_mmd_loss.py
sanaAyrml: That is correct

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: Soo now after doing that, the optimization of kernel parameters and model will happen in one step but with different optimizers, however there is a potential second approach can be to separate these optimizations. Meaning that during training for each batch first we freeze model and optimize kernel, then freeze kernel and optimize model.

research/flamby/fed_isic2019/ditto_deep_mmd/client.py
```diff
@@ -0,0 +1,200 @@
...
+        required=False,
+    )
+    parser.add_argument(
+        "--mu",
+        action="store",
+        type=float,
+        help="Weight for the mkmmd losses",
+        required=False,
+    )
+    parser.add_argument(
```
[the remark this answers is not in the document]
sanaAyrml: Deep MMD loss depth is used in FedIsic2019DittoClient class.

research/flamby/fed_isic2019/ditto_deep_mmd/client.py
sanaAyrml: True this is extra argument.

research/flamby/fed_isic2019/ditto_deep_mmd/run_fold_experiment.slrm
sanaAyrml: Hahahah that was so funny 😂

fl4health/losses/deep_mmd_loss.py
sanaAyrml: I did that and also added a simple test for Deep MMD loss to make sure everything works as expected.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,224 @@
...
+        if is_unbiased:
+            xx = torch.div((torch.sum(Kx) - torch.sum(torch.diag(Kx))), (nx * (nx - 1)))
+            yy = torch.div((torch.sum(Ky) - torch.sum(torch.diag(Ky))), (ny * (ny - 1)))
+            # one-sample U-statistic.
+            if use_1sample_U:
+                xy = torch.div((torch.sum(Kxy) - torch.sum(torch.diag(Kxy))), (nx * (ny - 1)))
+            else:
+                xy = torch.div(torch.sum(Kxy), (nx * ny))
+            mmd2 = xx - 2 * xy + yy
+        else:
```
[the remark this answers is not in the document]
emersodb: It's been a while since I left this comment but I think we are computing the very specific notation above. I think it's worth being precise (if I was current in the measure being computed), that is \hat{\text{MMD}}_b^2 in your comment.

fl4health/losses/deep_mmd_loss.py
emersodb: It doesn't look like we did this? I think both the rename and comment are worthwhile

fl4health/losses/deep_mmd_loss.py
```diff
@@ -0,0 +1,250 @@
...
+        k_y: torch.Tensor,
+        k_xy: torch.Tensor,
+        is_var_computed: bool,
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+        """compute value of MMD and std of MMD using kernel matrix."""
+        nx = k_x.shape[0]
+        ny = k_y.shape[0]
+
+        if self.is_unbiased:
+            # compute the unbiased MMD estimator (MMD_u^2) defined in Eq. (2) of the paper
```
emersodb, line 109: I think we want to be specific with what mmd is being calculated to match the paper, that is \hat{\text{MMD}}_u^2 (that is, which the hat).

fl4health/losses/deep_mmd_loss.py
emersodb: I'd include the notation mention in your comment as well if you're alright with it (\hat{J}_{\lambda})

emersodb (commented):
Really nice changes. Just added a few small comments and reminders of a few pieces you might have overlooked in my comments. Very close to ready to go!

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -24,119 +24,167 @@ def __init__(
...
-        self.deep_mmd_losses = {}
-        for layer in self.flatten_feature_extraction_layers.keys():
-            self.deep_mmd_losses[layer] = DeepMmdLoss(
-                device=self.device,
-                input_size=self.size_feature_extraction_layers[layer],
-                layer_name=layer,
-            ).to(self.device)
-
-        self.init_global_model: nn.Module
+        if feature_extraction_layers_with_size:
```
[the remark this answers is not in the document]
sanaAyrml: Done

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -174,22 +221,20 @@ def compute_loss_and_additional_losses(
...
-                    features[layer], features[" ".join(["init_global", layer])]
-                )
-                additional_losses["_".join(["deep_mmd_loss", layer])] = layer_deep_mmd_loss
-                total_deep_mmd_loss += layer_deep_mmd_loss
+            for layer, layer_deep_mmd_loss in self.deep_mmd_losses.items():
+                deep_mmd_loss = layer_deep_mmd_loss(features[layer], features[" ".join(["init_global", layer])])
+                additional_losses["_".join(["deep_mmd_loss", layer])] = deep_mmd_loss
+                total_deep_mmd_loss += deep_mmd_loss
             total_loss += self.deep_mmd_loss_weight * total_deep_mmd_loss
             additional_losses["deep_mmd_loss_total"] = total_deep_mmd_loss
```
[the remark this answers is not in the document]
sanaAyrml: I added that but I am checking bunch of other ditto versions and we don't have any where. I am wondering whether I should update them or not.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: oh I see I updated to full notation now.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: Yeah I missed this

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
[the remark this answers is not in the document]
emersodb: Discussed together.

emersodb (approved):
I think this looks ready to go! 

I think the only open question we still have is how we get around that exception with the repeated backward call.

PR #172, opened 2024-06-13
Fix some bugs that slipped through in CR and trying to make type chec…

jewelltaylor:
…ker shutup

# PR Type

# Short Description

Nice job @scarere for already finding some bugs! 

Small changes to address indexing we are indexing into lists of args from parser (instead of tuples) and to resolve an issue called by using stem when extensions may have multiple "." in them. Was able to run the preprocessing on the cluster so can confirm it works. 

# Tests Added
NA


emersodb (approved):
Good to go!

PR #173, opened 2024-06-14
FedSimCLR

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [FedSimClr Implementation](https://app.clickup.com/t/8688tfmpd)

This PR adds an implementation of an extension of the Self Supervised Learning method [SimCLR](https://arxiv.org/abs/2002.05709) to the Federated Learning Setting presented in [Fed-X](https://arxiv.org/pdf/2207.09158). The example includes pretraining and finetuning on the CIFAR10 dataset. It is rather simple and I will include more complex examples in the future. The additions in the PR include: 
- A model base that groups an encoder, projection head and prediction head
- An implementation of the contrastive loss used by SimCLR NT-Xent
- An addition of a method to client classes that allows the user to specify a transformation applied to the target. In this case we need to run the transformed input (target) through the model to obtain a representation. 

Happy to hear any feedback on the implementation because I was exploring several potential options. I wanted to avoid having a specific client so it can be used with multiple server aggregation strategies. This can be accomplished with multiple inheritance but it may be burdensome for the user to be expected to reconcile different train_step methods if they collide.

Also as part of this PR I made some changes to the BaseDataset that I want to build on in the future to support having samplers defined not only datasets that store the entire input and target tensors in memory which is rather limiting. For our current datasets that make those assumptions, I have introduced a TensorDataset Class. In the follow up PR I will work to add a File Dataset that is compatible with Label Based Samplers. 

Lastly, I also fixed a small bug where there was crossover between validation and test sets for CIFAR. In doing so, I changed the datasets to split validation out of train instead of using test set for validation. Because of this, I had to update the smoke tests because the numbers changed (lower number of training samples per client). 

# Tests Added

 - Tests for contrastive loss


examples/fedsimclr_example/fedsimclr_finetuning_example/client.py
```diff
@@ -0,0 +1,73 @@
...
+from examples.models.cnn_model import SslEncoder, SslPredictionHead, SslProjectionHead
+from fl4health.clients.basic_client import BasicClient
+from fl4health.model_bases.fedsimclr_base import FedSimClrModel
+from fl4health.utils.dataset import TensorDataset
+from fl4health.utils.load_data import get_cifar10_data_and_target_tensors, split_data_and_targets
+from fl4health.utils.metrics import Accuracy
+
+
+def get_finetune_dataset(data_dir: Path, batch_size: int) -> Tuple[DataLoader, DataLoader]:
+    data, targets = get_cifar10_data_and_target_tensors(data_dir, False)
```
emersodb, line 23: very minor but maybe we keyword the train argument here to make it explicit that the boolean is saying load the test data? That is, `train=False`.

fl4health/losses/contrastive_loss.py
```diff
@@ -86,3 +87,75 @@ def forward(
...
+        similarity_ij = torch.diag(similarity_matrix, diagonal=batch_size)
+        similarity_ji = torch.diag(similarity_matrix, diagonal=-batch_size)
+        positives = torch.concatenate([similarity_ij, similarity_ji], dim=0)
+
+        # Numerator is the sum of the exponent of positive similarities
+        numerator = torch.exp(positives / self.temperature)
+
+        # Denominator is all pair combinations except for diagonal which corresponds to a features similarity to itself
+        mask = (torch.ones(2 * batch_size, 2 * batch_size) - torch.eye(2 * batch_size)).to(self.device)
+        similarity_matrix_without_diagonal = torch.mul(similarity_matrix, mask)
```
emersodb, line 155: This will include the positive pair values. I think that's what you want to do, but wanted to confirm?

fl4health/losses/contrastive_loss.py
```diff
@@ -86,3 +87,75 @@ def forward(
...
+        denominator = torch.exp(similarity_matrix_without_diagonal / self.temperature)
+
+        # Final loss negative log likelihood
+        losses = -torch.log(numerator / denominator.sum(dim=1))
+        loss = torch.sum(losses) / (2 * batch_size)
```
emersodb, line 160: Maybe a comment here that we're dividing by 2*batch_size since pairs are double counted due to the similarity matrix symmetry?

fl4health/losses/contrastive_loss.py
```diff
@@ -86,3 +87,75 @@ def forward(
         labels = torch.zeros(features.size(0)).to(self.device).long()
 
         return self.cross_entropy_function(logits, labels)
+
+
+class NtXentLoss(nn.Module):
+    def __init__(self, device: torch.device, temperature: float = 0.05) -> None:
```
emersodb, line 93: Maybe this is a reasonable default (`0.05`) for the temperature, but it seems super low 😂

fl4health/model_bases/fedsimclr_base.py
```diff
@@ -0,0 +1,44 @@
...
+                output of encoder to prediction in the finetuning stage.
+                Defaults to None.
+            pretrain (bool): Determines whether or not to use the projection_head
+                (True) or the prediction_head (False). Defaults to True.
+        """
+        super().__init__()
+        self.encoder = encoder
+        self.projection_head = projection_head
+        self.prediction_head = prediction_head
+        self.pretrain = pretrain
```
emersodb, line 36: It might also be worth throwing an error here if `self.pretrain == False` and `self.prediction_head is None`

fl4health/utils/dataset.py
```diff
@@ -39,36 +24,59 @@ def update_target_transform(self, g: Callable) -> None:
...
-        cifar10_dataset = CIFAR10(data_path, train=train, download=True)
-        self.data = torch.from_numpy(cifar10_dataset.data)
-        self.targets = torch.Tensor(cifar10_dataset.targets).long()
+        data: torch.Tensor,
+        targets: Optional[torch.Tensor] = None,
+        transform: Optional[Callable] = None,
+        target_transform: Optional[Callable] = None,
+    ) -> None:
+        super().__init__()
+        self.data: torch.Tensor = data
```
emersodb, line 45: I don't think we need the annotations for self.data or self.targets, given the typing on the inputs?

fl4health/utils/dataset.py
```diff
@@ -39,36 +24,59 @@ def update_target_transform(self, g: Callable) -> None:
...
+        self.targets: Optional[torch.Tensor] = targets
         self.transform = transform
         self.target_transform = target_transform
 
+    def __getitem__(self, index: int) -> Tuple[torch.Tensor, torch.Tensor]:
+        assert self.targets is not None
+
+        data, target = self.data[index], self.targets[index]
+
+        if self.transform is not None:
```
emersodb, line 55: This assumes that any transforms that have been defined would initially operate on a numpy array, which I think is a little confusing, since data and target are torch.Tensors. I'd say maybe kill the numpy conversion and do it as part of any transforms that are set?

fl4health/utils/load_data.py
```diff
@@ -1,85 +1,138 @@
...
 
-from fl4health.utils.dataset import BaseDataset, Cifar10Dataset, MnistDataset
+from fl4health.utils.dataset import TensorDataset
 from fl4health.utils.dataset_converter import DatasetConverter
+from fl4health.utils.random import set_all_random_seeds
 from fl4health.utils.sampler import LabelBasedSampler
 
 
+def split_data_and_targets(
+    data: torch.Tensor, targets: torch.Tensor, validation_proportion: float = 0.02
```
emersodb, line 19: `0.02` as a default proportion is very small, maybe you meant to write `0.2`?

fl4health/utils/load_data.py
```diff
@@ -1,85 +1,138 @@
...
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+
+    total_size = data.shape[0]
+    train_size = int(total_size * (1 - validation_proportion))
+    train_indices = random.sample(range(total_size), train_size)
+    val_indices = [i for i in range(total_size) if i not in train_indices]
+    train_data, train_targets = data[train_indices, ...], targets[train_indices, ...]
```
emersodb, line 26: I'm not sure I've seen the `,...` notation before? I assume you're just saying take the whole of the tensor indexed by `train_indices`. Is there a reason to use that over just doing `data[train_indices]` and `targets[train_indices]`?

fl4health/utils/load_data.py
```diff
@@ -1,85 +1,138 @@
...
+    transform: Optional[Callable] = None,
+    target_transform: Optional[Callable] = None,
     dataset_converter: Optional[DatasetConverter] = None,
+    validation_proportion: float = 0.2,
+    seed: int = 42,
 ) -> Tuple[DataLoader, DataLoader, Dict[str, int]]:
     """Load MNIST Dataset (training and validation set)."""
     log(INFO, f"Data directory: {str(data_dir)}")
+
+    set_all_random_seeds(seed)
```
emersodb, line 66: I think, but I'm not 100% sure, that this will actually set a global seed, not just for the loading and preparing of the data. It will also override any other seeds that someone has set (i.e. in other parts of their flow). So rather, I think we strictly want to seed locally here. That is, the seed should be optional and none by default, so that the user **has to** explicitly set the seed again if they want it. If it's set, we only set the seeds for local generators that are necessary for reproducibility of the things like splitting and sampling.

fl4health/utils/load_data.py
```diff
@@ -1,85 +1,138 @@
...
+    data_dir: Path,
+    batch_size: int,
+    sampler: Optional[LabelBasedSampler] = None,
+    validation_proportion: float = 0.2,
+    seed: int = 42,
 ) -> Tuple[DataLoader, DataLoader, Dict[str, int]]:
     """Load CIFAR-10 (training and validation set)."""
     log(INFO, f"Data directory: {str(data_dir)}")
+
+    set_all_random_seeds(seed)
```
emersodb, line 127: Same comment about the seeds here. That is, we need to only do it locally and have the provided seed be optional and none by default.

fl4health/losses/contrastive_loss.py
```diff
@@ -86,3 +87,75 @@ def forward(
...
+        """
+        super().__init__()
+        self.device = device
+        self.temperature = temperature
+
+    def forward(self, features: torch.Tensor, transformed_features: torch.Tensor) -> torch.Tensor:
+        """
+        Compute the contrastive loss based on the features and transformed_features. Given N features
+        and N transformed_features per batch, features[i] and transformed_features[i] are positive pairs
+        and the remaining 2N - 2 are negative pairs.
```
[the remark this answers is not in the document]
jewelltaylor: Not at all! So we have N inputs and then N transformed versions of the inputs that we consider targets. So in each batch that is 2N samples. For each sample, one of the other samples is positive. This leaves the remaining 2N - 2 samples as negatives. 2N - 2 because we are subtracting out the sample and its positive pair. Let me know if I can clarify the documentation at all!

fl4health/losses/contrastive_loss.py
jewelltaylor: Yup! Since we are doing a softmax essentially, we want the positive pairs to be included as well. Does that sound right?

jewelltaylor:
I th

> Most of the changes look great I think. The only really important thing that I noted is the setting of the random seeds for dataset loading, since they affect scope well outside of the loading itself.

Thanks for the great review! I think I have addressed all the comments you made

fl4health/model_bases/fedsimclr_base.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import annotations
```
emersodb, line 1: Is this used anywhere? I don't see where it is but I might be missing it.

jewelltaylor: This is used to be able to return an object of type class from a method in class https://docs.python.org/3/library/__future__.html

fl4health/losses/contrastive_loss.py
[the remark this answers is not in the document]
emersodb: Ah right. I was thinking of the whole similarity matrix minus the diagonal, but that's not quite right. This wording is fine 🙂

fl4health/losses/contrastive_loss.py
emersodb: Yeah, I see now in the way we're forming the denominator of the contrastive loss.

fl4health/model_bases/fedsimclr_base.py
emersodb: Oh cool. I didn't know that this existed! I've always just used the mypy escape mechanism of annotating in quotes as `-> "FedSimClrModel"`. I'll put a ticket in the backlog to add this in a few places where this is a little cleaner!

jewelltaylor: Yeah! Eventually this will be part of the default functionality as far as I understand. They were supposed to have that in python 3.10 but it got pushed so you still have to use the from __future__

emersodb (approved):
Good to go!

PR #179, opened 2024-06-18
Fix picai channel ordering

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Fix Picai Channel Ordering](https://app.clickup.com/t/8688u62qb)

The channel of the scans (t2w, adc, hbv) are being written with incorrect file names. The root of the cause was the write method of PicaiCase. The scan paths were being sorted when they should not. They are being passed with an assumed order. I also added to the documentation to make this more clear. 

# Tests Added

- Test to ensure we reading and writing picai case correctly

### DIFF
--- a/research/picai/preprocessing.py
+++ b/research/picai/preprocessing.py
@@ -120,12 +120,14 @@ class PicaiCase(Case):
         Class representing a case from the PICAI dataset.
 
         scan_paths filenames are assumed to have the following format: <patient_id>_<study_id>_<modality>.mha
-        where modality is a three letter string of ['t2w', 'adc', 'hbv']
+        where modality is a three letter string of ['t2w', 'adc', 'hbv']. Assumes the order of scan paths is
+        t2w, adc and hbv.
 
         annotation_path filename is assumed to have the following format: <patient_id>_<study_id>.nii.gz
 
         Args:
             scan_paths (Sequence[Path]): The set of paths where the scans associated with the Case are located.
+                Assumes the order of scan paths is t2w, adc, hbv.
             annotation_write_dir (Path): The path where the annotation associated with the Case is located.
             settings (PreprocessingSettings): The settings determining how the case is preprocessed.
         """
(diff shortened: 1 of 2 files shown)


research/picai/preprocessing.py
```diff
@@ -120,12 +120,14 @@ def __init__(
         Class representing a case from the PICAI dataset.
 
         scan_paths filenames are assumed to have the following format: <patient_id>_<study_id>_<modality>.mha
-        where modality is a three letter string of ['t2w', 'adc', 'hbv']
+        where modality is a three letter string of ['t2w', 'adc', 'hbv']. Assumes the order of scan paths is
+        t2w, adc and hbv.
```
emersodb, line 124: As discussed briefly in the call, I think we can make this comment slightly more clear in the sense that the order of the scan paths doesn't necessarily have to follow a specific structure but it MUST be consistent with the ordering of `self.scans` throughout. So something like 
```
where modality is a three letter string of ['t2w', 'adc', 'hbv']. NOTE: the ordering self.scan_path and self.scans must remain consistent.
```

research/picai/preprocessing.py
```diff
@@ -120,12 +120,14 @@ def __init__(
...
 
         annotation_path filename is assumed to have the following format: <patient_id>_<study_id>.nii.gz
 
         Args:
             scan_paths (Sequence[Path]): The set of paths where the scans associated with the Case are located.
+                Assumes the order of scan paths is t2w, adc, hbv.
```
emersodb, line 130: I think we can just say 
```
NOTE: self.scans will inherit the ordering of scan_paths and must remain consistently ordered.
```

tests/test_picai/test_case.py
```diff
@@ -0,0 +1,63 @@
...
+
+import numpy as np
+import SimpleITK as sitk
+
+from research.picai.preprocessing import PicaiCase, PreprocessingSettings
+
+
+def test_read_and_write_picai_case() -> None:
+    X, Y, Z = 32, 64, 128
+    with tempfile.TemporaryDirectory() as results_dir:
```
emersodb, line 13: I think this is fine, but pytest actually offers a temporary file system that it cleans up itself for this type of thing. You can see it in `tests/checkpointing/test_save_load.py`. 
```
def test_save_and_load_best_loss_checkpoint(tmp_path: Path) -> None:
    # Temporary path to write pkl to, will be cleaned up at the end of the test.
    checkpoint_dir = tmp_path.joinpath("resources")
    checkpoint_dir.mkdir()
    ...
```
I think it's a little cleaner. At the very least it saves some indents 😂 , but if there is a reason you like this better, I'm happy to hear it 🙂

tests/test_picai/test_case.py
```diff
@@ -0,0 +1,63 @@
...
+from pathlib import Path
+
+import numpy as np
+import SimpleITK as sitk
+
+from research.picai.preprocessing import PicaiCase, PreprocessingSettings
+
+
+def test_read_and_write_picai_case() -> None:
+    X, Y, Z = 32, 64, 128
```
emersodb, line 12: Very minor, but I'd maybe make these a little small to reduce overhead slightly. Maybe 8, 16, 32?

emersodb (approved):
Fix and test looks good! Just some basic suggestions

PR #181, opened 2024-06-19
Update MK-MMD Loss Suggestions and Tests

emersodb:
# PR Type
Other

# Short Description

This PR adds tests for the MK-MMD loss updates suggested in the target PR. The tests modify both the original MK-MMD tests associated with the linear approximation and adds tests to cover the full approximation calculations constructed in the update. 

I also did some structural refactors to sort of "split" the linear vs. full approximation processing where I thought it made sense. Let me know if you disagree with any of it.

# Tests Added

See above.

### DIFF
--- a/fl4health/losses/mkmmd_loss.py
+++ b/fl4health/losses/mkmmd_loss.py
@@ -1,6 +1,3 @@
-# multi-kernel maximum mean discrepancy
-# cao bin, HKUST, China, binjacobcao@gmail.com
-# free to charge for academic communication
 from logging import INFO
 from typing import Optional
 
@@ -18,7 +15,7 @@ class MkMmdLoss(torch.nn.Module):
         minimize_type_two_error: bool = True,
         normalize_features: bool = False,
         layer_name: Optional[str] = None,
-        construct_quadruples_per_kernel: bool = False,
+        perform_linear_approximation: bool = False,
     ) -> None:
         """
         Compute the multi-kernel maximum mean discrepancy (MK-MMD) between the source and target domains. Also allows
@@ -38,6 +35,9 @@ class MkMmdLoss(torch.nn.Module):
             normalize_features (Optional[bool], optional): Whether to normalize the features to have unit length before
                 computing the MK-MMD and optimizing betas. Defaults to False.
             layer_name (Optional[str], optional): The name of the layer to extract features from. Defaults to None.
+            perform_linear_approximation (Optional[bool], optional): Whether to use linear approximations for the
+                estimates of the mean and covariance of the kernel values. Experimentally, we have found that the
+                linear approximations largely hinder the statistical power of Mk-MMD. Defaults to False
         """
         super().__init__()
         self.device = device
@@ -61,7 +61,10 @@ class MkMmdLoss(torch.nn.Module):
         self.minimize_type_two_error = minimize_type_two_error
         self.normalize_features = normalize_features
         self.layer_name = layer_name
-        self.construct_quadruples_per_kernel = construct_quadruples_per_kernel
+        self.perform_linear_approximation = perform_linear_approximation
+
+    def normalize(self, X: torch.Tensor) -> torch.Tensor:
+        return torch.div(X, torch.linalg.norm(X, dim=1, keepdim=True))
 
     def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor:
         """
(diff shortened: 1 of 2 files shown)


fl4health/losses/mkmmd_loss.py
```diff
@@ -1,6 +1,3 @@
-# multi-kernel maximum mean discrepancy
-# cao bin, HKUST, China, binjacobcao@gmail.com
-# free to charge for academic communication
```
emersodb, line 3: We departed from this a fair bit ago...so I think we can remove the attribution 😂

fl4health/losses/mkmmd_loss.py
```diff
@@ -61,7 +61,10 @@ def __init__(
         self.minimize_type_two_error = minimize_type_two_error
         self.normalize_features = normalize_features
         self.layer_name = layer_name
-        self.construct_quadruples_per_kernel = construct_quadruples_per_kernel
+        self.perform_linear_approximation = perform_linear_approximation
+
+    def normalize(self, X: torch.Tensor) -> torch.Tensor:
```
emersodb, line 66: Just moved this to the top.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
             - 2.0 * torch.mm(X, torch.transpose(Y, 0, 1))
         )
-        x_y_prime[x_y_prime < 0] = 0
         x_prime_y = (
             torch.sum((Y**2), dim=1).reshape(-1, 1)
             + torch.sum((X**2), dim=1).reshape(1, -1)
             - 2.0 * torch.mm(Y, torch.transpose(X, 0, 1))
         )
+
+        # Correct any values that ended up nearly but not identically zero (||x-y||^2 should be semi-definite)
```
emersodb, line 110: Just grouped these guys together for a comment explaining what's happening.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
 
+    def compute_euclidean_inner_products_linear(self, v_i_quadruples: torch.Tensor) -> torch.Tensor:
+        # Shape of v_i_quadruples is n_samples/2 x 4 x n_features
+        # v_i = [x_{2i-1}, x_{2i}, y_{2i-1}, y_{2i}]
+        #
+        # We want to compute the RBF kernel values. To do this, we need to compute ||x - y||^2 for the relevant pairs
+        # x and y. That is the inner product. Note that ||x - y||^2 = <x - y, x - y> = (x-y)^T(x-y)
+        # For the quadruples of the form (x,  x', y, y') we need distances for pairings (x, x'), (y, y'), (x, y'),
+        # (x, y')
+        x_x_prime = torch.sum((v_i_quadruples[:, 0, :] - v_i_quadruples[:, 1, :]) ** 2, dim=1, keepdim=True)
```
emersodb, line 130: I went back to the torch.sum here rather than doing the quadratic expansion that you had. This is the "true" linear approximation suggested by Gretton. It doesn't seem to work 🤦, but we want to implement it faithfully. I removed the torch.mean from the original implementation because that was dumb of me. It should just be sum.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
+        # each inner product is a tensor of dimension len(v_i_quadruples), we return a tensor of shape
+        # len(v_i_quadruples) x 4
+        return torch.cat([x_x_prime, y_y_prime, x_y_prime, x_prime_y], dim=1)
+
     def compute_h_u_from_inner_products(self, inner_products: torch.Tensor, gamma: torch.Tensor) -> torch.Tensor:
         # Gamma should be of shape torch.Tensor([gamma])
         assert gamma.shape == (1,)
-        # inner_products has shape number of 4 x v_i_quadruples x v_i_quadruples or 4 x len(X) x len(Y)
-        # h_u_componets should have the same shape
+        # inner_products has shape of 4 x n_samples x n_samples
```
emersodb, line 142: Technically `compute_h_u_from_inner_products` and `compute_h_u_from_inner_products_linear` have the same computations but the underlying values being acted on have different structure. So I broke them apart to allow for distinct comments.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
-        # - u(x_{2i-1}, y_{2i}) - u(x_{2i}, y_{2i-1})
-        h_u = h_u_components[0, :] + h_u_components[1, :] - h_u_components[2, :] - h_u_components[3, :]
-        # this results in a matrix of shape 1 xlen(v_i_quadruples) x len(v_i_quadruples) or 1 x len(X) x len(Y)
+        #       - u(x_{2i-1}, y_{2i}) - u(x_{2i}, y_{2i-1})
+        h_u = h_u_components[:, 0] + h_u_components[:, 1] - h_u_components[:, 2] - h_u_components[:, 3]
+        # this results in a matrix of shape 1 x number of v_i_quadruples
         return h_u.unsqueeze(0)
 
-    def compute_all_h_u_from_inner_products(self, inner_product_quadruples: torch.Tensor) -> torch.Tensor:
+    def compute_all_h_u_from_inner_products(self, inner_product_all_samples: torch.Tensor) -> torch.Tensor:
```
emersodb, line 170: `compute_all_h_u_from_inner_products` and `compute_all_h_u_from_inner_products_linear` actually have the same flow, but they act on different shaped tensors. So I broke them apart to allow for detailed comments.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
+        # We want to compute estimates of the expectation for each RBF kernel WITHOUT using a linear approximation.
+        # So we need to compute ||x - y||^2 for all pairs (x_j, x_k), (x_j, y_k), (x_k, y_j), and (y_j, y_k) for all
+        # j, k in range(n_samples).
+        # NOTE: ||x - y||^2 = <x - y, x - y> = <x, x> + <y, y> - 2<x, y>
         x_x_prime = (
             torch.sum((X**2), dim=1).reshape(-1, 1)
             + torch.sum((X**2), dim=1).reshape(1, -1)
             - 2.0 * torch.mm(X, torch.transpose(X, 0, 1))
         )
-        x_x_prime[x_x_prime < 0] = 0
```
sanaAyrml, line 123: This is something that they added for two sample tests. I am not apposing to delete it though.

fl4health/losses/mkmmd_loss.py
sanaAyrml: Actually if I remember it correctly, later on I change sum to mean as the values were getting so big for QP computation. I am still not sure why it can mess up things as it is just a linear scaling to control effect of high number of dataset features.

fl4health/losses/mkmmd_loss.py
sanaAyrml: That is true, thanks for correcting it. I think otherwise I would have get errors for running with linear approximation.

fl4health/losses/mkmmd_loss.py
emersodb: Ah you might be right. You're correct that it's just a weighted version of the inner product. So I think it's fine. However, since we offer the possibility of normalizing the features, I think we're okay to just leave it as a sum, unless you disagree 🙂

fl4health/losses/mkmmd_loss.py
emersodb: Maybe. I think you were handling most stuff properly actually. I just thought it might be easier to understand with some specific comments given that the structures are a little different.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
-        # and u(x_{2i}, y_{2i-1}), where u is the kernel_index^th kernel
+        # and u(x_{2i}, y_{2i-1}), where u is the kernel_index^th kernel and i indexes over quadruples
         # So we compute:
         #   h_u[x_{2i-1}, x_{2i},y_{2i-1}, y_{2i}] =  u(x_{2i-1}, x_{2i}) + u(y_{2i-1}, y_{2i})
-        # - u(x_{2i-1}, y_{2i}) - u(x_{2i}, y_{2i-1})
-        h_u = h_u_components[0, :] + h_u_components[1, :] - h_u_components[2, :] - h_u_components[3, :]
-        # this results in a matrix of shape 1 xlen(v_i_quadruples) x len(v_i_quadruples) or 1 x len(X) x len(Y)
+        #       - u(x_{2i-1}, y_{2i}) - u(x_{2i}, y_{2i-1})
+        h_u = h_u_components[:, 0] + h_u_components[:, 1] - h_u_components[:, 2] - h_u_components[:, 3]
+        # this results in a matrix of shape 1 x number of v_i_quadruples
```
sanaAyrml, line 167: That is true, thanks for correcting it. I think otherwise I would have get errors for running with linear approximation.

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
+            for j in range(self.kernel_num):
+                for k in range(j + 1):
+                    Q_k_matrix[j][k] += h_u_delta_w_i[j][i] * h_u_delta_w_i[k][i]
+                    if j != k:
+                        Q_k_matrix[k][j] += h_u_delta_w_i[j][i] * h_u_delta_w_i[k][i]
         # Q_k_matrix has shape number of kernels x number of kernels
-        return Q_k_matrix / ((len_w_is) ** 2 - 1)
+        return Q_k_matrix / len_w_is
+
+    def form_kernel_samples_minus_expectation(
```
sanaAyrml, line 261: this is a bit weird name for a function. 😅

emersodb: lol...I'm open to alternative suggestions. I wasn't quite sure how to write
$h_{k_i}(x_s, x_t, y_s, y_t ) - \hat{d}_{k_i}(p, q)$ for all kernels $k_i$, and sample pairs $(x_s, y_s)$, $(x_t, y_t)$

fl4health/losses/mkmmd_loss.py
```diff
@@ -77,156 +80,218 @@ def construct_quadruples(self, X: torch.Tensor, Y: torch.Tensor) -> torch.Tensor
...
-                Q_k_matrix[j][k] += torch.sum(h_u_delta_w_i[j] * h_u_delta_w_i[k])
-                if j != k:
-                    Q_k_matrix[k][j] += torch.sum(h_u_delta_w_i[j] * h_u_delta_w_i[k])
+        for i in range(len_w_is):
+            # For each basis function we're adding in the value of h_{j, \Delta}(w_i)*h_{k, \Delta}(w_i) from the
+            # construction above to the proper entry in Q. Note that Q is symmetric. So we can construct the symmetric
+            # entries at the same time.
+            for j in range(self.kernel_num):
+                for k in range(j + 1):
+                    Q_k_matrix[j][k] += h_u_delta_w_i[j][i] * h_u_delta_w_i[k][i]
```
sanaAyrml, line 255: Why don't we use torch.sum for second dim instead of having the first for loop?

emersodb: I don't know if I can exactly picture the operation you're suggesting, but if its mathematically equivalent, I'm all for not doing three nested loops 😂. Maybe you can write up the operation you're envisioning and I can test out its equivalence?

sanaAyrml: basically this old version I had:
```
     for j in range(self.kernel_num):
            for k in range(j + 1):
                Q_k_matrix[j][k] += torch.sum(h_u_delta_w_i[j] * h_u_delta_w_i[k])
                if j != k:
                    Q_k_matrix[k][j] += torch.sum(h_u_delta_w_i[j] * h_u_delta_w_i[k])
```

emersodb: Ah right. Sorry, I didn't mean to blow that away. I'll mock up a quick test just to make sure they are numerically equivalent and then I'll add it back in.

tests/losses/test_mkmmd_loss.py
```diff
@@ -206,13 +455,17 @@ def test_beta_with_largest_hat_d() -> None:
...
-    beta_target = torch.Tensor([[0, 1, 0]]).to(DEVICE).t()
+def test_optimize_betas_for_X_Y() -> None:
+    # The simple test cases above result in a set of hat_ds that are all negative for the linear estimate. In this
+    # case, we perform the selection of a single kernel as recommended in Gretton
+    degenerate_betas = mkmmd_loss_linear.optimize_betas(X, Y, lambda_m=0.0001)
+    beta_target = torch.Tensor([[1, 0, 0]]).to(DEVICE).t()
     assert torch.all(degenerate_betas.eq(beta_target))
 
+    non_degenerate_beta = mkmmd_loss.optimize_betas(X, Y, lambda_m=0.0001)
+    beta_target = torch.Tensor([[0, 0, 1.0]]).to(DEVICE).t()
```
sanaAyrml, line 466: it is funny that even here with and without linear approximation we have such a different optimized betas.

emersodb: Yeah, I'm thinking that the linear approximation is just bad. The full approximation values for the mean and variance make more sense to me.

fl4health/losses/mkmmd_loss.py
sanaAyrml: let's keep it sum for now to be consistent with non-linear version.

sanaAyrml:
Thanks so much to working through extensive testing of mkmmd.

emersodb:
> Thanks so much to working through extensive testing of mkmmd. Except one possible simplification for "for loop" other things looks good to me.

No problem! I'll refactor the for loop first thing tomorrow morning!

fl4health/losses/mkmmd_loss.py
emersodb: Tested the equivalence and we're good to go. Loops have been updated.

emersodb:
@sanaAyrml: I'll leave this to you for merging when you're ready.

PR #183, opened 2024-06-20
simple scripts for eval and pred with nnunet

scarere:
# Feature

# Short Description

Simple scripts for prediction and evaluation.

- The prediction script is specific to generating predicted segmentation maps for for nnUNet models
- The evaluation script generates the metrics used in the PICAI challenge, but could be applied to any set of predictions and ground truths. These are the metrics we aim to use when evaluating prostate segmentation models.


research/picai/nnunet_scripts/predict.py
```diff
@@ -0,0 +1,58 @@
+import argparse
+
+from nnunetv2.inference.predict_from_raw_data import nnUNetPredictor
+
+
+def main() -> None:
```
jewelltaylor, line 6: Similar to the last PR, can you could add a section to the README with a description of the purpose of the script, what it does and an example invocation

research/picai/nnunet_scripts/eval.py
```diff
@@ -0,0 +1,48 @@
...
+
+    args = parser.parse_args()
+
+    # Compute picai metrics
+    metrics = evaluate_folder(
+        y_det_dir=args.pred_path,
+        y_true_dir=args.gt_path,
+    )
+
+    # Print metrics
```
jewelltaylor, line 38: Maybe worth mentioning in the comments here or in the README what the metrics being computed are. If its a lot of info you can give a short overview and link to the picai documentation or somewhere else that expands on the metrics

research/picai/nnunet_scripts/predict.py
jewelltaylor: for the predict and eval script

jewelltaylor (approved):
Other than the request to add some documentation in the README about the scripts and an example invoking them, I think this looks good! Feel free to add that and then merge this in

PR #184, opened 2024-06-21
CIFAR Experiments PR Suggestions

emersodb:
# PR Type
Other

# Short Description

I figured it was a bit easier for me to suggest some changes via a PR for the sampler tests, since there were a few things I thought might be useful to add. Let me know what you think:

1. Creating a synthetic dataset so we don't have to rely on the presence of or downloading the MNIST dataset.
2. Tried to increase the reuse of the dataset so it didn't need to be individually loaded for each test.
3. Added a few small tests for the fixed hash setting to verify everything was sampled in the same exact manner.

# Tests Added

Modified some tests to reduce reliance on the MNIST dataset and use a faster to load synthetic dataset that we can reuse.

### DIFF
--- a/fl4health/utils/dataset.py
+++ b/fl4health/utils/dataset.py
@@ -89,5 +89,12 @@ class SyntheticDataset(BaseDataset):
         data: torch.Tensor,
         targets: torch.Tensor,
     ):
+        """
+        A dataset for synthetically created data strictly in the form of pytorch tensors.
+        Args:
+            data (torch.Tensor): Data tensor with first dimension corresponding to the number of datapoints
+            targets (torch.Tensor): Target tensor with first dimension corresponding to the number of datapoints
+        """
+        assert data.shape[0] == targets.shape[0]
         self.data = data
         self.targets = targets
--- a/fl4health/utils/dataset.py
+++ b/fl4health/utils/dataset.py
@@ -81,3 +81,13 @@ class DictionaryDataset(Dataset):
     def __len__(self) -> int:
         first_key = list(self.data.keys())[0]
         return len(self.data[first_key])
+
+
+class SyntheticDataset(BaseDataset):
+    def __init__(
+        self,
+        data: torch.Tensor,
+        targets: torch.Tensor,
+    ):
+        self.data = data
+        self.targets = targets
--- a/fl4health/utils/sampler.py
+++ b/fl4health/utils/sampler.py
@@ -1,3 +1,4 @@
+import copy
 import math
 from abc import ABC, abstractmethod
 from logging import INFO
@@ -25,9 +26,10 @@ class LabelBasedSampler(ABC):
 
     def select_by_indices(self, dataset: D, selected_indices: torch.Tensor) -> D:
         if isinstance(dataset, BaseDataset):
-            dataset.targets = dataset.targets[selected_indices]
-            dataset.data = dataset.data[selected_indices]
-            return dataset
+            modified_dataset = copy.deepcopy(dataset)
+            modified_dataset.targets = dataset.targets[selected_indices]
+            modified_dataset.data = dataset.data[selected_indices]
+            return modified_dataset
         elif isinstance(dataset, DictionaryDataset):
             new_targets = dataset.targets[selected_indices]
             new_data: Dict[str, List[torch.Tensor]] = {}
--- a/tests/utils/sampler_test.py
+++ b/tests/utils/sampler_test.py
@@ -46,7 +46,7 @@ def test_dirichlet_sampler_probability_assignment() -> None:
     # Set the random seeds
     set_all_random_seeds(2023)
 
-    # Very mildly heterogeneity with beta 100
+    # Very mild heterogeneity with beta 100
     sampler_1 = DirichletLabelBasedSampler(unique_labels=list(range(10)), sample_percentage=1.0, beta=100)
     prob_ratio_1 = min(sampler_1.probabilities) / max(sampler_1.probabilities)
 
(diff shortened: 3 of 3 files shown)


fl4health/utils/dataset.py
```diff
@@ -81,3 +81,13 @@ def __getitem__(self, index: int) -> Tuple[Dict[str, torch.Tensor], torch.Tensor
     def __len__(self) -> int:
         first_key = list(self.data.keys())[0]
         return len(self.data[first_key])
+
+
```
emersodb, line 85: Adding this in to be used for the tests, but I think it might also be useful more generally. So I put it here in this part of the library.

fl4health/utils/sampler.py
```diff
@@ -25,9 +26,10 @@ def __init__(self, unique_labels: List[Any]) -> None:
 
     def select_by_indices(self, dataset: D, selected_indices: torch.Tensor) -> D:
         if isinstance(dataset, BaseDataset):
-            dataset.targets = dataset.targets[selected_indices]
-            dataset.data = dataset.data[selected_indices]
-            return dataset
+            modified_dataset = copy.deepcopy(dataset)
+            modified_dataset.targets = dataset.targets[selected_indices]
+            modified_dataset.data = dataset.data[selected_indices]
+            return modified_dataset
```
emersodb, line 32: This is fixing a "bug" in the sense that it was returning a dataset but also modifying the dataset passed in-place, which is not what a user might expect. We should either modify the dataset in place and return None or create a new dataset and leave the one that was passed in alone. I went with the former here. It shouldn't have any affect on our examples/tests etc., since in all existing cases we're just sampling once at most. However, the fact that we're not modifying in-place is leveraged for reuse in the tests below.

research/cifar10/fedavg/__init__.py
emersodb, line 1: These init files are required to ensure that we properly define modules here.

tests/utils/sampler_test.py
```diff
@@ -1,36 +1,35 @@
...
+    torch.manual_seed(42)
+    random_inputs = torch.rand((10000, 3, 3))
+    # Note high is exclusive here
+    random_labels = torch.randint(low=0, high=10, size=(1000, 1))
+    # unset seed thereafter
+    torch.seed()
+    return SyntheticDataset(random_inputs, random_labels)
+
+
+SYNTHETIC_DATASET = construct_synthetic_dataset()
```
emersodb, line 22: This dataset will be reused for **all** tests herein

tests/utils/sampler_test.py
```diff
@@ -1,36 +1,35 @@
...
-    new_samples_per_class = [
-        int(num_samples * downsampling_ratio) if i in minority_numbers else num_samples
-        for i, num_samples in enumerate(samples_per_class)
-    ]
-
-    assert len(ds_new) == sum(new_samples_per_class)
-
-    # Testing
-    ds = MnistDataset(data_path=Path("examples/datasets/MNIST"), train=False)
+    ds = SYNTHETIC_DATASET
```
emersodb, line 32: We were essentially running the same test and the two different MNIST splits, which seemed sort of redundant. So I just did the one.

tests/utils/sampler_test.py
```diff
@@ -79,42 +78,41 @@ def test_dirichlet_sampler_with_assigned_probability() -> None:
...
+    new_ds = sampler.subsample(SYNTHETIC_DATASET)
     new_samples_per_class = [new_ds.targets[new_ds.targets == i].size(0) for i in range(10)]
-    new_probs = [x / sum(new_samples_per_class) for x in new_samples_per_class]
+    new_probabilities = [x / sum(new_samples_per_class) for x in new_samples_per_class]
 
-    # Assert that probabilities of sampled indicies are close to assigned probabilities
+    # Assert that probabilities of sampled indices are close to assigned probabilities
     for i in range(10):
-        assert pytest.approx(new_probs[i], abs=0.001) == sampler.probabilities[i]
+        assert pytest.approx(new_probabilities[i], abs=0.01) == sampler.probabilities[i]
```
emersodb, line 104: The synthetic dataset is a about 6x smaller than MNIST, so we saw a slight increase in statistical sampling fluctuation.

tests/utils/sampler_test.py
```diff
@@ -79,42 +78,41 @@ def test_dirichlet_sampler_with_assigned_probability() -> None:
...
+    new_ds = sampler.subsample(SYNTHETIC_DATASET)
     new_samples_per_class = [new_ds.targets[new_ds.targets == i].size(0) for i in range(10)]
-    new_probs = [x / sum(new_samples_per_class) for x in new_samples_per_class]
+    new_probabilities = [x / sum(new_samples_per_class) for x in new_samples_per_class]
 
-    # Assert that probabilities of sampled indicies are close to assigned probabilities
+    # Assert that probabilities of sampled indices are close to assigned probabilities
     for i in range(10):
-        assert pytest.approx(new_probs[i], abs=0.001) == sampler.probabilities[i]
+        assert pytest.approx(new_probabilities[i], abs=0.005) == sampler.probabilities[i]
```
emersodb, line 115: The synthetic dataset is a about 6x smaller than MNIST, so we saw a slight increase in statistical sampling fluctuation.

tests/utils/sampler_test.py
```diff
@@ -128,28 +126,7 @@ def test_dirichlet_sampler_without_hash_key() -> None:
...
-    # Assert that the new distribution with sampler_1 is different from the original distribution
-    assert p_val_1 < 0.01
-    # Assert that the new distribution with sampler_2 is different from the original distribution
-    assert p_val_2 < 0.01
-    # Assert that the new distributions with sampler_1 and sampler_2 are different due to different random seeds
-    assert p_val_3 < 0.01
-
-    # Testing
-    ds = MnistDataset(data_path=Path("examples/datasets/MNIST"), train=False)
+    ds = SYNTHETIC_DATASET
```
emersodb, line 129: Same as above, we were essentially running the same test for the two MNIST splits. So I just removed the redundancy.

fl4health/utils/sampler.py
sanaAyrml: That is a nice catch, though I think my experiments should not be affected by it. Correct me if I'm wrong.

tests/utils/sampler_test.py
sanaAyrml: I am confused why random_inputs and random_labels size are not matching?

fl4health/utils/sampler.py
emersodb: Yeah I don't think it should affect your experiments. It's specifically problematic if you so sampling on the dataset multiple times, which I don't think you're doing in your experiments?

sanaAyrml (commented):
Thanks so much for additions, things look good to me. I just have one comment that I think test and train data is not matching.

fl4health/utils/sampler.py
sanaAyrml: I agree

tests/utils/sampler_test.py
emersodb: lol...that's a good catch! I lost a zero somehow on the labels. I added an assertion in the synthetic dataset that will fail if this happens again. Apparently there is no assertion in the pytorch dataset class that forces this. I'll push the fix in a minute after the tests run.

PR #186, opened 2024-06-27
add setup_client call in get_parameters if not initialized

jewelltaylor:
# PR Type

# Short Description
Link(s) if applicable: [Fix Client Side Init Bug](https://app.clickup.com/t/8688xdtbd)

Add setup_client call in get_parameters if self.initialized is not true for all clients with get_parameters implementations.

# Tests Added

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -139,9 +139,13 @@ class BasicClient(NumPyClient):
                 parameters to be aggregated, but can contain more information.
         """
         if not self.initialized:
-            # If initialized==False, we are doing client side initialization (get_parameters called before fit)
-            # so we must call setup_client first
+            log(INFO, "Setting up client and providing full model parameters to the server for initialization")
+
+            # If initialized==False, the server is requesting model parameters from which to initialize all other
+            # clients. As such get_parameters is being called before fit or evaluate, so we must call
+            # setup_client first.
             self.setup_client(config)
+
             # Need all parameters even if normally exchanging partial
             return FullParameterExchanger().push_parameters(self.model, config=config)
         else:
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -138,9 +138,15 @@ class BasicClient(NumPyClient):
             NDArrays: These are the parameters to be sent to the server. At minimum they represent the relevant model
                 parameters to be aggregated, but can contain more information.
         """
-
-        assert self.model is not None and self.parameter_exchanger is not None
-        return self.parameter_exchanger.push_parameters(self.model, config=config)
+        if not self.initialized:
+            # If initialized==False, we are doing client side initialization (get_parameters called before fit)
+            # so we must call setup_client first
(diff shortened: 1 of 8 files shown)


fl4health/clients/basic_client.py
```diff
@@ -138,9 +138,15 @@ def get_parameters(self, config: Config) -> NDArrays:
             NDArrays: These are the parameters to be sent to the server. At minimum they represent the relevant model
                 parameters to be aggregated, but can contain more information.
         """
-
-        assert self.model is not None and self.parameter_exchanger is not None
-        return self.parameter_exchanger.push_parameters(self.model, config=config)
+        if not self.initialized:
+            # If initialized==False, we are doing client side initialization (get_parameters called before fit)
```
emersodb, line 142: I'd suggest fleshing this out slightly more to something like
```
If initialized==False, the server is requesting model parameters from which to initialize all other clients. As such get_parameters is being called before fit or evaluate, so we must call setup_client first.
```

fl4health/clients/basic_client.py
```diff
@@ -138,9 +138,15 @@ def get_parameters(self, config: Config) -> NDArrays:
...
+            # so we must call setup_client first
```
emersodb, line 143: Perhaps we can also add a log to make it visible that the client was asked to provide parameters for initialization? Maybe
```
log(INFO, "Setting up client and providing full model parameters to the server for initialization"
```

fl4health/clients/mr_mtl_client.py
```diff
@@ -83,8 +83,15 @@ def get_parameters(self, config: Config) -> NDArrays:
...
-        return self.parameter_exchanger.push_parameters(self.model, config=config)
+        if not self.initialized:
+            # If initialized==False, we are doing client side initialization (get_parameters called before fit)
+            # so we must call setup_client first
+            self.setup_client(config)
+            # Need all parameters even if normally exchanging partial
+            return FullParameterExchanger().push_parameters(self.model, config=config)
+        else:
+            assert self.model is not None and self.parameter_exchanger is not None
+            return self.parameter_exchanger.push_parameters(self.model, config=config)
```
emersodb, line 94: Maybe this is a naive question, but do we actually need to override the get_parameters function for MR-MTL clients? It looks identical to that of BasicClient.

jewelltaylor: I thought the same thing but I figured there was a reason haha, I will just call basic one

emersodb: I think you can just drop this override 🙂 and it will be called.

emersodb (approved):
A couple of minor comments. Generally looks good to go

fl4health/clients/basic_client.py
jewelltaylor: Thanks, this is much better :)

fl4health/clients/basic_client.py
jewelltaylor: Good idea!

PR #191, opened 2024-07-11
Model Merge Functionality

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Add Model Merge Server, Strategy and Client](https://app.clickup.com/t/86890un8j)

Adding `ModelMergeServer` `ModelMergeStrategy` and `ModelMergeClient` to support the setup in which clients have trained the same architecture on their local dataset and wish to perform model merging. Model merging simply involves averaging the weights of client models and evaluating the merged model on the client datasets and optionally a server side dataset. This PR also includes an example of model merging and a prerequisite `train_silo.py` script which trains client models initially. 

# Tests Added
Added tests for ModelMergeStrategy


fl4health/clients/model_merge_client.py
```diff
@@ -0,0 +1,292 @@
...
+    @abstractmethod
+    def get_test_data_loader(self, config: Config) -> DataLoader:
+        """
+        User defined method that returns a PyTorch Test DataLoader.
+
+        Args:
+            config (Config): The config from the server.
+
+        Returns:
+            DataLoader. Client test data loader.
```
emersodb, line 290: I think rather than a period we want a colon after Dataloader?

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+from flwr.server.strategy import Strategy
+
+from fl4health.reporting.metrics import MetricsReporter
+from fl4health.strategies.model_merge_strategy import ModelMergeStrategy
+
+
+class ModelMergeServer(Server):
+    def __init__(
+        self,
+        *,
```
emersodb, line 19: Do you mind reminding me what this * notation does? I think we can limit to just the inputs you have other than the star, but perhaps I'm forgetting something.

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+        metrics_reporter: Optional[MetricsReporter] = None,
+    ) -> None:
+        """
+        ModelMergeServer provides functionality to fetch client weights, perform a simple average,
+            redistirbute to clients for evaluation. Optionally can perform server side evaluation as well.
+        Args:
+            client_manager (ClientManager): Determines the mechanism by which clients are sampled by the server, if
+                they are to be sampled at all.
+            strategy (Optional[Strategy], optional): The aggregation strategy to be used by the server to handle.
+                client updates and other information potentially sent by the participating clients. If None the
```
emersodb, line 31: In this case, I don't think we'll be receiving other information? Maybe I'm wrong though.

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def fit(self, num_rounds: int, timeout: Optional[float]) -> Tuple[History, float]:
+        """
+        Performs a fit round in which the local client weights are evaluated on their test set,
+            uploaded to the server and averaged, then redistributed to clients for evaluation.
+            Optionally, can perform evaluation of the merged model on the server side as well.
+
+        Args:
+            num_rounds (int): Not used.
+            timeout (Optional[float]): Timeout in seconds that the server should wait for the clients to response.
```
emersodb, line 52: Maybe this was a typo that you copied from somewhere else but I think we want "respond" here 🙂

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+            timeout=timeout,
+        )
+
+        if res_fit is not None:
+            parameters_prime, fit_metrics, _ = res_fit  # fit_metrics_aggregated
+            if parameters_prime:
+                self.parameters = parameters_prime
+            history.add_metrics_distributed_fit(server_round=1, metrics=fit_metrics)
+        else:
+            log(DEBUG, "Federated Model Merging Failed")
```
emersodb, line 78: I think perhaps we log this at the ERROR level? Seems more serious than debug 😂

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+
+        res_fed = self.evaluate_round(server_round=1, timeout=timeout)
+        if res_fed is not None:
+            _, evaluate_metrics_fed, _ = res_fed
```
emersodb, line 82: Maybe just put a comment as to what we're discarding here?

fl4health/server/model_merge_server.py
```diff
@@ -0,0 +1,104 @@
...
+        Args:
+            client_manager (ClientManager): Determines the mechanism by which clients are sampled by the server, if
+                they are to be sampled at all.
+            strategy (Optional[Strategy], optional): The aggregation strategy to be used by the server to handle.
+                client updates and other information potentially sent by the participating clients. If None the
+                strategy is FedAvg as set by the flwr Server.
+            metrics_reporter (Optional[MetricsReporter], optional): A metrics reporter instance to record the metrics
+                during the execution. Defaults to an instance of MetricsReporter with default init parameters.
+        """
+        assert isinstance(strategy, ModelMergeStrategy)
```
emersodb, line 36: If we assert the strategy type here, maybe we update the comment for strategy, since we state that it can be none there to get default fedavg, but it actually can't due to this assert.

fl4health/strategies/model_merge_strategy.py
```diff
@@ -0,0 +1,271 @@
+from logging import WARNING
```
emersodb, line 1: I'm probably missing something here, so please correct me if I'm wrong. Is there any reason this strategy can't inherit from `BasicFedAvg` and we just override methods like `initialize_parameters` and `configure_fit` to do what is needed for this strategy?

emersodb: Maybe the above is more trouble than its worth, but I think it could help eliminate some code duplication and also give the user flexibility to model merge with uniform or weighted FedAvg as well.

fl4health/utils/load_data.py
```diff
@@ -94,6 +94,35 @@ def load_mnist_data(
     return train_loader, validation_loader, num_examples
 
 
+def load_mnist_test_data(
+    data_dir: Path,
+    batch_size: int,
+    sampler: Optional[LabelBasedSampler] = None,
+    transform: Optional[Callable] = None,
+) -> Tuple[DataLoader, Dict[str, int]]:
+    """Load MNIST test set only."""
```
emersodb, line 103: Maybe we can add true doc strings for this guy since it's net-new? The documentation is also likely easily transferable to the regular `load_mnist_data` function as well if you don't mind?

emersodb (commented):
Overall I think this is a good start. I think there are a couple of small places where a refactor would reduce a little code duplication. If we can inherit from our FedAvg strategy, that would also make your life easier because that strategy already has some nice tests.

examples/model_merge_example/README.md
```diff
@@ -0,0 +1,36 @@
+# Model Merge Example
+This example provides an example of a model merging setup on the MNIST dataset in which clients each have
```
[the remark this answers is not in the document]
jewelltaylor: word salad 🥗

fl4health/server/model_merge_server.py
jewelltaylor: Placing * in the parameter list makes all subsequent parameters keyword-only. This is done in all of the base classes in flwr (ie Server, Strategy, and Client). I think probably since they have large amounts of input arguments relying on the position is somewhat error prone. Thinking we should probably stick with the flwr convention even though we could technically remove it. Happy to discuss further if you disagree though :)

examples/model_merge_example/README.md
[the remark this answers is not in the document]
emersodb: lol...we could also go with "This example provides an example of an exemplary model merging setup with an MNIST dataset as an example."

fl4health/server/model_merge_server.py
emersodb: Ah right. Yes, I'm definitely good with that. Thanks for the reminder.

jewelltaylor:
I meant mine was word salad not your suggestion 😂😂 will definitely update to your suggestion sorry I wasn’t clear

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

examples/model_merge_example/README.md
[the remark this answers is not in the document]
jewelltaylor: Sorry I meant mine was word salad LOL! I'll update to yours it's much better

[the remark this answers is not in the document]
emersodb: No I know...I was just making a joke 😂

examples/model_merge_example/train_silo.py
```diff
@@ -0,0 +1,104 @@
+import argparse
```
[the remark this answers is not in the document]
jewelltaylor: Yeah I agree! I don't think the details here are helpful really, it just served to allow someone to run the example (almost) out of the box. Will update!

examples/model_merge_example/config.yaml
```diff
@@ -0,0 +1,10 @@
+# Parameters that describe server
+n_server_rounds: 2 # The number of rounds to run FL
```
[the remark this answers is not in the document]
jewelltaylor: Good catch! I enforce only a single round (merge + evaluation) occurs in the strategy but i should remove this as its confusing

examples/model_merge_example/server.py
```diff
@@ -0,0 +1,98 @@
...
+        for input, target in loader:
+            input, target = input.to(device), target.to(device)
+            preds = {"predictions": model(input)}
+            evaluate_metric_manager.update(preds, target)
+
+    return 0.0, evaluate_metric_manager.compute()
+
+
+def main(config: Dict[str, Any], data_path: Path, batch_size: int) -> None:
+    _, val_loader, _ = load_mnist_data(data_path, batch_size)
```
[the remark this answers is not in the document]
jewelltaylor: The reason I did this is because I wanted the client side evaluation dataset to be different than the server side evaluation set. Do you think its worth it or not? I can change it to be the same if you think thats better

examples/model_merge_example/server.py
```diff
@@ -0,0 +1,98 @@
...
+
+    parser.add_argument(
+        "--data_path",
+        action="store",
+        type=str,
+        help="Path to server side evaluation dataset.",
+        default="examples/datasets/MNIST",
+    )
+
+    parser.add_argument(
```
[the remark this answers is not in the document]
jewelltaylor: Yeah looking back at this, I think I wanted to loosen the uniform batch size requirement across clients in cases where VRAM may be different or something and we want to have different batch sizes for different clients. But I think that requires some more thought and is sufficiently different from the changes I am making here that it warrants a separate PR. I will revert this to just use the config and roll back the changes I made to config

fl4health/clients/model_merge_client.py
```diff
@@ -0,0 +1,292 @@
...
+                first fitting round.
+        """
+        assert self.initialized
+        self.parameter_exchanger.pull_parameters(parameters, self.model)
+
+    def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict[str, Scalar]]:
+        """
+        Initializes client, validates local client model on local test data and returns parameters,
+            test dataset length and test metrics. Importantly, parameters from Server, which is empty,
+            is not used to initialized the client model.
```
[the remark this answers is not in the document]
jewelltaylor: Validate is called in other Client fit function, the only difference here is that we are calling on the test set instead of the validation. The rationale of doing the client side evaluation is to see how well the model performs on the local test set prior to model merging. We then distribute the merged model to the clients who again evaluate on the test set. Then we can see the difference in performance. I could make this optional, but then if no server_side_evaluation_fn is supplied (optional), there will be no metrics. Let me know if this makes sense to do and if I am missing something. I can definitely add some documentation that lays that out if you think its the right design decision

fl4health/clients/model_merge_client.py
```diff
@@ -0,0 +1,292 @@
...
+        val_metrics = self.validate()
+
+        self.metrics_reporter.add_to_metrics_at_round(
+            1,
+            data={
+                "fit_metrics": val_metrics,
+            },
+        )
+
+        return self.get_parameters(config), len(self.test_loader.dataset), val_metrics  # type: ignore
```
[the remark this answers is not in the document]
jewelltaylor: We have to do this for the other clients as well, its just happening in the setup_client where we extract the sample counts from the datasets. Since technically not all pytorch dataset objects have a __len__ method but in practice they almost always do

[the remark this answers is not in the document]
jewelltaylor: I will move the definition into the setup_client, because ignoring there makes it ambiguous what the issue is. Thanks for flagging

examples/model_merge_example/server.py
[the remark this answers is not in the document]
emersodb: Nope. That's good with me! Maybe just put a comment there to short circuit such a question?

examples/model_merge_example/server.py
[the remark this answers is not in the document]
emersodb: Sure...maybe just log it in the backlog for us to tackle next at some point!

examples/model_merge_example/train_silo.py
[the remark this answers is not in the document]
emersodb: I'm good with either. It's a potentially useful piece of code. So if you want to keep it around please do.

fl4health/clients/model_merge_client.py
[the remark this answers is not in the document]
emersodb: Nope. That makes sense. Let's keep it like this and just add some discussion in the docstrings.

fl4health/clients/model_merge_client.py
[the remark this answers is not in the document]
emersodb: Ah right...I remember. Thanks for the explanation.

fl4health/server/model_merge_server.py
jewelltaylor: Yup, will fix this in the other place too!

fl4health/strategies/model_merge_strategy.py
jewelltaylor: Perhaps we can have a quick call tomorrow about this to hash it out but my thinking was we will also have to override aggregate_fit and aggregate_evaluate in addition to configure_fit and initialize_parameters because in this setting it seems to only make sense to do a uniform average. It feels a bit weird to average the model weights proportional to the amount of samples in the client test set. Perhaps it makes sense for the evaluation though. However, I also have to modify aggregate_evaluate to not compute a loss because we are not defining one. So at that point I have to override some of the core methods. And then on top of that, there is a few arguments that don't make sense for ModelMergeClient like initial_parameters and potentially weighted aggregation and weight eval losses depending on what we land on for supporting weighting for parameters and losses for model merging.

jewelltaylor: I am open to extending the BasicFedAvg just wanted to add where I was coming from and see what you think! Lets discuss tomorrow :)

examples/model_merge_example/train_silo.py
```diff
@@ -0,0 +1,104 @@
...
+    for _ in range(num_epochs):
+        for input, target in train_loader:
+            input, target = input.to(device), target.to(device)
+            optimizer.zero_grad()
+            preds = model(input)
+            loss = criterion(preds, target)
+            loss.backward()
+            optimizer.step()
+
+    model.cpu()
```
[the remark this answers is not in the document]
jewelltaylor: You are right for what its worth! Deleted the file so its not relevant anymore though :)

.gitignore
```diff
@@ -166,4 +166,5 @@ settings.json
 **/*.pkl
 **/*.png
 **/*.pt
+!assets/checkpoints_for_examples/model_merge_example/*.pt
```
emersodb, line 169: I think this is okay to do the pattern ignore here, since the pt files won't change too much. Just wanted to note (in case you didn't know) that you can override ignores for files by doing a `git commit -f` with a path to the specific file. Those will then be pushed to the remote, but any changes to them will remain untracked in the future.

fl4health/clients/model_merge_client.py
```diff
@@ -0,0 +1,292 @@
...
+            TypeError: raised if data is not of type torch.Tensor or Dict[str, torch.Tensor]
+        """
+        if isinstance(data, torch.Tensor):
+            return data.to(self.device)
+        elif isinstance(data, dict):
+            return {key: value.to(self.device) for key, value in data.items()}
+        else:
+            raise TypeError("data must be of type torch.Tensor or Dict[str, torch.Tensor].")
+
+    def narrow_config_type(self, config: Config, config_key: str, narrow_type_to: Type[T]) -> T:
```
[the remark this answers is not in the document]
emersodb: Just want to check on this. I see that you pulled `generate_hash` into utils which is great. Did you rule out moving this narrowing functionality? If it's more trouble than its worth that's fine. I'll leave it to you to decide 🙂

fl4health/utils/load_data.py
```diff
@@ -124,9 +155,9 @@ def load_mnist_test_data(
 
 
 def get_cifar10_data_and_target_tensors(data_dir: Path, train: bool) -> Tuple[torch.Tensor, torch.Tensor]:
-    mnist_dataset = CIFAR10(data_dir, train=train, download=True)
```
emersodb, line 98: Nice catch!

fl4health/strategies/model_merge_strategy.py
emersodb: Yeah I think that makes sense. I hadn't thought deeply about it. Was just thinking we might be able to reuse some code. Let's have a brief discussion today. However, your reasoning for not doing it above seems good to me.

.gitignore
jewelltaylor: Good to know! In that case, I will just remove it :)

emersodb (approved):
Good to go from my perspective.

PR #195, opened 2024-07-22
Added epoch hook for lr scheduling and made client logs look better

scarere:
# PR Type

# Short Description

- Added an update_before_epoch hook method to BasicClient
- Used update_before_epoch to enable lr scheduling in nnUNetClient
- Added the add_to_initial_log_str attribute to Basic Client to allow users to add information to the initial log string (such as current LR)
- Reformatted logging in _handle_logging method of BasicClient to make things more readable

Note: The add_to_initial_log_str seemed like the cleanest solution to me to allow users to add info to the inital log string. I made sure to add a comment in the BasicClient init where the attribute is defined to explain how it works. Although it is a functionality that could be easily overlooked if users don't look through the BasicClient code. 

Additionally, one thing i did notice is that by introducing this attribute to basic client, you can actually remove the current_epoch argument of _handle_logging and instead just do ```self.add_to_initial_log_str += f' Current Epoch: {local_epoch}'``` in train_by_epochs. Curious to get your opinion on whether or not I should make this change, however i didn't for now since it might affect other clients

Here is a screenshot of how the logs look now:

Screenshot from 2024-07-22 12-46-55

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -77,6 +77,10 @@ class BasicClient(NumPyClient):
 
         self.initialized = False  # Whether or not the client has been setup
 
+        # Variable to store additional information to be added to initial log
+        # string. This variable is reset every time _handle_logging is called
+        self.add_to_initial_log_str = ""
+
         # Loss and Metric management
         self.train_loss_meter = LossMeter[TrainingLosses](loss_meter_type, TrainingLosses)
         self.val_loss_meter = LossMeter[EvaluationLosses](loss_meter_type, EvaluationLosses)
@@ -420,19 +424,24 @@ class BasicClient(NumPyClient):
             current_epoch (Optional[int]): The current epoch of local training.
             logging_mode (LoggingMode): The logging mode (Training, Validation, or Testing).
         """
+        log(INFO, "")  # An empty log line for aesthetics
+
         initial_log_str = f"Current FL Round: {str(current_round)}\t" if current_round is not None else ""
         initial_log_str += f"Current Epoch: {str(current_epoch)}" if current_epoch is not None else ""
+        initial_log_str += self.add_to_initial_log_str  # Add user defined info
         if initial_log_str != "":
             log(INFO, initial_log_str)
+            self.add_to_initial_log_str = ""  # Reset variable
 
-        metric_string = "\t".join([f"{key}: {str(val)}" for key, val in metrics_dict.items()])
-        loss_string = "\t".join([f"{key}: {str(val)}" for key, val in loss_dict.items()])
-
+        # Log loss/losses
         metric_prefix = logging_mode.value
-        log(
-            INFO,
-            f"Client {metric_prefix} Losses: {loss_string} \n" f"Client {metric_prefix} Metrics: {metric_string}",
-        )
+        log(INFO, f"Client {metric_prefix} Losses:")
+        [log(INFO, f"\t {key}: {str(val)}") for key, val in loss_dict.items()]
+
+        # Log metrics if any
+        if len(metrics_dict.keys()) > 0:
+            log(INFO, f"Client {metric_prefix} Metrics:")
+            [log(INFO, f"\t {key}: {str(val)}") for key, val in metrics_dict.items()]
 
     def _handle_reporting(
         self,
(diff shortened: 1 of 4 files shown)


fl4health/clients/basic_client.py
```diff
@@ -77,6 +77,10 @@ def __init__(
 
         self.initialized = False  # Whether or not the client has been setup
 
+        # Variable to store additional information to be added to initial log
+        # string. This variable is reset every time _handle_logging is called
+        self.add_to_initial_log_str = ""
```
emersodb, line 82: I like the idea of providing a way for child clients to inject logging of specific information a lot. I'm not opposed to the route you have taken here, but sometime managing, appending to, and clearing "state" like this string can get a little convoluted. I'll suggest another potential way to do this below and you can see if you like it
```
def get_client_specific_logs(self) -> str:
    """
    This function can be used to provide any client specific or state 
    For example, perhaps a client uses a special LR scheduler and wants to 
    each logging round.
    NOTE: Needs to be overridden by an inheriting client to produce a 
    Returns:
        str: The logging value to be added to the logging string
    """
    return ""
```
Then you would simply do
```
initial_log_str += self.get_client_specific_logs()  # Add user defined info
```
in `_handle_logging` and then you'd just have 
```
def update_before_epoch(self, epoch: int) -> None:
        self.nnunet_trainer.lr_scheduler.step(epoch)

def get_client_specific_logs(self) -> str:
    lr = self.optimizers["global"].param_groups[0]["lr"]
    return f" Current LR: {lr}"
```
in your nnUnetClient code. This feels slightly cleaner to me, but would like to hear your thoughts.

fl4health/clients/basic_client.py
```diff
@@ -420,19 +424,24 @@ def _handle_logging(
...
-        loss_string = "\t".join([f"{key}: {str(val)}" for key, val in loss_dict.items()])
-
+        # Log loss/losses
         metric_prefix = logging_mode.value
-        log(
-            INFO,
-            f"Client {metric_prefix} Losses: {loss_string} \n" f"Client {metric_prefix} Metrics: {metric_string}",
-        )
+        log(INFO, f"Client {metric_prefix} Losses:")
+        [log(INFO, f"\t {key}: {str(val)}") for key, val in loss_dict.items()]
```
emersodb, line 439: Interesting. I didn't know you could actually wrap a log this way.

emersodb (commented):
Overall, I think the PR looks good. I made a suggestion on how we might handle the logging in an alternative fashion, rather than having a kind of "buffer" that we add to and clear. 

For the thought on removing the argument for `_handle_logging`, I'd say even if we go your route, we can leave that argument, as I think it might still be useful at some point downstream, even. if there is an alternative way of putting the information into the logs.

fl4health/clients/basic_client.py
scarere: I agree that this is cleaner, and then makes it more obvious to the user that this is an available functionality. I do have some thoughts though that I'd like to know your opinion on. 

First is that it might be unclear exactly when handle_logging is called and by extension when get_client_specific_logs is called. For example the user might not be clear on whether LR is the current lr that is about to be used, or the most recent LR that was used. An easy fix would be to commit to always calling handle logs as the last step. The one exception might be the update_after_epoch hook, but I've actually removed this and replaced it with update before epoch. Either way if we can commit to when _handle_logging will be invoked, then we can communicate that to the user and i think that addresses that issue.

My second thought is that if we are going to add a ```get_client_specific_logs method```, then we should also add a ```get_client_specific_reports``` method to stay consistent.

My third thought is that ```get_client_specific_logs``` doesn't really communicate to the user that what is return is just appended to the initial log string. Again we could just communicate this in the docstring, or perhaps change the method name, although it does seem a bit weird to allow a user to add to the initial log string, but not to the rest of the logs itself. One idea that comes to mind is to have ```get_client_specific_logs``` return two values, a string and maybe an optional list of logs to add after logging the losses and metrics.

```python
class LogLevel(Enum):
    ERROR = logging.ERROR
    INFO = logging.INFO
    ...

def get_client_specific_logs(self) -> Tuple[Optional[str], Optional[List[Tuple[LogLevel, str]]]]:
    logs = []
    lr = self.optimizers["global"].param_groups[0]["lr"]
    logs.append((INFO, 'some more info that i want on a seperate line)
    return f" Current LR: {lr}", logs
```

Overall i think there are more upsides and less downsides with your approach. Curious to know what you think about some of these ideas

fl4health/clients/basic_client.py
scarere: neither did I, just seemed like something that should work so I tried it and it did. My editor wraps at 80 characters so I like to minimize lines as much as possible if they'll fit since my files are already longer bc of the word wrapping

emersodb (approved):
Looks good to go to me.

PR #198, opened 2024-07-25
Nnunet ensemble

scarere:
# PR Type

# Short Description

This PR introduces major updates to the prediction and evaluation pipelines for nnunet models. Namely a script that can be called to do all of inference, lesion extraction and picai_eval evaluation. 

- Updates to logging to make everything look nice
- Included information on how long each step takes to run
  - This will be useful when inference datasets get large to know what bottleneck is
- Update predict script to return probabilities, annotations and case identifiers. 
- Updated eval script to predict from probabilities and extract detection maps manually
- predict_and_eval script that combines everything
  - performs inference for arbitrary number of nnunet models
  - Ensembles the predicted probabilities for different checkpoints and nnunet configs
  - Extracts lesion detection maps and predicted annotations/segmentations
  - Computes picai eval metrics from detection maps
  - Allows optional saving of probabilities, detection maps, annotations/segmentations and evaluation result metrics

### Notes
I would like to mention that in the second half of working on this, the code had to become a lot more involved than I would have liked. I think this is a good solution for now as it is robust, generalizable to multiple use cases (ie. different amounts of models, multiclass segmenetation etc.) and just works. However in the future I would love to refactor this somehow, make it shorter and cleaner. I think there will be an opportunity across the entire picai research folder to refactor at some point and delete code. However I didn't want to expand the scope of this PR as refactoring would potentially be much more involved under the hood of the various API's being used and take a lot of time. Additionally, waiting on some of these packages to fix issues or make improvements to their own API's (I create an issue and they normally just ask me to do the PR for them lol)

That being said, if you see some easy simplifications I'd love to know. I'm really in deep on this one, so my head hurts when trying to reimagine it. This was a frustrating PR because I don't love the comprimises I had to make.

- Currently everything needs to be loaded into memory simultaneously.
  - nnunet uses multiprocessing exactly for this reason to avoid having to have the inference dataset in memory
  - However we need to have it in memory currently in order to do the other steps without having to save and load files intermediately
  - Potential future improvement that would be great would be to have the whole pipeline done one sample at a time with multiprocessing, and then just add our own arguments/flags to save each individual sample. However this would involve refactoring everything and redoing nnunet's inference code in part
    - However I'm already kind of overriding nnunet's main inference method to get it to return more information 
- I'd love to add some convenient file loading functionality in the future
  - picai eval has some cool stuff in their fileloading that allows filenames to have 'postfixes'. Basically it learns to ignore a string such as '_probs' at the end of the filename when looking for case identifiers 
  - I'd love for there to be an option to specify a nnunet dataset id or the current method of direct folder paths
  - Right now we need a dataset json as an input, might be worth removing this dependency as the only info we need from the dataset json are the file endings and the labels for the different classes.


scarere:
Static code check failure was due to a new vulnerability found in pytorch versions under 2.2.0. I tried updating the dependencies but poetry still resolves torch version 2.1.2 despite the pyproject.toml allowing any version above 2.1.1. I'm assuming some other dependency is preventing us from upgrading to a newer torch version so I just ignored the vulnerability for now

scarere:
As for the smoke tests, it has something to do with the metrics, but I didn't change anything in fl4health, all the changes were restricted to the research folder, so i don't know why the smoke tests are failing. Also I'm not sure if there's a way to see which smoke tests are failing but here is the error I get when I run them myself.

Screenshot from 2024-07-25 11-09-44

scarere:
Ok not sure whats going on because the smoketests seem to be running now, i tried running it locally again and got a different error. I've updated my dependencies, did a git pull, so everything should be the same, but when I run python -m tests.smoke_tests.run_smoke_tests i get an error

Screenshot from 2024-07-25 11-18-54

scarere:
Ok strange that they all passed now but I'll take it

scarere:
Also I was able to remove ignoring the mlflow vulnerabilities so I guess that issue was solved

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+import warnings
+from collections import defaultdict
+from os.path import join
+from typing import Any, Dict, Iterable, List, Optional
 
-from picai_eval import evaluate_folder
+import numpy as np
+from numpy.typing import NDArray
+
+with contextlib.redirect_stdout(open(os.devnull, "w")):
```
jewelltaylor, line 12: If we don't log this to the console, should we maybe redirect to a file or something? Suppressing it entirely may make it difficult to catch things if they pop up in picai_eval

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+            file. Typically just the filenames without the extension
+        postfixes (Optional[List[str]], optional): A list of strings to append
+            to the case identifiers when looking for files. For example
+            '_labels'. Defaults to None.
+        extensions (List[str], optional): A list of possible image extensions.
+            Defaults to [".nii.gz", ".nii", ".mha", ".mhd", ".npz", ".npy"].
+
+    Returns:
+        NDArray: A numpy array containing the images for each case identifier.
+            The first dimension will be the number of images and thos images
```
jewelltaylor, line 42: nit: thos -> those

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+            assumed to be at index 0. Therefore the output shape is
+            (num_samples, num_classes-1)
+    """
+    num_classes = probability_maps.shape[1]
+
+    detection_maps = []
+    for prob_map in probability_maps:
+        cls_probs = []
+        for cls in range(1, num_classes):  # Ignore background class at idx 0
+            cls_probs.append(extract_lesion_candidates(prob_map[cls], num_lesions_to_extract=5)[0])
```
jewelltaylor, line 80: Maybe worth elaborating why we are setting num_lesions_to_extract to 5. If it is an arbitrary parameter that we may want to change, perhaps we include as an argument to this function and default to 5.

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+    **kwargs: Any
+) -> PicaiEvalMetrics:
+    """
+    Computes the picai evaluation metrics provided the predicted lesion
+    detection maps and the ground truth annotations. Extends picai_eval to
+    allow multiclass evaluation
+
+    Args:
+        detection_maps (NDArray): The predicted lesion detection maps. Must
+            have shape (num_samples, num_lesion_classes, ...). The background
```
jewelltaylor, line 99: Assuming these are the same detection maps returned by get_detection_maps, maybe we unify the term used to describe the number of classes (ie num_classes used in get_detection_maps vs num_lesion_classes in get_picai_metrics) just so people don't get confused

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+            arguments
+
+    Returns:
+        picai_eval.metrics.Metrics: A picai eval metrics object that has
+            combined the results from all classes into a single object
+    """
+    if "y_det" in kwargs or "y_true" in kwargs or "subject_list" in kwargs:
+        raise KeyError(
+            """Got one of 'y_det', 'y_true' or 'subject_list' in keyword
+            arguments. The arguments to get_picai_metrics are being passed to
```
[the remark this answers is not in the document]
jewelltaylor: also small nit: We could probably list the keys that actually violate this condition as opposed to saying Got one of  y_true, y_det and subject_list. And I think technically we can get one *or more keys that violate this condition

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+            these keys"""
+        )
+
+    if detection_maps.shape[1] != ground_truth_annotations.shape[1]:
```
jewelltaylor, line 125: Since we are augmenting the ground_truth_annotations shape to remove the background class, maybe its worth logging this

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+        # Assume background class has not been removed
+        ground_truth_annotations = ground_truth_annotations[:, 1:]
+        assert (
+            detection_maps.shape == ground_truth_annotations.shape
+        ), "Got unexpected shapes for detection maps and ground truth annotations"
```
jewelltaylor, line 130: nit: We can probably add in the shapes of detections_maps and ground_truth_annotations into this

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+            evaluate(
+                y_det=detection_maps[:, cls],
+                y_true=ground_truth_annotations[:, cls],
+                subject_list=case_identifiers,
+                **kwargs
+            )
+        )
+
+    subject_list = metrics[0].subject_list
+    assert isinstance(subject_list, list), "Got unexpected subject list from picai eval metrics object"
```
jewelltaylor, line 146: Just curious for my understanding, in what cases would this subject_list not be a list?

research/picai/nnunet_scripts/eval.py
```diff
@@ -30,18 +192,20 @@ def main() -> None:
     args = parser.parse_args()
 
     # Compute picai metrics
-    metrics = evaluate_folder(
+    # Warning, this method assumes binary segmentation (ie. only two classes)
```
jewelltaylor, line 195: Curious why we aren't leveraging the get_picai_metrics here but we are in predict_and_eval.py

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+    output_folder: Optional[str] = None,
+    save_probability_maps: bool = False,
+    save_detection_maps: bool = False,
+    save_annotations: bool = False,
+    verbose: bool = True,
+) -> None:
+    """
+    Runs inference on raw input data given a number of compatible nnunet
+    models. Then extracts detection maps from those predictions and evaluates
+    the model using the standardPICAI evaluation metrics.
```
jewelltaylor, line 29: nit: standardPICAI to standard PICAI

jewelltaylor: If not done elsewhere, maybe worth listing the standard computed metrics in brackets

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Args:
+        config_path (str): Path to a yaml config file. The three required keys
+            are plans, dataset_json and one or more nnunet_configs (eg. 2d,
+            3d_fullres etc.). The nnunet config keys should contain a list of
+            paths. If the path points to a file it should be a model
+            checkpoint. The model checkpoints can be dicts with thE
```
jewelltaylor, line 36: nit: thE -> the

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+        input_folder=inputs_folder,
+        probs_folder=prob_out_folder,
+        annotations_folder=annotations_folder,
+        verbose=verbose,
+    )
+    del annotation_preds
+
+    # Extract lesion detection maps
+    if verbose:
+        log(INFO, "")
```
jewelltaylor, line 95: Is this just to create seperation or something?

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+        det_output_folder = join(output_folder, "detection_maps")
+        if not exists(det_output_folder):
+            os.makedirs(det_output_folder)
+        t = time.time()
+        for det, case in zip(detection_maps, case_identifiers):
+            ofile = join(det_output_folder, case + ".npz")
+            np.savez_compressed(ofile, detection_map=det)
+        secs = time.time() - t
+        if verbose:
+            log(INFO, f"Saved detection maps to disk: {secs:.2f}s total, {secs/len(detection_maps):.1f}s/case")
```
jewelltaylor, line 116: Nice additions with logging the run times for different components!

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+            log(INFO, f"\tPicai Score: {metrics.score}")
+
+    # Save metrics
+    if output_folder is not None:
+        metrics.save_full(join(output_folder, "picai_eval_metrics.json"))
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        prog="Predict and Evaluate nnUNet models or nnUNet Ensembles",
```
jewelltaylor, line 149: Amazing documentation! Thank you

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+        )
+
+    subject_list = metrics[0].subject_list
+    assert isinstance(subject_list, list), "Got unexpected subject list from picai eval metrics object"
+    lesion_results: Dict[Any, list] = defaultdict(list)
+    lesion_weights: Dict[Any, list] = defaultdict(list)
+    case_targets: Dict[Any, int] = defaultdict(int)
+    case_preds: Dict[Any, float] = defaultdict(float)
+    for s in subject_list:
+        # Ignoring mypy errors here for now because i don't know how to get around them
```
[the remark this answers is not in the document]
jewelltaylor:
```python
for s in subject_list:
  for m in metrics:
        assert (
            m.lesion_results is not None
            and not isinstance(m.lesion_results, Path)
            and not isinstance(m.lesion_results, os.PathLike)
            and not isinstance(m.lesion_results, str)
        )
        lesion_results[s].extend(m.lesion_results[s])
        assert m.lesion_weight is not None
        lesion_weights[s].extend(m.lesion_weight[s])

    # Should probably do an assertion that none of the m.case_target/m.case_pred are None instead of filtering 
    case_targets[s] = int(any(m.case_target[s] for m in metrics if m.case_target is not None))
    case_preds[s] = max([m.case_pred[s] for m in metrics if m.case_pred is not None])
```

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+        [lesion_results[s].extend(m.lesion_results[s]) for m in metrics]  # type: ignore
```
jewelltaylor, line 153: Even though its a little more verbose, I think having a for loop here is better practice than modifying a list inside of a list comprehension

jewelltaylor (commented):
All in all, some great work! I have comments mosts around small details, clarifications and documentation requests. 

I also echo Masoom's point about supporting evaluation without requiring the entire test set in RAM. As you mention, it seems like a lot of this would be transferrable. Up to you if you want to tackle that in this PR, or start another branch off this one for future PR adding that functionality. If there is a good reason to still support the RAM-based evaluation we can keep it in, otherwise it may be simpler and less code to support solely the disk-based evaluation.

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+    """
+    Computes the picai evaluation metrics provided the predicted lesion
+    detection maps and the ground truth annotations. Extends picai_eval to
+    allow multiclass evaluation
+
+    Args:
+        detection_maps (NDArray): The predicted lesion detection maps. Must
+            have shape (num_samples, num_lesion_classes, ...). The background
+            class should not be included in the detection maps
+        ground_truth_annotations (NDArray): The ground truth annotations. Must
```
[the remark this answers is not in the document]
scarere: Yeah so the idea was the num_classes = num_lesion_classes +1 where the +1 is the background class. Hence why I started using the term num_lesion_classes instead of num_classes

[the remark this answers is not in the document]
scarere: I will look to clarify this

scarere:
Ok modifying stuff to use disk ended up being a bit more work than I thought. What we have works and is a nice quick and dirty method. We can discuss a big refactor for a future PR. I will look into the comments John made to see if any of them still apply

research/picai/nnunet_scripts/eval.py
scarere: I think this is a great suggestiong. Especially since there's quite a few things I'm silencing under the nnunet umbrella (at the moment as a lot of these packages have reminders to cite them printed to console). Id say we leave it as is for now and make log file generation a seperate PR. There is some complexity in terms of figuring out where to place the log file. Placing it in the working directory feels a bit messy. Ideally we could place it in the output directory, but we receive that output directory as an argument. So we will run into not only mypy errors but potentially some import errors when using multiprocessing if we import picai_eval within a method after we know the output folder.

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,26 +1,188 @@
...
+
+    Args:
+        probability_maps (NDArray): A numpy array containing the predicted
+            probability maps. Should be shape (num_samples, num_classes, ...)
+
+    Returns:
+        NDArray: A numpy array of the detection maps for each class. Note that
+            a detection map is not created for the background class which is
+            assumed to be at index 0. Therefore the output shape is
+            (num_samples, num_classes-1)
```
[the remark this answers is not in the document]
scarere: they do my bad

research/picai/nnunet_scripts/eval.py
scarere: Leaving as is as this function got moved to the legacy folder, but addressed in the new file by passing kwargs

research/picai/nnunet_scripts/eval.py
scarere: We don't actually save the 'augmented' ground truth. We only remove the background class for the calculation. I've added a comment to clarify what is happening when we do this in the new eval script

research/picai/nnunet_scripts/eval.py
scarere: Its defined as optional in the metrics class, if none they generate one automatically. Mypy would complain so I had to narrow the type

scarere: this code is no longer in the new script

research/picai/nnunet_scripts/predict_and_eval.py
scarere: Yep

research/picai/nnunet_scripts/predict_and_eval.py
```diff
@@ -0,0 +1,252 @@
...
+        labels_one_hot = (np.arange(num_classes) == labels[..., None]).astype(int)
+        labels_one_hot = np.moveaxis(labels_one_hot, -1, 1)
+
+    # Calculate Metrics
+    t = time.time()
+    metrics = get_picai_metrics(detection_maps, labels_one_hot)
+    secs = time.time() - t
+
+    # Print metrics
+    with warnings.catch_warnings():
```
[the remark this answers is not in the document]
scarere: Warnings are gone i guess with your update to the dependencies so this isn't actaully in the new script anymore

research/picai/nnunet_scripts/eval.py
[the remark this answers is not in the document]
scarere: I was able to avoid this mess by hooking into the lower level function. I was trying to avoid doing that before, but with the new version that uses disk, there was no way around without giving up support for multiclass segmentation

research/picai/README.md
```diff
@@ -19,6 +21,30 @@ poetry install --with "picai"
 This will setup an environment with the proper dependencies to run the provided scripts out of the box. For more information about environment configuration, please refer to the [documentation](/CONTRIBUTING.md).
 
 ## Data
+### Important Definitions
```
jewelltaylor, line 24: Awesome documentation 👌

research/picai/README.md
```diff
@@ -19,6 +21,30 @@ poetry install --with "picai"
...
+
+  ![alt text](images/detmap.png)
+- **Predicted Annotations/Segmentations**: Model predictions that have been thresholded to contain only class labels. If one hot encoded they must be binary {0, 1} or boolean {False, True} tensors. If not one hot encoded they must be be tensors containing only integers that represent the class labels (eg. constrained to {0, 1, 2, ...}). An example of a binary or one-hot-encoded predicted annotation is shown below
+
+  ![alt text](images/annotation.png)
+
+Below are some terms we will use to differentiate how models are combined
+
+- **Merging:** Merging models refers to combining the weights of several different models in some way to obtain a single model that can make predictions on the same data as the original models. This is distinct from federated learning in the sense that no training is traditionally done in this regime.
+- **Ensembling:** Ensembling refers to combining the outputs of several different models in some way to acheive a single agreed upon prediction
```
jewelltaylor, line 46: nit: acheive -> achieve

research/picai/README.md
```diff
@@ -19,6 +21,30 @@ poetry install --with "picai"
...
+### Important Definitions
+Below are the terms we will use to define model outputs at various stages of
+post-processing. They are ordered sequentially from least to most processed
+
+- **Logits:** The outputs of the model <ins>prior</ins> to the activation function. Values are unconstrained (-inf, inf)
+- **Probabilities:** The outputs of the model <ins>after</ins> a normalizing activation function such as softmax or sigmoid. Values are constrained to the range (0, 1). An example for a 2d image is shown below.
+
+  ![alt text](images/probs.png)
+- **Detection Maps:** Model predictions that contain an arbitrary number of distinct detected segmentation volumes derived from the output probabilities. Values are constrained to range [0, 1]. Example for a 2d image is shown below Detected segmentation volumes are defined as:
+  - Each detected volume is a connected component that must be non-connected and non-overlapping (mutually exclusive) with other volumes of the same class. (Therefore detection maps for multiclass segementation must be one hot encoded)
```
jewelltaylor, line 33: nit: segementation -> segmentation

research/picai/README.md
```diff
@@ -19,6 +21,30 @@ poetry install --with "picai"
...
+  - The [report guided annotation](https://github.com/DIAGNijmegen/Report-Guided-Annotation) is a common api used for deriving detection maps from model output probabilities
+
+  ![alt text](images/detmap.png)
+- **Predicted Annotations/Segmentations**: Model predictions that have been thresholded to contain only class labels. If one hot encoded they must be binary {0, 1} or boolean {False, True} tensors. If not one hot encoded they must be be tensors containing only integers that represent the class labels (eg. constrained to {0, 1, 2, ...}). An example of a binary or one-hot-encoded predicted annotation is shown below
+
+  ![alt text](images/annotation.png)
+
+Below are some terms we will use to differentiate how models are combined
+
+- **Merging:** Merging models refers to combining the weights of several different models in some way to obtain a single model that can make predictions on the same data as the original models. This is distinct from federated learning in the sense that no training is traditionally done in this regime.
```
jewelltaylor, line 45: nit: "can make predictions on the same data as the original models" maybe we say "format of data" just to highlight we aren't implying it has to be the exact same data

research/picai/nnunet_scripts/eval.py
```diff
@@ -1,111 +1,332 @@
...
+    # Get list of input and output files
+    case_ids = scan_folder_for_cases(input_folder, postfixes, extensions)
+    input_files = get_case_files(input_folder, case_ids, postfixes, extensions)
+    output_files = [join(output_folder, case + ".npz") for case in case_ids]
+
+    if not exists(output_folder):
+        os.makedirs(output_folder)
+
+    # Using threading over multiprocessing since task is I/O bound
+    with concurrent.futures.ThreadPoolExecutor(num_threads) as pool:
```
jewelltaylor, line 240: 🔥

jewelltaylor (approved):
LGTM!

PR #199, opened 2024-07-30
Nnunet test

scarere:
# PR Type

# Short Description

This PR improves the functionality of nnUNetClient

- Added a shutdown method to nnUNetClient to ensure processes spawned by dataloaders are terminated
- Added nnunet_plans to properties returned by get_properties for nnUNetClient
  -  If the plans are not provided in the server config, then the client generates them and sends them back to the server
  - This allows nnunet plans to be an optional parameter as the server can request a random client to generate them
- Added the FlServerWithInitializer server class.
  - Ensures a config is sent to the client when requesting initial parameters
    - nnUNetclients no longer require a starting checkpoint because of this
    - https://github.com/adap/flower/issues/3770
  - Adds an initialize hook method to allow further server initialization before training
- Added an NnUNetServer subclass of FlServerWithInitializer
  - uses the initialize hook method to request nnunet plans from a random client if the plans were not provided in the config
- Added load_msd_data to fl4heath.utils.load_data to dynamically download and extract any MSD dataset
- Added an nnunet example which requires no set up
  - Environment variables are dynamically set
  - Works with any MSD dataset and will download and prepare the dataset on the fly
- fixed the startup messages in run_smoke_test

# Tests Added

- Added two smoke tests with the nnunet example. One for a 2d config and one for a 3d_fullres config
  - Uses the Task04_Hippocampus dataset which is the smallest of the MSD datasets and is downloaded on the fly
  - Does not check outputs, just ensures that the code runs without errors


scarere:
The nnunet smoke tests work locally for me, but none of the other tests do. I get tolerance errors on at least two of them but probably all of them. I tried checking out main and I still get the errors so I think this must be related to my setup somehow.

scarere:
Not sure why nnunet smoke tests are failing. Here is the error found in the logs
```bash
ImportError: cannot import name 'packaging' from 'pkg_resources' (/home/runner/work/FL4Health/FL4Health/.venv/lib/python3.9/site-packages/pkg_resources/__init__.py)
```

examples/nnunet_example/README.md
```diff
@@ -0,0 +1,40 @@
+# NnUNetClient Example
+
+This example demonstrates how to use the NnUNetClient to train nnunet segmentation models in a federated setting.
+
+By default this example trains an nnunet model on the Task04_Hippocampus dataset from the Medical Segmentation Decathelon (MSD). However, any of the MSD datasets can be used by specifying them with the msd_dataset_name flag for the client. To run this example first create a config file for the server. An example config has been provided in this directory. The required keys for the config are:
```
jewelltaylor, line 5: nit: Decathelon -> Decathlon

examples/nnunet_example/README.md
```diff
@@ -0,0 +1,40 @@
...
+nnunet_config: 2d
+```
+
+The only additional parameter required by nnunet is nnunet_config which is one of the official nnunet configurations (2d, 3d_fullres, 3d_lowres, 3d_cascade_fullres)
+
+One may also add the following optional keys to the config yaml file
+
+```yaml
+# Optional config parameters
+nnunet_plans: /Path/to/nnunet/plans/json
```
jewelltaylor, line 24: Is this a folder of nnunet plans or just a single one?

examples/nnunet_example/README.md
```diff
@@ -0,0 +1,40 @@
...
+python -m examples.nnunet_example.server --config_path examples/nnunet_example/config.yaml
+```
+
+Once the server has started, start the necessary number of clients specified by the n_clients key in the config file. Each client can be started by running the following command in a seperate session. To view a list of optional flags use the --help flag.
+
+```bash
+python -m examples.nnunet_example.client --dataset_path examples/datasets/nnunet
+```
+
+The MSD dataset will be downloaded and prepared automatically by the nnunet client if it does not already exist. The dataset_path flag is used as more of a data working directory by the client. The client will create nnunet_raw, nnunet_preprocessed and nnunet_results sub directories. The dataset itself will be stored in a folder within nnunet_raw. Therefore when checking if the data already exists, the client will look for the following folder '{dataset_path}/nnunet_raw/{dataset_name}'
```
jewelltaylor, line 40: Does the client create nnunet_raw, nnunet_preprocessed and nnunet_results subdirectories even in the case they already exist? The case I am thinking about is if someone already has multiple datasets stored in the structure nnunet expects

examples/nnunet_example/client.py
```diff
@@ -0,0 +1,177 @@
...
+from os.path import exists, join
+from pathlib import Path
+from typing import Union
+
+with warnings.catch_warnings():
+    # Need to import lightning utilities now in order to avoid deprecation
+    # warnings. Ignore flake8 warning saying that it is unused
+    # lightning utilities is imported by some of the dependencies
+    # so by importing it now and filtering the warnings
+    # https://github.com/Lightning-AI/utilities/issues/119
```
jewelltaylor, line 14: haahha seems like this is also from the pkg_resources module of setuptools. I was looking and it seems there was fix merged in for this that looks like its available in the [latest release](https://github.com/Lightning-AI/utilities/releases/tag/v0.11.6):
https://github.com/Lightning-AI/utilities/pull/281
also referenced here:
https://github.com/Lightning-AI/utilities/pull/292

Could be worth checking out if updating to the release fixes the error you are getting here

examples/nnunet_example/server.py
```diff
@@ -0,0 +1,121 @@
...
+from functools import partial
+from typing import Optional
+
+with warnings.catch_warnings():
+    # Need to import lightning utilities now in order to avoid deprecation
+    # warnings. Ignore flake8 warning saying that it is unused
+    # lightning utilities is imported by some of the dependencies
+    # so by importing it now and filtering the warnings
+    # https://github.com/Lightning-AI/utilities/issues/119
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
```
jewelltaylor, line 14: Same comment as above, perhaps updating the latest version will fix this

jewelltaylor: If not, no worries but worth a shot

fl4health/server/base_server.py
```diff
@@ -350,3 +350,69 @@ def _hydrate_model_for_checkpointing(self) -> nn.Module:
         model_ndarrays = parameters_to_ndarrays(self.parameters)
         self.parameter_exchanger.pull_parameters(model_ndarrays, self.server_model)
         return self.server_model
+
+
+class FlServerWithInitializer(FlServer):
+    initialized = False  # Add attribute
```
jewelltaylor, line 356: Are we sure we want this to be a class attribute instead of an instance attribute?

fl4health/server/base_server.py
```diff
@@ -350,3 +350,69 @@ def _hydrate_model_for_checkpointing(self) -> nn.Module:
...
+        parent function except we provide a config to the client when
+        requesting initial parameters
+        https://github.com/adap/flower/issues/3770
+
+        Note:
+            I have to use configure_fit to bypass mypy errors since
+            on_fit_config is not defined in the Strategy base class. The
+            downside is that configure fit will wait until enough clients for
+            training are present instead of just sampling one client. I
+            thought about defining a new init_config attribute but this
```
jewelltaylor, line 370: Thanks for the additional context! It seems like you forgot to finish the thought at the end though

fl4health/utils/msd_dataset_sources.py
```diff
@@ -0,0 +1,63 @@
+from enum import Enum
```
jewelltaylor, line 1: Nice! Thanks for going to the length so support all these datasers

fl4health/server/base_server.py
scarere: You're right instance attribute would be better, I just trying to avoid writing an __init__ method to keep it concise

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -521,3 +537,72 @@ def update_before_epoch(self, epoch: int) -> None:
     def get_client_specific_logs(self) -> Tuple[str, List[Tuple[LogLevel, str]]]:
         lr = self.optimizers["global"].param_groups[0]["lr"]
         return f" Current LR: {lr}", []
+
+    def get_properties(self, config: Config) -> Dict[str, Scalar]:
+        """
+        Return properties (sample counts and nnunet plans) of client.
+
+        If nnunet plans are not provided by the server, creates a new set of
```
jewelltaylor, line 545: Great functionality!

jewelltaylor (commented):
Looks great! I will take one last pass once you have the chance to address a few of the small comments I had. But this is pretty much good to go from my POV

fl4health/server/base_server.py
jewelltaylor: Yeah I feel you hahah I don't think I have ever used a class attribute. It would be nice if you could do this for instance attribute

examples/nnunet_example/README.md
scarere: just a single one, I will change default

examples/nnunet_example/README.md
scarere: The client doesn't create the subdirectories, just the example does. When using the nnunet client, its up to the user to ensure their environment variables are set and their datasets are set up. In the example, to make things easy for people, they just specify a directory for nnunet, and the example creates the subdirectories if they don't already exist. If the user already has an nnunet_raw subdirectory in that folder and already has the MSD dataset that was requested, then the program just uses that, it won't redownload and reprocess the same dataset. The only restriction is that nnunet_raw, nnunet_preprocessed and nnunet_results folders must all be in the same directory. I will change the wording to make it clearer that it is the example that creates these subdirectories and downloads the data.

examples/nnunet_example/client.py
scarere: So we are already on the newest version for lightning utilities so we'll probably have to wait until the next release comes out. I just checked and we definitely still need it. However it seems like the updated dependencies somehow got rid of the warnings I was getting from importing flwr so I can take that out and only silence the lightning utilities warnings

fl4health/server/base_server.py
scarere: This is an example where that docstring parsing pr I mentioned would be useful within fl4health. The exact same length docstring occurs 3 times in this file. I think typically people would use **kwargs to avoid this, but as David has mentioned it obscures the arguments of the function.

research/picai/fl_nnunet/nnunet_client.py
scarere: It's a little hacky right now due to mypy complaints. I'd like to get flwr to support something more official so we don't have to have this in our repo

examples/nnunet_example/README.md
jewelltaylor: Sounds good, thanks for the clarification!

examples/nnunet_example/client.py
jewelltaylor: Okay, no worries!

fl4health/server/base_server.py
jewelltaylor: Added a ticket in clickup, we can discuss more details in our meeting!

jewelltaylor (approved):
LGTM!

PR #204, opened 2024-08-06
Added progress bar and fixed bugs

scarere:
# PR Type

# Short Description

Added an optional progress bar to basic_client. Also fixed two bugs that I found in nnunet client.
-Optional progress bar shows training progress
-Fixed a bug nnunet client would train for way too many steps per epoch
-Fixed a bug where the lr scheduler did not anneal the LR fast enough


scarere:
sorry about the messy commit history. I used rebase instead of merge without knowing what I was doing.

emersodb:
> sorry about the messy commit history. I used rebase instead of merge without knowing what I was doing.

I am also terrible at using rebase. I think my rebase success rate is like 60%. 😂

fl4health/clients/basic_client.py
```diff
@@ -612,13 +630,16 @@ def train_by_epochs(
                 Loss is a dictionary of one or more losses that represent the different components of the loss.
         """
         self.model.train()
-        local_step = 0
         for local_epoch in range(epochs):
+            local_step = 0  # Reset local step
```
jewelltaylor, line 634: I should add some documentation to make this more clear. But here we assume that local_step the step number in this current round of FL for a given client. So we don't want to reset this every epoch. The reason being is that for the APFL method, we only update the alpha parameter (representing the mix of client and global models) on the first step of the first epoch. We don't want this to happen multiple times if there is more than one epoch. Here is the relevant code in [ApflClient](https://github.com/VectorInstitute/FL4Health/blob/dc574f70c88cac5eb3456168d867976684dfc5d1/fl4health/clients/apfl_client.py#L35). You should be able to back out the step for your purposes with local_step mod batches_per_epoch

jewelltaylor (changes requested):
Looks great other than the one change request I had. I can add some better documentation to that part of the code once we get this merged in so its more clear.

fl4health/clients/basic_client.py
scarere: Ok I can change this back to the way it was and add a bit of clarifying documentation

fl4health/clients/basic_client.py
```diff
@@ -382,23 +388,26 @@ def _handle_logging(
...
 
         initial_log_str = f"Current FL Round: {str(current_round)}\t" if current_round is not None else ""
         initial_log_str += f"Current Epoch: {str(current_epoch)}" if current_epoch is not None else ""
 
         # Maybe add client specific info to initial log string
-        client_str, client_logs = self.get_client_specific_logs()
-        initial_log_str += client_str
+        client_str, client_logs = self.get_client_specific_logs(current_round, current_epoch, logging_mode)
 
         if initial_log_str != "":
```
emersodb, line 398: This is perhaps a silly question that I missed in reviewing before, but it seems like there isn't a setting in which this log string is empty (""), since we set it on line 352? If that's correct, maybe we just ditch this if?

emersodb: Ah I see...if both current_round and current_epoch are None then this could be empty.

fl4health/clients/basic_client.py
```diff
@@ -612,13 +630,16 @@ def train_by_epochs(
...
-        local_step = 0
+        steps_this_round = 0  # Reset number of steps this round
         for local_epoch in range(epochs):
             self.train_metric_manager.clear()
             self.train_loss_meter.clear()
             # update before epoch hook
             self.update_before_epoch(epoch=local_epoch)
-            for input, target in self.train_loader:
+            # Print initial log string on epoch start, not epoch end
+            self._handle_logging({}, {}, current_round, local_epoch)
```
emersodb, line 640: Perhaps instead of "co-opting" `_handle_logging` in this way, which is a little different than how it was intended, we can create a separate function for this? I know it will have a little code duplication, but we can also reuse that separate function in `_handle_logging` if we want.

fl4health/clients/basic_client.py
```diff
@@ -1100,3 +1137,32 @@ def update_before_epoch(self, epoch: int) -> None:
...
+        Returns:
+            Iterable: an iterator which acts exactly like the original
+                iterable, but prints a dynamically updating progress bar every
+                time a value is requested. Or the original iterable if
+                self.progress_bar is False
+        """
+        if not self.progress_bar:
+            return iterable
+        else:
+            # Create a clean looking tqdm instance that matches the flwr logging
```
emersodb, line 1159: Cool!

research/picai/fl_nnunet/config.yaml
```diff
@@ -1,7 +1,12 @@
...
 n_clients: 1
-nnunet_config: 2d
-# nnunet_plans: /home/shawn/Code/nnunet_storage/nnUNet_preprocessed/Dataset012_PICAI-debug/nnUNetPlans.json
 n_server_rounds: 1
-local_epochs: 2
-# starting_checkpoint: /home/shawn/Code/nnunet_storage/nnUNet_results/Dataset012_PICAI-debug/nnUNetTrainer_1epoch__nnUNetPlans__2d/fold_0/checkpoint_best.pth
+server_address: '0.0.0.0:8080'
+local_epochs: 3
+# local_steps: 50
+#starting_checkpoint: /home/shawn/Code/nnunet_storage/nnUNet_results/Dataset004_Hippocampus/nnUNetTrainer_1epoch__nnUNetPlans__2d/fold_0/checkpoint_best.pth
```
emersodb, line 7: Really dumb, but a space between the # and "starting_checkpoint" would be nice 🙂

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -311,6 +314,7 @@ def setup_client(self, config: Config) -> None:
         # Get nnunet config
         self.nnunet_config = get_valid_nnunet_config(narrow_config_type(config, "nnunet_config", str))
 
+        # Check if dataset fingerprint has been extracted
```
emersodb, line 317: This seems like a duplicate comments?

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -329,6 +334,18 @@ def setup_client(self, config: Config) -> None:
                 device=self.device,
             )
 
+            # Need to modify num_epochs before initializing so that LRScheduler
+            # recieves the right number of epochs
```
emersodb, line 338: recieves -> receives 🙂

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -329,6 +334,18 @@ def setup_client(self, config: Config) -> None:
...
+            # This will anneal LR to near zero each round
+            # Default nnunet behaviour is do decrease LR every 250 steps
+            # Maybe LRScheduler should just be an argument
+            local_epochs, local_steps, _, _ = self.process_config(config)
```
emersodb, line 342: I might be missing something here, but i think you did this extraction already on line 326?

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -329,6 +334,18 @@ def setup_client(self, config: Config) -> None:
                 device=self.device,
             )
 
+            # Need to modify num_epochs before initializing so that LRScheduler
+            # recieves the right number of epochs
+            # This will anneal LR to near zero each round
+            # Default nnunet behaviour is do decrease LR every 250 steps
+            # Maybe LRScheduler should just be an argument
```
emersodb, line 341: If this is a note to self type question, there is a special comment tag of TODO: Maybe ... that has some nice properties with vscode where you can count how many are in the code etc. if you want to use it

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -362,7 +379,7 @@ def setup_client(self, config: Config) -> None:
 
     def predict(self, input: TorchInputType) -> Tuple[TorchPredType, Dict[str, torch.Tensor]]:
         """
-        Generate model outputs. Overridden because nnunets output lists when
+        Generate model outputs. Overridden because nnunet's output lists when
```
emersodb, line 382: I actually think the previous version of this was correct here 🙂. I don't care too much either way though ha

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -329,6 +334,18 @@ def setup_client(self, config: Config) -> None:
                 device=self.device,
             )
 
+            # Need to modify num_epochs before initializing so that LRScheduler
+            # recieves the right number of epochs
+            # This will anneal LR to near zero each round
+            # Default nnunet behaviour is do decrease LR every 250 steps
```
emersodb, line 340: Perhaps "...is to decrease LR every..."

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -542,13 +559,44 @@ def empty_cache(self) -> None:
...
+        not the LR from the previous round
+        """
+        self.nnunet_trainer.lr_scheduler.step(0)
+
+    def update_before_step(self, step: int) -> None:
+        """
+        Update the learning rate. Need to define this in case train by steps is being
+        used. This is the current step for the entire round, not the current epoch.
+        """
+        # By default nnunet lr schedulers use epochs
```
emersodb, line 574: We have this comment above 
```
Default nnunet behaviour is do decrease LR every 250 steps
```
but here we're saying that the LR schedulers use epochs by default, which seems contradictory. However, it's possible I'm missing a nuance here?

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -580,8 +628,6 @@ def get_properties(self, config: Config) -> Dict[str, Scalar]:
         planner = ExperimentPlanner(dataset_name_or_id=self.dataset_id, plans_name="temp_plans")
         with contextlib.redirect_stdout(None):  # Prevent print statements from experiment planner
             plans = planner.plan_experiment()
-
-        plans["plans_name"] = "nnUNetPlans"  # Set plans name to default
```
emersodb, line 584: Just want to confirm we wanted to delete this line? 🙂

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -593,7 +639,7 @@ def get_properties(self, config: Config) -> Dict[str, Scalar]:
         # plans since client needs to be initialized to get properties
         config["nnunet_plans"] = plans_bytes
         properties = super().get_properties(config)
-        properties["nnunet_plans"] = plans_bytes
+        properties["nnunet_plans"] = pickle.dumps(plans_bytes)
```
emersodb, line 642: Maybe a clarifying comment here might be helpful. It's sort of funny that we're doing a `pickle.dumps(plans)`, which should return a binary object, and then we're doing a `pickle.dumps` again on that binary object.

research/picai/fl_nnunet/start_server.py
```diff
@@ -73,11 +73,6 @@ def main(config: dict, server_address: str) -> None:
         # Of course nnunet stores their pytorch models differently.
         params = ndarrays_to_parameters([val.cpu().numpy() for _, val in model["network_weights"].items()])
     else:
-        # raise Exception(
```
emersodb, line 76: Did they end up fixing this on their end?

research/picai/fl_nnunet/nnunet_client.py
scarere: There is a nuance and I've clarified in the comments. Default nnunet behaviour is to define 1 epoch as 250 steps, train for "1000" epochs and decrease the learning rate every "epoch". My current implementation defines an epoch as what it actually is ((num_samples / batch_size) steps), so we are technically doing it differently right now. My next PR after this one is merged is going to address this by updating based on the number of steps instead of the number of epochs. Additionally I'm considering making a LrScheduler base class that takes the server round, current step, current epoch etc. as arguments when taking a step, as well as some subclasses that will just wrap easily around pytorch lr schedulers. This will be in the next PR

research/picai/fl_nnunet/nnunet_client.py
scarere: Thank you, that line helps prevent unecessary preprocessing and files from being generated

research/picai/fl_nnunet/nnunet_client.py
scarere: I swear I've deleted this change like 4 times and it keeps sneaking back in with the merges. We're not supposed to be double pickling the plans. I have no idea why the code still runs when this error is present.

research/picai/fl_nnunet/start_server.py
scarere: They haven't even commented on the issue yet. Likely because I tagged it as a feature request. Maybe should have tagged it as a bug but can't change it now. I've added a comment with the issue link just to keep us up to date

emersodb: Gotcha...a comment is good!

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -624,13 +614,18 @@ def get_properties(self, config: Config) -> Dict[str, Scalar]:
         if not self.fingerprint_extracted:
             self.maybe_extract_fingerprint()
 
-        # Create experiment planner and plans
+        # Create experiment planner and plans. Plans name must be temp_plans so that we
+        # can safely delete the plans file generate by the experiment planner
```
emersodb, line 618: generate -> generated

emersodb (approved):
Too small cosmetic things, but looks ready to go from me. I'd make sure @jewelltaylor  is good with the new changes before merging just to be sure his change request is satisfied.

jewelltaylor (approved):
Looks good to me!

PR #205, opened 2024-08-07
Modified base server to accept sequences of checkpointers

scarere:
# PR Type

# Short Description

Super small PR. Changes the typing of the checkpointer argument for FlServer and FlServerWithCheckpointing to also accept sequences of checkpointers. Each checkpointer is run individually. For example you could pass both Latest and Best checkpointers.

Modified the basic example to save both the latest and best checkpoint seperately to ensure that the multiple checkpointing functionality works

# Tests Added

I did not add a test, but since I updated the basic example which is used as a smoke test


fl4health/server/base_server.py
```diff
@@ -43,16 +43,20 @@ def __init__(
...
+                sequence to checkpoint based on multiple criteria. Defaults to
+                None.
             metrics_reporter (Optional[MetricsReporter], optional): A metrics reporter instance to record the metrics
                 during the execution. Defaults to an instance of MetricsReporter with default init parameters.
         """
 
         super().__init__(client_manager=client_manager, strategy=strategy)
         self.wandb_reporter = wandb_reporter
-        self.checkpointer = checkpointer
+        self.checkpointer = [checkpointer] if isinstance(checkpointer, TorchCheckpointer) else checkpointer
```
jewelltaylor, line 59: Should we assert that the names of the checkpointers are different to avoid users accidentally overwriting the checkpoints?

scarere: Added clarification in comments

jewelltaylor (approved):
LGTM!

PR #222, opened 2024-08-19
Changed nnunet LR Schedule and fixed memory leak

scarere:
# PR Type

# Short Description

1. Updated the nnunet client LR Scheduler
    - Now updates every step
    - Takes into account the total number of rounds when determining max step
    
2. Fixed an issue where RAM usage would steadily increase throughout training and cause OOM
    - Not sure if this issue was specific to the nnunet client or affects all clients
    - RAM usage would steadily increase after each round eventually leading to an OOM.
    - Resolved issue by running garbage collection at the beginning of each round. Implemented this only for the nnunet client using a hook.

3. Added Gradient Clipping to nnunet client
    - By default, nnunet 2.5.1 does gradient some gradient clipping during training
    - Added this to the nnunet client so that we could be consistent
    - In order to add this functionality, had to add another hook function to the basic client
    - This repository definitely needs a cleanup/refactor. Particularly Basic Client

4. Changed how plans files are named
    - Previously, if the client was asked to generate a plans file, it was named the nnunet default which is nnUNetPlans
    - The final plans file had a prefix to specify it was for FL, however if in a previous experiment another client was asked to generate the default plans, then those plans would have had the same name and been overwritten.
    - source plans are now named after the dataset from which they were generated

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -608,6 +608,7 @@ class BasicClient(NumPyClient):
 
         # Compute backward pass and update parameters with optimizer
         losses.backward["backward"].backward()
+        self.transform_gradients(losses)
         self.optimizers["global"].step()
 
         return losses, preds
@@ -1192,3 +1193,14 @@ class BasicClient(NumPyClient):
                 "bar_format": f"{LOG_COLORS['INFO']}INFO{LOG_COLORS['RESET']}" + " :        {l_bar}{bar}{r_bar}",
             }
             return tqdm(iterable, **kwargs)
+
+    def transform_gradients(self, losses: TrainingLosses) -> None:
+        """
+        Hook function for model training only called after backwards pass but before
+        optimizer step. Useful for transforming the gradients (such as with gradient
+        clipping) before they are applied to the model weights.
+
+        Args:
+            losses (TrainingLosses): The losses object from the train step
+        """
+        pass
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -266,7 +266,7 @@ class BasicClient(NumPyClient):
             raise ValueError("Must specify either local_epochs or local_steps in the Config.")
 
         # Update after train round (Used by Scaffold and DP-Scaffold Client to update control variates)
-        self.update_after_train(local_steps, loss_dict)
+        self.update_after_train(local_steps, loss_dict, config)
 
         # Check if we should run an evaluation with validation data after fit
         # (for example, this is used by FedDGGA)
@@ -1115,7 +1115,7 @@ class BasicClient(NumPyClient):
         """
         pass
 
-    def update_after_train(self, local_steps: int, loss_dict: Dict[str, float]) -> None:
+    def update_after_train(self, local_steps: int, loss_dict: Dict[str, float], config: Config) -> None:
         """
         Hook method called after training with the number of local_steps performed over the FL round and
         the corresponding loss dictionary. For example, used by Scaffold to update the control variates
@@ -1127,6 +1127,7 @@ class BasicClient(NumPyClient):
             local_steps (int): The number of steps so far in the round in the local
                 training.
             loss_dict (Dict[str, float]): A dictionary of losses from local training.
+            config (Config): The config from the server
         """
         pass
 
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -386,8 +386,8 @@ class BasicClient(NumPyClient):
                 training. Defaults to None.
         """
 
-        log_str = f"Current FL Round: {str(current_round)}\t" if current_round is not None else ""
-        log_str += f"Current Epoch: {str(current_epoch)}" if current_epoch is not None else ""
+        log_str = f"Current FL Round: {int(current_round)} " if current_round is not None else ""
+        log_str += f"Current Epoch: {int(current_epoch)} " if current_epoch is not None else ""
 
         # Maybe add client specific info to initial log string
         client_str, _ = self.get_client_specific_logs(current_round, current_epoch, logging_mode)
@@ -635,14 +635,16 @@ class BasicClient(NumPyClient):
         return losses, preds
 
     def train_by_epochs(
-        self, epochs: int, current_round: Optional[int] = None
+        self,
+        epochs: int,
+        current_round: Optional[int] = None,
     ) -> Tuple[Dict[str, float], Dict[str, Scalar]]:
         """
         Train locally for the specified number of epochs.
 
         Args:
             epochs (int): The number of epochs for local training.
-            current_round (Optional[int]): The current FL round.
+            current_round (Optional[int], optional): The current FL round.
 
         Returns:
             Tuple[Dict[str, float], Dict[str, Scalar]]: The loss and metrics dictionary from the local training.
@@ -658,7 +660,7 @@ class BasicClient(NumPyClient):
             # update before epoch hook
             self.update_before_epoch(epoch=local_epoch)
             for input, target in self.maybe_progress_bar(self.train_loader):
-                self.update_before_step(steps_this_round)
+                self.update_before_step(steps_this_round, current_round)
                 # Assume first dimension is batch size. Sampling iterators (such as Poisson batch sampling), can
                 # construct empty batches. We skip the iteration if this occurs.
                 if self.is_empty_batch(input):
@@ -670,7 +672,7 @@ class BasicClient(NumPyClient):
                 losses, preds = self.train_step(input, target)
                 self.train_loss_meter.update(losses)
                 self.update_metric_manager(preds, target, self.train_metric_manager)
-                self.update_after_step(steps_this_round)
+                self.update_after_step(steps_this_round, current_round)
                 self.total_steps += 1
                 steps_this_round += 1
             metrics = self.train_metric_manager.compute()
@@ -684,13 +686,16 @@ class BasicClient(NumPyClient):
         return loss_dict, metrics
 
     def train_by_steps(
-        self, steps: int, current_round: Optional[int] = None
+        self,
+        steps: int,
+        current_round: Optional[int] = None,
     ) -> Tuple[Dict[str, float], Dict[str, Scalar]]:
         """
         Train locally for the specified number of steps.
 
         Args:
             steps (int): The number of steps to train locally.
+            current_round (Optional[int], optional): The current FL round
 
         Returns:
             Tuple[Dict[str, float], Dict[str, Scalar]]: The loss and metrics dictionary from the local training.
@@ -706,7 +711,7 @@ class BasicClient(NumPyClient):
         self._log_header_str(current_round)
         for step in self.maybe_progress_bar(range(steps)):
 
-            self.update_before_step(step)
+            self.update_before_step(step, current_round)
 
             try:
                 input, target = next(train_iterator)
@@ -727,7 +732,7 @@ class BasicClient(NumPyClient):
             losses, preds = self.train_step(input, target)
             self.train_loss_meter.update(losses)
             self.update_metric_manager(preds, target, self.train_metric_manager)
-            self.update_after_step(step)
+            self.update_after_step(step, current_round)
             self.total_steps += 1
 
         loss_dict = self.train_loss_meter.compute().as_dict()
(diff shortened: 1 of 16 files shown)


scarere:
I should note that I don't think I've fixed a memory leak so much as I have addressed an issue with memory not being cleaned up enough frequently causing OOM's. Also worth noting that the gen2 objects are supposed to be cleaned up automatically after every 10 collections/cleanups of gen1 (https://docs.python.org/3/library/gc.html). For some reason this is not occurring, python refuses to clean up gen 3 objects after the first round (I would presume this could be further narrowed down to the first pass) unless its done manually

scarere:
Also the nnunet client would randomly just not work sometimes. It might have been OOM issues, but it also might have been something to do with torch.compile. The nnunet model is compiled with torch.compile which for some reason spawns like 30 processes. If one of these processes were terminated or killed (say perhaps due to an OOM issue?), it would automatically generate an error due to the change that flwr made causing all process terminations or inturruptions to raise an exception. I was getting around this issue with the dataloaders which also spawn threads by temporarily changing the signal handlers back to the python defaults before creating the dataloaders and spawning new processes and then changing them back. I've now moved the entire setup within the scope of the default signal handlers so hopefully if one of the pytorch dynamo processes is terminated, it won't generate an exception and we won't get random exceptions and the client and smoke tests will be more consistent

fl4health/clients/basic_client.py
```diff
@@ -386,8 +386,8 @@ def _log_header_str(
                 training. Defaults to None.
         """
 
-        log_str = f"Current FL Round: {str(current_round)}\t" if current_round is not None else ""
-        log_str += f"Current Epoch: {str(current_epoch)}" if current_epoch is not None else ""
+        log_str = f"Current FL Round: {int(current_round)} " if current_round is not None else ""
```
jewelltaylor, line 389: Just curious if there was any reason for the cast to int? Is there situations where this isn't an int or None?

scarere: I think mypy was complaining at one point about current round being an optional int. Probably was due to something else and I just never changed it back

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -661,3 +721,24 @@ def shutdown(self) -> None:
...
+        # The above issue seems to be the situation I was in.
+        gc.collect()  # Cleans up unused variables
+        # As the linked issue above points out, calling gc.freeze() greatly reduces the
+        # overhead of garbage collection. (from 1.5s to 0.005s)
+        if current_server_round == 2:
+            # Collect runs even faster if we freeze after the end of the first iteration
+            # Likely because a lot of variables are created in the first pass
+            # If we freeze before the first pass, gc.collect has to check all those
+            # variables
+            gc.freeze()
```
jewelltaylor, line 737: Maybe I have to look into the issue and details about the Python garbage collector further, but it looks like gc.freeze() is going to temporarily disable the automatic collection of unreachable objects by the garbage collector until gc.unfreeze() is called. If gc.freeze() is called, calls to gc.collect() only trigger the collection cycle is triggered but will not remove any objects. Are we calling gc.unfreeze() anywhere or is this somehow not required? Does gc.freeze() have anything to do with the OOM memory that is described or does it just make garbage collection faster? Why only on current_server_round=2 ?

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -199,9 +188,71 @@ def get_model(self, config: Config) -> nn.Module:
...
+
+        Args:
+            source_plans (Dict): The nnunet plans dict that is being modified
+            max_factor (float): The fraction of the approx total number of voxels to
+                use as the maximum. Defaults to 0.05 corresponding to 5% as this is the nnunet default in 2.5.1
+
+        Returns:
+            float: Maximum number of voxels that can be in a batch
+        """
+        # Need to determine input dimensionality
```
jewelltaylor, line 244: It seems as though this if statement is doing the same thing regardless of NnUNetConfig._3D_FULLRES.value

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -661,3 +721,24 @@ def shutdown(self) -> None:
...
+            # If we freeze before the first pass, gc.collect has to check all those
+            # variables
+            gc.freeze()
+
+    def transform_gradients(self, losses: TrainingLosses) -> None:
+        """
+        Apply the gradient clipping performed by the default nnunet trainer. This is
+        the default behaviour for nnunet 2.5.1
+        """
+        torch.nn.utils.clip_grad_norm_(self.model.parameters(), 12)
```
jewelltaylor, line 744: Should the max_norm be configurable at all? Since norm has a dependence on the tensor size as well as the scale of the input data

research/picai/fl_nnunet/nnunet_utils.py
```diff
@@ -179,6 +180,12 @@ def __iter__(self) -> DataLoader:  # type: ignore
         # mypy gets angry that the return type is different
         return self
 
+    def shutdown(self) -> None:
```
jewelltaylor, line 183: Some elaboration on what is involved in the shutdown of different augmenters may be helpful

research/picai/fl_nnunet/nnunet_client.py
scarere: Maybe I should explain further in the comments, but gc.freeze doesn't stop gc.collect from working, it essentially takes a snapshot of what is currently in memory and makes it permanent. Any newly created objects after the freeze will still be cleaned up by gc.collect. The benefit of this is that gc.collect doesn't bother to check the objects that were frozen, therefore it only has to check items in memory that were created after the freeze, making things significantly faster. So the goal with gc.freeze is to ideally freeze only the objects that we are going to need in memory throughout the entire training process, so that gc.collect can focus on cleaning up anything that is only needed temporarily. Since we call collect right before the freeze, we can be pretty confident that most of whats in memory is going to be needed long term. The reason that we do this on current_server_round=2 is because we want to do the freeze after the first pass has occured. The first pass seems to create a lot of variables that we are going to need in memory throughout the entire training process, and so we would prefer those to be frozen so that gc doesn't have check them over and over every time we collect. The more items that gc is tracking, the slower the collect call is.

research/picai/fl_nnunet/start_client.py
```diff
@@ -69,7 +69,7 @@ def main(
         # BaseClient Args
         device=DEVICE,
         metrics=metrics,
-        progress_bar=True,
+        progress_bar=False,
```
jewelltaylor, line 72: I am not sure if this has anything to do with the issues that are popping up but I noticed the the number of processes to be used by the data augmenter. It is being set in this function in [nnUNetv2](https://github.com/MIC-DKFZ/nnUNet/blob/master/nnunetv2/utilities/default_n_proc_DA.py#L5). I am using a small machine to test and the call to os.cpu_count() reports much higher than what I have and even the default of 12 is more than the logical cpus I have. With the additional overhead of the server process, and eventually more than 1 client\, we need to make sure this is being set appropriately. I suggest you add a command line argument that allows you to pass the number of processes explicitly  and sets the environment variable (`$nnUNet_n_proc_DA`) that is leveraged in the linked function.

research/picai/fl_nnunet/nnunet_client.py
scarere: gc unfreeze unfreezes the objects in memory that were frozen. So maybe it would be a good idea to unfreeze and then collect on train end in case the user has anything else they want to do. Not sure if that would clean up memory or not

jewelltaylor: Ah I see so it will collect AND remove as per usual for object allocated after the gc.freeze() call. Seems like we would still want to call gc.unfreeze() and then gc.collect() at the end as you mentioned.

fl4health/clients/basic_client.py
```diff
@@ -1185,3 +1193,14 @@ def maybe_progress_bar(self, iterable: Iterable) -> Iterable:
                 "bar_format": f"{LOG_COLORS['INFO']}INFO{LOG_COLORS['RESET']}" + " :        {l_bar}{bar}{r_bar}",
             }
             return tqdm(iterable, **kwargs)
+
+    def transform_gradients(self, losses: TrainingLosses) -> None:
```
jewelltaylor, line 1197: Great addition! This can be leveraged by the scaffold client to simplify the code. I put a [ticket](https://app.clickup.com/t/8689drd6a) in the backlog for me to port the client to leverage this once this PR is merged in.

jewelltaylor:
All in all, things look good! A few small comments but mostly clarifying questions for my understanding. Perhaps we can meet tomorrow after the meeting with Masoom and discuss some of the issues that popped up so you can get me on the same page and we can discuss how to look into them going forward. Won't delay getting this merged in but perhaps we put some other tickets on the backlog to do some further investigating

research/picai/fl_nnunet/nnunet_client.py
jewelltaylor: If we just use gc.collect() without the gc.freeze() part does the OOM still happen? Or is this just for performance benefits?

scarere: If we use only gc collect without freezing, it still resolves the OOM issue as the memory is still being cleaned up (particularly the generation 2 memory objects which are typically checked very infrequently). The calls to gc collect however take about 1.5s on my small dummy dataset. Using gc freeze reduces the overhead of gc collect to i think 0.005 seconds.

research/picai/fl_nnunet/nnunet_client.py
scarere: Good catch

research/picai/fl_nnunet/nnunet_client.py
scarere: I'll add it as a parameter but nnunet uses 12 regardless of the input data or scale. My guess is that 12 is a conservatively large norm and that this is in there just as a precaution

research/picai/fl_nnunet/start_client.py
scarere: I can confirm that the number of processes spawned by the augmenters is not causing the OOM issues as that was one of the first things I checked. PyTorch itself launches way more processes under the hood for reasons I don't understand, but I think it has something to do with the model being compiled. That being said, I agree that the number of processes used for augmentation should be a configurable parameter. Not sure how os.cpu_count() would report more than you have though? Is it possible that you have some high performance cpu's that are being counted as multiple cpu's?

jewelltaylor: I had 2 cpus with 4 cores each with hyper-threading so 16 logical cpus and os.cpu_count() was returning 32. I am not sure if its returning all of the cpus on a given node or what. 

Thanks for adding though, this will be really helpful when we are doing FL with more than 1 client anyways

research/picai/fl_nnunet/nnunet_client.py
jewelltaylor: I am little hesitant to add the additional complexity but if it's making a meaningful change on the actual dataset I am okay with it

research/picai/fl_nnunet/nnunet_client.py
```diff
@@ -639,25 +714,54 @@ def get_properties(self, config: Config) -> Dict[str, Scalar]:
...
+            gc.freeze()
+
+    def transform_gradients(self, losses: TrainingLosses) -> None:
+        """
+        Apply the gradient clipping performed by the default nnunet trainer. This is
+        the default behaviour for nnunet 2.5.1
+        """
+        torch.nn.utils.clip_grad_norm_(self.model.parameters(), self.max_grad_norm)
+
+    def update_after_train(self, local_steps: int, loss_dict: Dict[str, float], config: Config) -> None:
```
jewelltaylor, line 762: One thing I just thought about is that if we don't have full client participation, this will only be called on clients who are sampled in the final round. We could do an extra round at the end of the training with full participation, enforce no training happens and call the gc.unfreeze(). Or we could require full participation as a short term fix and work towards supporting client sampling. What do you think?

scarere: Would moving this to client.shutdown resolve the issue?

scarere: The only downside to that is if you don't have full client participation AND are doing something with the client afterwards before shutting it down. The only thing that I can think that you would be doing with the client afterwards would be finetuning and inference, in which case maybe its not terrible if the frozen objects stay frozen in memory since their probably needed in both anyways

jewelltaylor: Yeah I think that should work! I think that is fine for now and we can make note of it to change if becomes a problem

scarere:
> Pretty much good to go! I just want to get your take on the one comment I left before getting this merged in.
> 
> I also tried to detail some of the discussions today in a clickup ticket to monitor and follow up on. If there is anything I didn't capture that you think should be added, please feel free to add it or let me know!
> 
> https://app.clickup.com/t/8689e8urk

I made a few edits, I'm freezing the objects left in memory after the first round not the second. The reason I check for current_round=2 is because the hook function occurs before training

jewelltaylor (approved):
LGTM!

PR #225, opened 2024-08-27
Created reset signal handler decorator

scarere:
# PR Type

# Short Description

Was currently bypassing as signal handling [issue](https://github.com/adap/flower/issues/3837) with flwr by temporarily resetting the signal handlers before child processes were spawned by the nnunet client. There was one child subprocess spawned during fingerprint extraction that was accidentally being spawned before the signal handlers were temporarily reset.

Ensured this subprocess was inheriting the default signal handlers. Also improved the implementation by creating a decorator function that would do this so that it doesn't have to be in the main function code. Since the errors were happening stochastically, its hard to conclusively say that this has resolved the issue, but I'm pretty confident this will fix it. We should however continue to monitor any StopIteration errors generated by flwr as a result of a subprocess being terminated or inturrupted.

Note that flwr has indicated that they intend to fix this signal handling issue in the next release at which point we can refactor this code out.

### DIFF
--- a/research/picai/fl_nnunet/nnunet_utils.py
+++ b/research/picai/fl_nnunet/nnunet_utils.py
@@ -1,4 +1,6 @@
+import signal
 import warnings
+from collections.abc import Callable
 from logging import ERROR
 from typing import Any, Dict, List, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


research/picai/fl_nnunet/nnunet_utils.py
```diff
@@ -197,3 +199,31 @@ def __init__(self, loss: nn.Module, **kwargs: Any) -> None:
...
         return self.loss(pred, target)
+
+
+def use_default_signal_handlers(fn: Callable) -> Callable:
+    """
+    This is a decorator that resets the SIGINT and SIGTERM signal handlers back to the
+    python defaults for the execution of the method
+
+    flwr 1.9.0 overrides the default signal handlers with handlers that raise an error
+    on any inturruption or termination. Since nnunet spawns child processes which
```
emersodb, line 210: interruption 🙂

emersodb (approved):
Looks good to me. I like the use of the decorators!

research/picai/fl_nnunet/nnunet_utils.py
scarere: fixed.

PR #226, opened 2024-08-28
Adding a README to the examples folder.

emersodb:
# PR Type
Documentation

# Short Description

Clickup Ticket(s): [Link(s) if applicable.](https://app.clickup.com/t/8689erm57)

Adding a README, along the lines specified in the ticket below, to the examples folder.

# Tests Added

NA

### DIFF
--- a/examples/README.MD
+++ b/examples/README.MD
@@ -1,0 +1,353 @@
+## FL4Health Examples
+
+This folder contains a large collection of example implementations of the various methods and capabilities of the FL4Health library. There are several broad categories covered by these examples, including different state-of-the-art global and personalized FL (pFL) methods, aggregation strategies, pre-processing techniques and more.
+
+Below is an index of these examples, followed by a brief description of what they demonstrate. A README is included in each example folder with instructions on how to run the example and a bit more detail on what the example implements. The examples are listed in alphabetical order. For more details on the library, installing the appropriate environment etc. see [README.MD](../README.md)
+
+<table>
+<tr>
+<th style="text-align: left; width: 250px"> Example Folder </th>
+<th style="text-align: center; width: 350px"> Notes </th>
+</tr>
+<tr>
+<td>
+
+[ae_examples](ae_examples)
+</td>
+<td>
+This folder contains examples implementing federated training of auto-encoders (AEs), variational auto-encoders (VAEs), and conditional variational auto-encoders (CVAEs).
+</td>
+</tr>
+<tr>
+<td>
+
+[apfl_example](apfl_example)
+</td>
+<td>
+This implements an example of training a model using the pFL approach, APFL. Server-side modifications are minimal, but the client-side training modifications are handled by the appropriate parent class.
+
+Dataset: MNIST
+</td>
+</tr>
+<tr>
+<td>
+
+[basic_example](basic_example)
+</td>
+<td>
+This is one of the simplest FL examples. It demonstrates federally training a model with basic FedAvg. It also provides an example of using server-side checkpointing and evaluating on a client-side test dataset in every round.
+
+Dataset: CIFAR 10
+</td>
+</tr>
+<tr>
+<td>
+
+[ditto_example](ditto_example)
+</td>
+<td>
+This example implements the pFL method Ditto to train individual client models on a heterogeneous version of MNIST. Ditto is one of a small set of methods that actually requires two optimizers for training.
+
+Dataset: MNIST
+</td>
+</tr>
+<tr>
+<td>
+
+[dp_fed_examples](dp_fed_examples)
+</td>
+<td>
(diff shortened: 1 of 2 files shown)


examples/README.MD
```diff
@@ -0,0 +1,353 @@
...
+Dataset: MNIST
+</td>
+</tr>
+<tr>
+<td>
+
+[dp_fed_examples](dp_fed_examples)
+</td>
+<td>
+The examples in this folder are basic implementations of differentially private (DP) FL training. There are two levels of DP guarantees. Client-level DP refers to guarantees protecting <b>client</b> participation in FL training. Instance-level DP refers to guarantees protecting <b>datapoint</b> participation in FL training. Instance-level DP is classically achieved using DP-SGD on the client-side. The client-level examples include adaptive clipping implementations.
```
scarere, line 60: Maybe one sentence explaining what differential privacy is? What does it mean to 'protect client participation' or 'protect datapoint participation'? This is also partially my own curiosity. Is it guarantees against information leakage from the client dataset to the server/other clients?

examples/README.MD
```diff
@@ -0,0 +1,353 @@
...
+Dataset: Mimic3 Tabular Dataset
+</td>
+</tr>
+<tr>
+<td>
+
+[fedbn_example](fedbn_example)
+</td>
+<td>
+This is an example of implementing FedBN. The implementation is quite simple in that it is essentially the basic_example but with an exclusion parameter exchanger.
```
scarere, line 115: nit: 'of implementing the FedBN strategy' ... 'exclusion parameter exchanger to exclude batch normalization layers when exchanging weights between the clients and the server'. A little nitpicky but I think we should specify that it's a strategy. End bit is to implicitly clarify that BN stands for batch normalization

examples/README.MD
```diff
@@ -0,0 +1,353 @@
+## FL4Health Examples
+
+This folder contains a large collection of example implementations of the various methods and capabilities of the FL4Health library. There are several broad categories covered by these examples, including different state-of-the-art global and personalized FL (pFL) methods, aggregation strategies, pre-processing techniques and more.
```
scarere, line 3: Maybe add a link to the main README to point the user to the descriptions of the different strategies so that we can get away with not defining them here

scarere (commented):
John might have better comments as he is more well versed on FL strategies, but overall was pretty clear to for me who has limited knowledge on the subject which I think is a good thing.

examples/README.MD
scarere: By descriptions I mainly mean just an expansion of the acronyms

examples/README.MD
```diff
@@ -0,0 +1,353 @@
...
+Dataset: CIFAR-10
+</td>
+</tr>
+<tr>
+<td>
+
+[warm_up_example](warm_up_example)
+</td>
+<td>
+This folder provides several implementations of what we refer to as warm-start FL. That is, a model is federally trained using some mechanism, perhaps FedAvg. Thereafter, that pre-trained model is used to initialize further FL training using a different method. The examples herein also provide illustrations of "model surgery" wherein only portions of the warm model are transferred into the new model that will received continued FL training.
```
jewelltaylor, line 346: nit: that will *receive continued training

examples/README.MD
```diff
@@ -0,0 +1,353 @@
...
+</tr>
+<tr>
+<td>
+
+[feature_alignment_example](feature_alignment_example)
+</td>
+<td>
+Provides an implementation of automatic feature alignment for tabular datasets. When data is represented in dataframes, this implementation automatically unifies the shared features and feature spaces to prepare the datasets for FL prior to training
+
+Dataset: Mimic3 Tabular Dataset
```
jewelltaylor, line 106: nit: The rest of the datasets that are acronyms are capitalized

jewelltaylor (approved):
Other than the small changes mentioned by Shawn and I, this looks good!

examples/README.MD
emersodb: Sure. Will add a sentence there. In brief, there are different "threat-models" that change the privacy you're guaranteeing. However, most of what we consider is a "third-party" perspective, which is an outside observer who only receives the final model. Client-level DP prevents this "third-party" observer from being able to tell if a particular client participated in training the model with certainty. Instance-level DP prevents this observer from being able to tell if a particular data point was used during training.

examples/README.MD
emersodb: Will add.

examples/README.MD
emersodb: Sure. Will add here as well.

examples/README.MD
emersodb: 👍

examples/README.MD
emersodb: 👍

scarere (approved):

PR #230, opened 2024-09-03
Update requirements

jewelltaylor:
# PR Type

# Short Description
Update torch, torchvision and tensorflow and remove pinned setuptools version.

Although it does not explicitly say in the [pypi page](https://pypi.org/project/torch/), it seems that that torch 2.3.0 =< is not compatible with python 3.9 on macOS. I spent a while trying to figure this out and the only thing that worked is using python 3.10+. So I updated the pyproject.toml to reflect this. 

Clickup Ticket: [Update torch and tensorflow](https://app.clickup.com/t/8689je1et)

### DIFF
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -1,13 +1,12 @@
-# only has to pass for python 3.10
 name: Smoke Tests
 
 on:
   push:
     branches:
-      main
+      - main
   pull_request:
     branches:
-      main
+      - main
 
 jobs:
   test:
@@ -15,31 +14,34 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+
       - name: Set up Python 3.10
         uses: actions/setup-python@v5
         with:
-          python-version: 3.10
-      # Display the Python version being used
+          python-version: "3.10"
+
       - name: Display Python version
         run: python -c "import sys; print(sys.version)"
+
       - name: Set up file descriptor limit
-        run: |
-          ulimit -n 4096
+        run: ulimit -n 4096
+
       - name: Install and configure Poetry
         uses: snok/install-poetry@v1
         with:
           virtualenvs-create: true
           virtualenvs-in-project: true
+
       - name: Set up cache
         uses: actions/cache@v4
         id: cached-poetry-dependencies
         with:
           path: .venv
           key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/poetry.lock') }}
+
       - name: Install dependencies
-        run: poetry install --with "dev, test"
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --with "dev, test"
+
       - name: Run Script
-        run: |
-          source .venv/bin/activate
-          python -m tests.smoke_tests.run_smoke_test
+        run: poetry run python -m tests.smoke_tests.run_smoke_test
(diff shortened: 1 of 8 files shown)


scarere:
You should update the tests to use python 3.10 as I believe they are currently using 3.9. This can be done by modifying the workflow files in .github

jewelltaylor:
> You should update the tests to use python 3.10 as I believe they are currently using 3.9. This can be done by modifying the workflow files in .github

Good catch, completely slipped my mind!

.github/workflows/smoke_tests.yaml
```diff
@@ -1,45 +1,47 @@
...
         uses: actions/cache@v4
         id: cached-poetry-dependencies
         with:
           path: .venv
           key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/poetry.lock') }}
+
       - name: Install dependencies
-        run: poetry install --with "dev, test"
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --with "dev, test"
```
emersodb, line 44: Maybe a dumb question, but does this `run:...` need to be indented to be "inside" the if?

.github/workflows/static_code_checks.yaml
```diff
@@ -41,10 +41,7 @@ jobs:
         uses: pypa/gh-action-pip-audit@v1.1.0
         with:
           virtual-environment: .venv/
-          # Ignoring vulnerability that just popped up in pytorch
-          # versions < 2.2.0, GHSA-pg7h-5qx3-wjr3
-          # Poetry currently resolves torch version 2.1.2, not sure why, some
-          # other package must be preventing the pytorch version from being
-          # upgraded
+          # Ignoring vulnerability that for cryptography
```
emersodb, line 44: The two lines of this comment are slightly hard to read. Maybe clean it up slightly?

pyproject.toml
```diff
@@ -12,22 +12,17 @@ classifiers = [
...
 flwr = "1.9.0"
 opacus = "^1.3.0"
-torch = "^2.1.1"
+torch = "^2.3.0"
 pycyclops = "^0.2.5"
 dp-accounting = "^0.4.3"
 torchmetrics = "^1.3.0"
 aiohttp = "^3.9.3"
 urllib3 = "^2.2.2"
-setuptools = "69.5.1"
```
emersodb, line 26: Is this resolved now?

emersodb (approved):
Some small comments. Nothing major. Just need to merge in main I think and see that everything passes and should be good to go.

.github/workflows/smoke_tests.yaml
jewelltaylor: I thought that as well initially since I am not super familiar with the syntax of the workflow yaml files. But looking into it, it actually should be at the same level of indentation. The run line will be executed iff the if statement above it (at the same level of indentation) is true

pyproject.toml
jewelltaylor: Yeah with the update of torch, this is resolved! in torch 2.2.0 they fully move over to the new api from the deprecated one.

More on this: https://github.com/vllm-project/vllm/issues/4961#issuecomment-2141323698

.github/workflows/static_code_checks.yaml
jewelltaylor: yeah hahah this is definitely confusing, will update!

pyproject.toml
emersodb: Good to know!

.github/workflows/smoke_tests.yaml
emersodb: Interesting. Good to know as well!

PR #231, opened 2024-09-03
Update scaffold

jewelltaylor:
# PR Type

# Short Description

In PR #222, @scarere introduced a transform_gradients hook. This could be leveraged by the ScaffoldClient to avoid having to redefine the train_step function to modify the gradient to combat client drift. 

Clickup Ticket(s): [Convert Scaffold to leverage transform_gradient hook](https://app.clickup.com/t/8689drd6a)

### DIFF
--- a/fl4health/clients/scaffold_client.py
+++ b/fl4health/clients/scaffold_client.py
@@ -17,7 +17,6 @@ from fl4health.parameter_exchange.parameter_exchanger_base import ParameterExcha
 from fl4health.parameter_exchange.parameter_packer import ParameterPackerWithControlVariates
 from fl4health.utils.losses import LossMeterType, TrainingLosses
 from fl4health.utils.metrics import Metric
-from fl4health.utils.typing import TorchInputType, TorchPredType, TorchTargetType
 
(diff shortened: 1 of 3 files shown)


emersodb (approved):
Nice!

emersodb:
One "trick" that I do when I have PRs that build off one another (and this is likely kind of hacky, but it makes them a little cleaner) like you're doing with your change to bring in the upgrade to python 3.10, is to merge that PR into this branch, then target the PR to the pending PR you're working off of (the python 3.10 upgrade PR in this case). The diffs will then only include the changes you intend to make with this PR.

jewelltaylor:
> One "trick" that I do when I have PRs that build off one another (and this is likely kind of hacky, but it makes them a little cleaner) like you're doing with your change to bring in the upgrade to python 3.10, is to merge that PR into this branch, then target the PR to the pending PR you're working off of (the python 3.10 upgrade PR in this case). The diffs will then only include the changes you intend to make with this PR.

That is a really good point! I am not sure why I even based this branch off the other since they are independent anyway. Will keep this in mind for the future!

PR #232, opened 2024-09-04
Basic client checkpointing and LR Scheduler

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Add Per Round Checkpointing to to BasicClient and FlServerWithCheckpointing](https://app.clickup.com/t/8689k9uca) and [BasicClient LR Scheduler Support](https://app.clickup.com/t/8689k9ypn).

In #227, @scarere migrated NnunetClient and NnunetServer to the main library. In order to do experiments on the vector cluster (where long running jobs are guaranteed to be pre-empted), we need to add support for checkpointing. Per round checkpointing is already supported in PicaiClient and PicaiServer. This PR involves moving that functionality to the BasicClient and filling in any gaps to make it compatible with all FL methods. Since the NnunetClient leverages a LR Scheduler we need to also checkpoint it. So this PR also includes an initial functionality to support LR Schedulers in BasicClient.

# Tests Added

Work in progress. Tests to be added soon.


scarere:
maybe this is a naive question, but would it make more sense to merge this into [nnunet_core_refactor](https://github.com/VectorInstitute/FL4Health/pull/227) since it is branched off from there? Will we run into issues trying to merge these two branches directly to main separately?

fl4health/clients/basic_client.py
```diff
@@ -1105,6 +1151,29 @@ def get_model(self, config: Config) -> nn.Module:
...
+                to the optimizer we are optionally defining a learning rate
+                scheduler for.
+            config (Config): The config from the server.
+
+        Returns:
+            Union[None, _LRScheduler]: Client learning rate schedulers.
+        """
+        return None
+
+    def update_lr_schedulers(self) -> None:
```
scarere, line 1170: Might be useful to have access to state information here such as the current epoch and step. However i know the upcoming basic client refactor will be addressing this to some extent. Also might be useful to mention that self.total_steps is available in the docstring

PR #234, opened 2024-09-10
Basic client checkpointing and LR Scheduling

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Add Per Round Checkpointing to to BasicClient and FlServerWithCheckpointing](https://app.clickup.com/t/8689k9uca) and [BasicClient LR Scheduler Support](https://app.clickup.com/t/8689k9ypn).

In #227, @scarere migrated NnunetClient and NnunetServer to the main library. In order to do experiments on the vector cluster (where long running jobs are guaranteed to be pre-empted), we need to add support for checkpointing. Per round checkpointing is already supported in PicaiClient and PicaiServer. This PR involves moving that functionality to the BasicClient and filling in any gaps to make it compatible with BasicClient and NnunetClient. Since the NnunetClient leverages a LR Scheduler we need to also checkpoint it. So this PR also includes an initial functionality to support LR Schedulers in BasicClient.

The future BasicClient refactor will build on this PR by supporting checkpointing more broadly as well as other features. This is an intermediate PR where I had to make some compromises so it is manageable size and get it up in a reasonable amount of time.

# Tests Added
- Smoke tests to verify training model from scratch to a given point is the same as training model from scratch to an intermediate point and then from an intermediate point on yields the same results
- Updated tests for PerRoundCheckpointer to reflect its simplification


fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,51 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+
+        assert "lr_schedulers" in ckpt and isinstance(ckpt["lr_schedulers"], dict)
+        assert "client_name" in ckpt and isinstance(ckpt["client_name"], str)
+        assert "total_steps" in ckpt and isinstance(ckpt["total_steps"], int)
+        assert "metrics_reporter" in ckpt and isinstance(ckpt["metrics_reporter"], MetricsReporter)
+        assert "optimizers_state" in ckpt and isinstance(ckpt["optimizers_state"], dict)
+
+        self.total_steps = ckpt["total_steps"]
+        self.client_name = ckpt["client_name"]
+        self.total_steps = ckpt["total_steps"]
```
jewelltaylor, line 1317: bad

fl4health/clients/nnunet_client.py
```diff
@@ -58,6 +57,7 @@
 class NnunetClient(BasicClient):
     def __init__(
         self,
+        data_path: Path,
```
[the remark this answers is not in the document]
jewelltaylor: Merged in changes, should be good now

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,51 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+    def save_checkpoint(self) -> None:
+        """
+        Saves checkpoint dict consisting of client name, total steps, lr schedulers,
+            metrics reporter and optimizers state. Method can be overridden to augment saved checkpointed state.
+        """
+
+        assert self.per_round_checkpointer is not None
+
+        ckpt = {
+            "lr_schedulers": self.lr_schedulers,
```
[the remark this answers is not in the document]
jewelltaylor: Good call

fl4health/clients/nnunet_client.py
```diff
@@ -259,16 +264,20 @@ def get_lr_scheduler(self, config: Config) -> _LRScheduler:
...
+        local_epochs, local_steps, _, _ = self.process_config(config)
         if local_steps is not None:
-            self.steps_per_round = local_steps
+            steps_per_round = local_steps
+            total_steps = int(config["n_server_rounds"]) * steps_per_round
         elif local_epochs is not None:
-            self.steps_per_round = local_epochs * len(self.train_loader)
-        self.max_steps = int(config["n_server_rounds"]) * self.steps_per_round
+            steps_per_round = local_epochs * len(self.train_loader)
+            total_steps = int(config["n_server_rounds"]) * steps_per_round
```
scarere, line 273: this line can be moved outside of the if else block since it occurs either way

fl4health/clients/nnunet_client.py
```diff
@@ -259,16 +264,20 @@ def get_lr_scheduler(self, config: Config) -> _LRScheduler:
...
+        else:
+            raise ValueError("One of local steps and local epochs must be None")
```
scarere, line 275: I think you might mean 'One of local_steps and local_epochs must specified'. This exception is only reached when they are both None.

examples/apfl_example/server.py
```diff
@@ -68,6 +68,7 @@ def main(config: Dict[str, Any]) -> None:
     )
 
     server.metrics_reporter.dump()
```
emersodb, line 70: Maybe we can put this in the backlog. It feels like we can put the metrics_reporter dump in the server.shutdown() routine?

scarere (approved):
Looks good to me as a temporary solution

fl4health/clients/basic_client.py
```diff
@@ -1105,6 +1144,37 @@ def get_model(self, config: Config) -> nn.Module:
         """
         raise NotImplementedError
 
+    def get_lr_scheduler(self, optimizer_key: str, config: Config) -> Union[None, _LRScheduler]:
```
emersodb, line 1147: I think perhaps, in a future iteration, we should use the get/set paradigm that we've used for the optimizer construction/initialization? That is, we have a `get_lr_scheduler` function that the user overrides to define the LRs that they want to use. The `set_lr_schedulers` function is called by the `setup_client` function and invokes the `get_lr_scheduler` method. The reason I like this is because the `set_lr_schedulers` function can do sanity checks by default and encapsulate the behavior you have above. That is,
```
def set_lr_schedulers(self, config: Config) -> Optional[_LRScheduler]:
       # Ensure that the optimizers have been initialized prior to setting schedulers
       assert self.optimizers
        # set schedulers for the optimizers
        self.lr_schedulers = {}
        for optimizer_key in self.optimizers.keys():
            lr_scheduler = self.get_lr_scheduler(optimizer_key, config)
            if lr_scheduler is not None:
                self.lr_schedulers[optimizer_key] = lr_scheduler
```

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,55 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
             losses (TrainingLosses): The losses object from the train step
         """
         pass
+
+    def save_checkpoint(self) -> None:
```
emersodb, line 1280: This is a bit generic and might be misleading for a user in the sense that this is actually a very specific type of checkpointing. Maybe we call this something like `save_client_state` in a future refactor?

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,55 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+        """
+
+        assert self.per_round_checkpointer is not None
+
+        ckpt = {
+            "lr_schedulers_state": {key: scheduler.state_dict() for key, scheduler in self.lr_schedulers.items()},
+            "total_steps": self.total_steps,
+            "client_name": self.client_name,
+            "metrics_reporter": self.metrics_reporter,
+            "optimizers_state": {key: opt.state_dict()["state"] for key, opt in self.optimizers.items()},
```
emersodb, line 1293: super minor, but perhaps we can expand opt to optimizer?

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,55 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+        """
+        assert self.per_round_checkpointer is not None and self.per_round_checkpointer.checkpoint_exists()
+
+        ckpt = self.per_round_checkpointer.load_checkpoint()
+
+        assert "lr_schedulers_state" in ckpt and isinstance(ckpt["lr_schedulers_state"], dict)
+        assert "client_name" in ckpt and isinstance(ckpt["client_name"], str)
+        assert "total_steps" in ckpt and isinstance(ckpt["total_steps"], int)
+        assert "metrics_reporter" in ckpt and isinstance(ckpt["metrics_reporter"], MetricsReporter)
+        assert "optimizers_state" in ckpt and isinstance(ckpt["optimizers_state"], dict)
```
emersodb, line 1313: The `narrow_config_type` function, essentially does this type of thing. Maybe we should name it something more general and use it here, since it's just effectively working on a dictionary at the end of the day?

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,55 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+
+        self.client_name = ckpt["client_name"]
+        self.total_steps = ckpt["total_steps"]
+        self.metrics_reporter = ckpt["metrics_reporter"]
+
+        # Optimizer is updated in setup_client to reference model weights from server
+        # Thus, only optimizer state (per parameter values such as momentum)
+        # should be loaded
+        for opt, state in zip(self.optimizers.values(), ckpt["optimizers_state"].values()):
```
emersodb, line 1322: This zip makes me slightly nervous, as I'm not 100% confident that dictionary.values() is always going to have the same order between two different dictionaries. I don't think that is guaranteed, but I might be wrong? I would suggest we use `ckpt["optimizers_state"].items()` and then use the key to get the correct optimizer from `self.optimizers`.

fl4health/clients/basic_client.py
```diff
@@ -248,6 +266,11 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
         if not self.initialized:
             self.setup_client(config)
 
+            # If per_round_checkpointer not None and checkpoint exists load it and set attributes.
+            # Model not updated because FL restarted from most recent FL round (redo pre-empted round)
+            if self.per_round_checkpointer is not None and self.per_round_checkpointer.checkpoint_exists():
+                self.load_checkpoint()
```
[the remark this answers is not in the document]
emersodb: I'm curious to hear your thoughts, but I wonder if, rather than always loading from a checkpoint if it exists and a checkpointer has been defined, we should have a bool for loading from a state checkpoint? It's not perfect to add another constructor variable, but the situation I imagine is that you start a run with a `per_round_checkpointer` for preemption safety, but there is a pre-existing state file that you don't know about or thought would just be overridden, but this would actually load the state without the user explicitly requesting state loading? I might be over-thinking it, but figured I'd record the thought here.

fl4health/clients/nnunet_client.py
```diff
@@ -259,16 +264,20 @@ def get_lr_scheduler(self, config: Config) -> _LRScheduler:
...
+        else:
+            raise ValueError("One of local steps or local epochs must be specified")
+
+        total_steps = int(config["n_server_rounds"]) * steps_per_round
 
-        # Create and return LR Scheduler. This is nnunet default for version 2.5.1
-        return PolyLRScheduler(
-            self.optimizers["global"], initial_lr=self.nnunet_trainer.initial_lr, max_steps=self.max_steps
+        # Create and return LR Scheduler Wrapper for the PolyLRScheduler so that it is
+        # compatible with Torch LRScheduler
```
emersodb, line 278: Maybe we want to add back Shawn's comment on the nnunet default?

fl4health/clients/nnunet_client.py
```diff
@@ -624,23 +630,6 @@ def empty_cache(self) -> None:
         elif self.device.type == "mps":
             torch.mps.empty_cache()
 
-    def update_after_step(self, step: int, current_round: Optional[int] = None) -> None:
```
emersodb, line 627: So in this code, regardless of whether we were training by epochs or by steps in the client, the lr scheduler was being provided with self.total_steps as part of its update. Now, we won't be providing that regardless of the training scheme. 

I don't know the inner workings of the PolyLRScheduler, but at the very least, we're not preserving the warning that Shawn had here. Perhaps we should fold this in by overriding the `update_lr_schedulers` function?

fl4health/server/base_server.py
```diff
@@ -1,25 +1,33 @@
...
 from fl4health.parameter_exchange.parameter_exchanger_base import ParameterExchanger
 from fl4health.reporting.fl_wandb import ServerWandBReporter
 from fl4health.reporting.metrics import MetricsReporter
 from fl4health.server.polling import poll_clients
 from fl4health.strategies.strategy_with_poll import StrategyWithPolling
 from fl4health.utils.metrics import TEST_LOSS_KEY, TEST_NUM_EXAMPLES_KEY, TestMetricPrefix
+from fl4health.utils.random import generate_hash
+
+
+def get_initial_model_parameters(client_model: nn.Module) -> Parameters:
```
emersodb, line 28: I might be missing something, but we have `get_all_model_parameters` in our utilities that should already do this?

fl4health/server/base_server.py
```diff
@@ -350,17 +365,198 @@ def __init__(
...
+            # parent method includes call to report metrics to wandb if specified
+            history, elapsed_time = super().fit(num_rounds, timeout)
+
+        return history, elapsed_time
+
+    def fit_with_per_epoch_checkpointing(self, num_rounds: int, timeout: Optional[float]) -> Tuple[History, float]:
+        """
+        Runs federated learning for a number of rounds. Heavily based on the fit method from the base
+        server provided by flower (flwr.server.server.Server) except that it is resilient to pre-emptions.
+        It accomplishes this by checkpointing the sever state each round. In the case of pre-emption,
```
emersodb, line 429: server state each round 🙂

fl4health/server/base_server.py
```diff
@@ -350,17 +365,198 @@ def __init__(
...
+        assert "metrics_reporter" in ckpt and isinstance(ckpt["metrics_reporter"], MetricsReporter)
+        assert "history" in ckpt and isinstance(ckpt["history"], History)
+
+        log(INFO, f"Loading server state from checkpoint at {self.per_round_checkpointer.checkpoint_path}")
+
+        self.current_round = ckpt["current_round"]
+        self.server_name = ckpt["server_name"]
+        self.metrics_reporter = ckpt["metrics_reporter"]
+        self.history = ckpt["history"]
+        self.parameters = get_initial_model_parameters(ckpt["model"])
```
emersodb, line 558: I don't think we want to rely on the model architecture to get the parameters, but rather should dump the self.parameters that the server has. The reason for this is situations where we are exchanging more information with the server than just model parameters (FedProx, SCAFFOLD, etc.). This will fail when that is happening. I'm also not sure that this will work properly or the way we want it to when we're doing partial exchange?

fl4health/utils/nnunet_utils.py
```diff
@@ -410,3 +410,21 @@ def write(self, buf: str) -> int:
...
+class PolyLRSchedulerWrapper(_LRScheduler):
+    def __init__(
+        self, optimizer: torch.optim.Optimizer, initial_lr: float, max_steps: int, exponent: float = 0.9
+    ) -> None:
+        self.optimizer = optimizer
+        self.initial_lr = initial_lr
+        self.max_steps = max_steps
+        self.exponent = exponent
+        self._step_count: int
+        super().__init__(optimizer, -1, False)
```
emersodb, line 424: Maybe a comment here on the -1?

fl4health/utils/nnunet_utils.py
```diff
@@ -410,3 +410,21 @@ def write(self, buf: str) -> int:
...
+
+    @no_type_check
```
emersodb, line 426: Perhaps we can leave a comment as to why we can't do type checking here?

emersodb (commented):
Just leaving some comments for future consideration. There are definitely a few things I think we should probably clean up, as I think there are some edge cases that might bite us 🙂

examples/apfl_example/server.py
```diff
@@ -68,6 +68,7 @@ def main(config: Dict[str, Any]) -> None:
...
+    server.shutdown()
```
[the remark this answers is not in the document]
jewelltaylor: Added to backlog

fl4health/clients/basic_client.py
```diff
@@ -1206,3 +1276,55 @@ def transform_gradients(self, losses: TrainingLosses) -> None:
...
+            "client_name": self.client_name,
+            "metrics_reporter": self.metrics_reporter,
+            "optimizers_state": {key: opt.state_dict()["state"] for key, opt in self.optimizers.items()},
+        }
+
+        self.per_round_checkpointer.save_checkpoint(ckpt)
+
+        log(INFO, f"Saving client state to checkpoint at {self.per_round_checkpointer.checkpoint_path}")
+
+    def load_checkpoint(self) -> None:
```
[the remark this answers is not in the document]
jewelltaylor: Agreed!

fl4health/checkpointing/checkpointer.py
```diff
@@ -172,32 +170,26 @@ def __init__(self, checkpoint_dir: Path, checkpoint_name: Path) -> None:
...
 
         Args:
             checkpoint_dict (Dict[str, Any]): A dictionary with string keys and values of type
                 Any representing the state to checkpoint.
-
-        Raises:
-            NotImplementedError: To be implemented by child classes.
         """
-        raise NotImplementedError
+        torch.save(checkpoint_dict, self.checkpoint_path)
```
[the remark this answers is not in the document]
jewelltaylor: Agreed! As part of the refactor, I want to explore unifying the checkpointers (metric-based and "per round"). I think the experience of passing a sequence of checkpointers is better from a users perspective than dealing with two different ones. So perhaps I can generalize this class and leverage it what I stated prior ends up making sense once I look into more deeply

fl4health/clients/basic_client.py
```diff
@@ -64,15 +68,27 @@ def __init__(
...
 
         self.data_path = data_path
         self.device = device
         self.metrics = metrics
         self.checkpointer = checkpointer
         self.progress_bar = progress_bar
+        self.client_name = client_name if client_name is not None else generate_hash()
 
-        self.client_name = generate_hash()
+        self.per_round_checkpointer: Union[None, PerRoundCheckpointer]
```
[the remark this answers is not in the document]
scarere: Since the project requirements have been upgraded to python 3.10, we can also use `|` in place of Union. I would propose `PerRoundCheckpointer | None`, it's less characters, imo equally clear, and removes the need to import Union or Optional which if we're lucky removes a line in the imports.

fl4health/clients/basic_client.py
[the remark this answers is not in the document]
jewelltaylor: It definitely could be put in setup_client and maybe thats the direction we go, but I opted not too because I thought it made sense to have the load_checkpoint and save_checkpoint in the sample place (fit method).

fl4health/clients/basic_client.py
jewelltaylor: Good idea!

fl4health/clients/basic_client.py
jewelltaylor: As of Python 3.7 dictionaries are ordered (ie maintains insertion order and successive calls to values/items/keys will return the same ordering), so I am pretty sure its okay. But I can definitely add a note to change it over so things are more explicit

jewelltaylor: ckpt["optimizers_state"] is generated by iterating through self.optimizers, so it must have the same ordering

fl4health/clients/nnunet_client.py
jewelltaylor: The reason I had to remove the warning is that the LR Scheduler has attributes (ie max_steps) that are not shared by all LR Scheduler. So we couldn't have general logic to raise an exception if total steps exceeds max_steps. But I agree I can add this check in by overriding the method

fl4health/server/base_server.py
jewelltaylor: Forgot about that!

fl4health/server/base_server.py
jewelltaylor: Absolutely, definitely agree! Just for this PR, it seemed to much too tackle supporting all methods, so this was one of the compromises I made and pushed it to the refactor

PR #235, opened 2024-09-12
Adaptive Ditto and MR-MTL

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8688d097n)

The idea here is to make the adaptive drift loss weight proposed in the FedProx paper available to the Ditto and MR-MTL clients. That is, instead of a static drift loss weight, the server can be used to adapt that weight based on the loss it is seeing during training. This was useful in FedProx and I'm curious whether it will also be useful for Ditto and MR-MTL. My intention is to design several experiments to test this out, along with some other hypothesis in future PRs.

# Tests Added

Migrated the tests for Ditto, MR-MTL, FedProx, and FENDA+Ditto to work in the new structures.


fl4health/clients/nnunet_client.py
```diff
@@ -201,7 +201,7 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
         # Set the number of processes for each dataloader.
         if self.n_dataload_proc is None:
             # Nnunet default is 12 or max cpu's. We decrease max by 1 just in case
-            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)
+            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)  # type: ignore
```
emersodb, line 204: I had to do this for my machine because `sched_getaffinity` is apparently not available for the older intel macs in Python 3.10 🤔. I'm not sure why. I think it's fine to leave this here, as it seems to work for you guys and I likely won't be running this until I get my new machine anyway.

fl4health/utils/metrics.py
```diff
@@ -13,6 +13,7 @@
 
 
 class TestMetricPrefix(Enum):
+    __test__ = False
```
emersodb, line 16: This had to be added so that this class gets ignored when pytest collects tests

fl4health/clients/nnunet_client.py
scarere: All good with me. Maybe add a short comment as to why its being ignored? Maybe worth looking into late.

emersodb: Yeah good call. Will add a comment. I looked into it a bit and there really isn't a great alternative for older MacOS setups. Most people just end up punting to getting all cpus rather than those accessible from the process thread. There is a new function in os for python 3.13 and up that seems to do this better for MacOS but it's not available below that...

fl4health/clients/adaptive_drift_constraint_client.py
```diff
@@ -0,0 +1,214 @@
...
+        # This is the loss value to be sent back to the server on which adaptation decisions will be made.
+        self.loss_for_adaptation: float
+        # Function to compute the penalty loss.
+        self.penalty_loss_function = WeightDriftLoss(self.device)
+
+    def get_parameters(self, config: Config) -> NDArrays:
+        """
+        Packs the parameters and training loss into a single NDArrays to be sent to the server for aggregation. If
+        the client has not been initialized, this means the server is requesting parameters for initialization and
+        just the model parameters are sent. When using the FedAvgWithAdaptiveConstraint strategy. This should not
```
jewelltaylor, line 80: nit: When using the FedAvgWithAdaptiveConstraint strategy, this should not (seems like you want a comma and not period here)

fl4health/clients/adaptive_drift_constraint_client.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        # These are the tensors that will be used to compute the penalty loss
+        self.drift_penalty_tensors: List[torch.Tensor]
+        # Exchanger with packing to be able to exchange the weights and auxiliary information with the server for
+        # adaptation
+        self.parameter_exchanger: FullParameterExchangerWithPacking
+        # Weight on the penalty loss to be used in backprop. This is what might be adapted via server calculations
+        self.drift_penalty_weight: float
+        # This is the loss value to be sent back to the server on which adaptation decisions will be made.
+        self.loss_for_adaptation: float
```
jewelltaylor, line 72: nit: In the docstring above, you refer to this as: "the criterion loss (i.e. loss without the penalty)". Maybe its worth updating the language to be more consistent so its clear that is the loss that is being referred to

fl4health/clients/adaptive_drift_constraint_client.py
```diff
@@ -0,0 +1,214 @@
...
+            TrainingLosses: an instance of TrainingLosses containing backward loss and additional losses indexed
+                by name. Additional losses includes penalty loss.
+        """
+        loss, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
+        if additional_losses is None:
+            additional_losses = {}
+
+        additional_losses["loss"] = loss.clone()
+        # adding the vanilla loss to the additional losses to be used by update_after_train for potential adaptation
+        additional_losses["adaptation_loss"] = loss.clone()
```
jewelltaylor, line 162: nit: Adaptation loss is a bit confusing to me. It seems like it could mean either the loss that we are using to adapt the constant applied to the weight drift loss or the weight drift loss which is adapted based on the constant from the server. I kind of liked criterion loss or loss without penalty as you described it in the init docstring

fl4health/clients/adaptive_drift_constraint_client.py
```diff
@@ -0,0 +1,214 @@
...
+            local_steps (int): The number of steps so far in the round in the local training.
+            loss_dict (Dict[str, float]): A dictionary of losses from local training.
+            config (Config): The config from the server
+        """
+        assert "adaptation_loss" in loss_dict
+        # Store current loss which is the vanilla loss without the penalty term added in
+        self.loss_for_adaptation = loss_dict["adaptation_loss"]
+        super().update_after_train(local_steps, loss_dict, config)
+
+    def compute_penalty_loss_and_store(self, additional_losses: Dict[str, torch.Tensor]) -> torch.Tensor:
```
jewelltaylor, line 198: nit: I think having the storing side effect and returning a value may be a bit confusing. How about we have this just return the value and have the compute_training_loss handle packing in the additional loss as its primarily responsible for doing that

jewelltaylor: very minor nit: Also, did we envision that this is something that will need to be overridden? I definitely agree breaking up functionality into smaller methods can enhance code reusability and readability, especially if those methods are expected to be overridden or extended in the future. However, if we don't anticipate the need for overriding, keeping the logic within a single method might simplify understanding and maintenance by not fragmenting the logic making the execution flow hard to follow

fl4health/clients/ditto_client.py
```diff
@@ -9,31 +9,35 @@
...
+        metrics_reporter: Optional[MetricsReporter] = None,
+        progress_bar: bool = False,
     ) -> None:
         """
         This client implements the Ditto algorithm from Ditto: Fair and Robust Federated Learning Through
         Personalization. The idea is that we want to train personalized versions of the global model for each client.
         So we simultaneously train a global model that is aggregated on the server-side and use those weights to also
         constrain the training of a local model. The constraint for this local model is identical to the FedProx loss.
 
+        NOTE: lambda, the drift loss weight, is initially set and potentially adapted by the server akin to the
```
jewelltaylor, line 38: Maybe it worth mentioning that using adaptation is optional and specified in the strategy. Just so users are aware they can use the original Ditto if they need to do so

fl4health/clients/ditto_client.py
```diff
@@ -119,51 +126,51 @@ def get_parameters(self, config: Config) -> NDArrays:
...
+            packed_params = self.parameter_exchanger.pack_parameters(global_model_weights, self.loss_for_adaptation)
+            return packed_params
 
     def set_parameters(self, parameters: NDArrays, config: Config, fitting_round: bool) -> None:
         """
-        The parameters being pass are to be routed to the global model and saved as the initial global model tensors to
-        be used in a penalty term in training the local model. In the first fitting round, we assume the both the
-        global and local models are being initialized and use the FullParameterExchanger() to set all model weights.
+        Assumes that the parameters being passed contain model parameters concatenated with a penalty weight. They are
+        unpacked for the clients to use in training. The parameters being pass are to be routed to the global model.
```
jewelltaylor, line 147: nit: The parameters being passed (instead of pass)

fl4health/clients/ditto_client.py
```diff
@@ -177,40 +184,27 @@ def initialize_all_model_weights(self, parameters: NDArrays, config: Config) ->
...
-            current_round (Optional[int]): The current FL round.
-
-        Returns:
-            Tuple[Dict[str, float], Dict[str, Scalar]]: The loss and metrics dictionary from the local training.
-                Loss is a dictionary of one or more losses that represent the different components of the loss.
-        """
-        # Need to also set the global model to train mode
-        self.global_model.train()
-        return super().train_by_epochs(epochs, current_round)
+    def save_initial_global_tensors(self) -> None:
```
jewelltaylor, line 187: nit: Maybe set_initial_global_tensors is a more fitting name?

fl4health/clients/fenda_ditto_client.py
```diff
@@ -25,18 +26,23 @@ def __init__(
...
+        progress_bar: bool = False,
         freeze_global_feature_extractor: bool = False,
     ) -> None:
         """
-        This client implements the Ditto algorithm from Ditto: Fair and Robust Federated Learning Through
-        Personalization with a local FENDA model. The idea is that we want to train a local FENDA model
-        along with the global model for each client. So we simultaneously train a global model that is
+        This client implements a combination of the Ditto algorithm from Ditto: Fair and Robust Federated Learning
+        Through Personalization with FENDA-FL models. The idea is that we want to train a local FENDA model
+        along with the global model for each client. We simultaneously train a global model that is
```
jewelltaylor, line 36: In light of our conversation, perhaps adding this (no need to use words exactly): 
The global ditto model consists of a feature extractor and classification head, where the feature extractor shares the architecture of the global and local feature extractors for FENDA.
After this: The idea is that we want to train a local FENDA model along with the global model for each client.

And before getting into specifics about the nature of the constraint (because it depends on the mode of operation), be clear that there are two different modes of operation.
1. global DITTO model feature extractor SETS AND FREEZES weights of global FENDA feature extractor, local components of FENDA model are trained and an additional drift loss is computed between the local FENDA feature extractor and global feature extractor
2. global DITTO model feature extractor INITIALIZES weights of global FENDA feature extractor, both local and global components of FENDA are trained and a drift loss is calculated between DITTO global feature extractor and FENDA global feature extractor

Let me know if you think this is wrong or even more unclear!

jewelltaylor (commented):
All in all, I think the changes look great! All the comments I made are minor or documentation related so I think this is pretty much good to merge in. I want to take another quick pass through once you have the chance to go through the comments to ensure I did not miss anything

fl4health/clients/adaptive_drift_constraint_client.py
emersodb: Added some language in the init docstring to better tie these together. It doesn't necessarily have to be the criterion loss. So I made the docstring for the class more general.

fl4health/clients/adaptive_drift_constraint_client.py
emersodb: My only hesitation with calling it criterion loss or loss without penalty is that it doesn't _necessarily_ have to be that. Do you think "loss_for_adaptation" is a bit better?

fl4health/clients/adaptive_drift_constraint_client.py
emersodb: Sure. That's good with me.

fl4health/clients/ditto_client.py
emersodb: Sure!

fl4health/clients/fenda_ditto_client.py
emersodb: Nope that looks great. Added to the documentation.

fl4health/clients/ditto_client.py
emersodb: Sure!

fl4health/clients/adaptive_drift_constraint_client.py
jewelltaylor: Yeah, i think that is a bit more clear!

jewelltaylor (approved):
LGTM!

PR #236, opened 2024-09-13
Add updates for smaller CRs by David

jewelltaylor:
# PR Type

# Short Description

Quick follow up of #234 addressing some of Davids smaller CRs.

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -1321,7 +1321,7 @@ class BasicClient(NumPyClient):
         # should be loaded
         for key in self.optimizers.keys():
             optimizer = self.optimizers[key]
-            optimizer_state = ckpt[key]
+            optimizer_state = ckpt["optimizers_state"][key]
             optimizer_state_dict = optimizer.state_dict()
             optimizer_state_dict["state"] = optimizer_state
             optimizer.load_state_dict(optimizer_state_dict)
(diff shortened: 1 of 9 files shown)


fl4health/clients/basic_client.py
```diff
@@ -1319,10 +1319,12 @@ def load_checkpoint(self) -> None:
         # Optimizer is updated in setup_client to reference model weights from server
         # Thus, only optimizer state (per parameter values such as momentum)
         # should be loaded
-        for opt, state in zip(self.optimizers.values(), ckpt["optimizers_state"].values()):
-            opt_state_dict = opt.state_dict()
-            opt_state_dict["state"] = state
-            opt.load_state_dict(opt_state_dict)
+        for key in self.optimizers.keys():
+            optimizer = self.optimizers[key]
```
emersodb, line 1323: super minor, but you can eliminate this line by doing `for key, optimizer in self.optimizers.items():`

fl4health/utils/nnunet_utils.py
```diff
@@ -423,6 +423,7 @@ def __init__(
         self._step_count: int
         super().__init__(optimizer, -1, False)
 
+    # mypy incorrectly infers get_lr returns a float
```
emersodb, line 426: Is this a bug that we can link too? If it's more subtle than that, maybe we can briefly describe why it happens?

emersodb (approved):
Looks good to me. Only had two small comments that would be good to address, but they aren't deal breaking 🙂

PR #237, opened 2024-09-15
Adding in Dirichlet allocation partitioner

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8689qhwfz)

This PR adds in label-based Dirichlet partitioning. In the Dirichlet sampler that we have, we take an existing dataset and subsample the datapoints according to a Dirichlet distribution with a given $\beta$. In this partitioning example, we take an existing dataset and **split** it into pieces by iterating through the labels and use a Dirichlet distribution to assign a portion of those labels to each client. These two approaches are slightly different and both have been used in literature. 

However, the later is used in two benchmarks ([1](https://arxiv.org/pdf/2206.03655), [2](https://www.computer.org/csdl/journal/oj/2024/01/10316561/1S2UbvQk5Tq)), which, for future experiments, we'd like to reproduce.

# Tests Added

Added a test to ensure that the partitioning is happening as expected and has the right relationship to the Dirichlet $\beta$ parameter.

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -201,7 +201,11 @@ class NnunetClient(BasicClient):
         # Set the number of processes for each dataloader.
         if self.n_dataload_proc is None:
             # Nnunet default is 12 or max cpu's. We decrease max by 1 just in case
-            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)
+            # NOTE: The type: ignore here is to skip issues where a local operating system is not compatible
+            # with sched_getaffinity (older versions of MacOS, for example). The code still won't run but mypy won't
+            # complain. Workarounds like using os.cpu_count(), while not exactly the same, are possible.
+            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)  # type: ignore
+
         os.environ["nnUNet_n_proc_DA"] = str(self.n_dataload_proc)
 
         # The batchgenerators package used under the hood by the dataloaders creates an
--- a/fl4health/utils/dataset.py
+++ b/fl4health/utils/dataset.py
@@ -1,14 +1,15 @@
+import copy
 from abc import ABC, abstractmethod
-from typing import Callable, Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, List, Optional, Tuple, TypeVar, Union, cast
 
 import torch
 from torch.utils.data import Dataset
 
 
 class BaseDataset(ABC, Dataset):
-    def __init__(self) -> None:
-        self.transform: Union[Callable, None]
-        self.target_transform: Union[Callable, None]
+    def __init__(self, transform: Optional[Callable], target_transform: Optional[Callable]) -> None:
+        self.transform = transform
+        self.target_transform = target_transform
 
     def update_transform(self, f: Callable) -> None:
         if self.transform:
@@ -41,11 +42,9 @@ class TensorDataset(BaseDataset):
         transform: Optional[Callable] = None,
         target_transform: Optional[Callable] = None,
     ) -> None:
-        super().__init__()
+        super().__init__(transform, target_transform)
         self.data = data
         self.targets = targets
-        self.transform = transform
-        self.target_transform = target_transform
 
     def __getitem__(self, index: int) -> Tuple[torch.Tensor, torch.Tensor]:
         assert self.targets is not None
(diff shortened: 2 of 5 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -201,7 +201,11 @@ def get_data_loaders(self, config: Config) -> Tuple[DataLoader, DataLoader]:
         # Set the number of processes for each dataloader.
         if self.n_dataload_proc is None:
             # Nnunet default is 12 or max cpu's. We decrease max by 1 just in case
-            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)
+            # NOTE: The type: ignore here is to skip issues where a local operating system is not compatible
+            # with sched_getaffinity (older versions of MacOS, for example). The code still won't run but mypy won't
+            # complain. Workarounds like using os.cpu_count(), while not exactly the same, are possible.
+            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)  # type: ignore
+
```
sanaAyrml, line 208: Good call, I was also keep getting error on my laptop for this.

fl4health/utils/dataset.py
```diff
@@ -103,3 +102,55 @@ def __getitem__(self, index: int) -> Tuple[Dict[str, torch.Tensor], torch.Tensor
...
+        return data, target
+
+    def __len__(self) -> int:
+        return len(self.data)
+
+
+D = TypeVar("D", bound=Union[TensorDataset, DictionaryDataset])
+
+
+def select_by_indices(dataset: D, selected_indices: torch.Tensor) -> D:
```
sanaAyrml, line 136: Is there a specific reason we are moving select_by_indices out of LabelBasedSampler?

fl4health/utils/dataset.py
```diff
@@ -103,3 +102,55 @@ def __getitem__(self, index: int) -> Tuple[Dict[str, torch.Tensor], torch.Tensor
...
+    if isinstance(dataset, TensorDataset):
+        modified_dataset = copy.deepcopy(dataset)
```
sanaAyrml, line 138: I think the reason to do deepcopy here because it would get a new memory than directly changing input dataset, so we can apply selection multiple times and wouldn't modify the original one. But should not we do the same for DictionaryDataset as well in line 146?

fl4health/clients/nnunet_client.py
emersodb: Yeah. CPU checking is a bit fluky across operating systems. We'll just punt this for now.

fl4health/utils/dataset.py
sanaAyrml: Oh I see you want to also use it in partitioner. 👍

emersodb: Good question. I extracted the function because it doesn't use any class specific variables and it's actually agnostic of the fact that you're using any kind of label based sampler. That is, this can be used by any type of sampler (say random sampler for a trivial example) as long as you generate indicates to select in the dataset.

fl4health/utils/dataset.py
emersodb: I'm not 100% sure. So correct me if I'm wrong here. However, in the original implementation of `select_by_indices` we were making modifications to the `dataset.targets` and `dataset.data` in-place, which resulted in changes to the original object. Here, we're creating a new DictionaryDataset with these values on line 154. So the original dataset should still be intact and unchanged.

fl4health/utils/partitioners.py
```diff
@@ -0,0 +1,99 @@
...
+import torch
+from flwr.common.logger import log
+
+from fl4health.utils.dataset import DictionaryDataset, TensorDataset, select_by_indices
+
+T = TypeVar("T")
+D = TypeVar("D", bound=Union[TensorDataset, DictionaryDataset])
+
+
+class LabelBasedDirichletAllocation:
```
sanaAyrml, line 15: In order to be consistent with DirichletLabelBasedSampler maybe we can rename this as DirichletLabelBasedAllocation?

emersodb: Will do!

fl4health/utils/dataset.py
sanaAyrml: Oh I see in the last line we do that. Yes yes then it should be fine.

sanaAyrml (approved):
Everything looks great, and the current Dirichlet allocation implementation aligns well with what I developed for my master's project. One suggestion would be to include an example that demonstrates how to use the partitioning module. While I was able to figure it out by examining the test functions, having a dedicated example would make it easier for users to understand without needing to refer to the tests.

emersodb:
> Everything looks great, and the current Dirichlet allocation implementation aligns well with what I developed for my master's project. One suggestion would be to include an example that demonstrates how to use the partitioning module. While I was able to figure it out by examining the test functions, having a dedicated example would make it easier for users to understand without needing to refer to the tests.

Yup, will do. I won't implement a full example in our examples folder, since this is a fairly small piece, but I will add a code snippet in the class documentation that gives a basic example 🙂

PR #238, opened 2024-09-17
Adding in synthetic dataset generators based on the original FedProx paper

emersodb:
# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8689qhw2n)

This small PR implements synthetic dataset generators based on the original FedProx paper. These generators are to be used in future experiments and replication studies. There are a few small modifications to how the generation takes place, but most of the process is inspired by the original implementation.

**Note**: These synthetic datasets end up with pretty skewed labels, which is a bit disappointing. See the tests for additional illustration. However, they will still be useful. It's just something to be aware of.

# Tests Added

Several tests are implemented to check correctness of both generators (IID, non-IID)

### DIFF
--- a/fl4health/utils/data_generation.py
+++ b/fl4health/utils/data_generation.py
@@ -92,6 +92,9 @@ class SyntheticFedProxDataset(ABC):
             List[TensorDataset]: Synthetic datasets for each client.
         """
         client_tensors = self.generate_client_tensors()
+        assert (
+            len(client_tensors) == self.num_clients
+        ), "The tensors returned by generate_client_tensors should have the same length as self.num_clients"
         client_datasets = [TensorDataset(X, Y) for X, Y in client_tensors]
         return client_datasets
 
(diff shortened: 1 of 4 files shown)


fl4health/utils/data_generation.py
```diff
@@ -0,0 +1,295 @@
...
+        return F.one_hot(samples, num_classes=self.output_dim).squeeze()
+
+    def generate(self) -> List[TensorDataset]:
+        """
+        Based on the class parameters, generate a list of synthetic TensorDatasets, one for each client.
+
+        Returns:
+            List[TensorDataset]: Synthetic datasets for each client.
+        """
+        client_tensors = self.generate_client_tensors()
```
fatemetkl, line 94: Super minor, but maybe it is worthwhile to assert that `generate_client_tensors` returns a list with length self.num_clients.

fl4health/utils/data_generation.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+        super().__init__(
+            num_clients=num_clients,
+            temperature=temperature,
+            input_dim=input_dim,
+            output_dim=output_dim,
+            samples_per_client=samples_per_client,
+        )
+
+        # As described in the original paper, the affine transformation is SHARED by all clients and the elements
```
fatemetkl, line 257: Just to highlight that `v` is also the same across clients, maybe, for clarity, we can put this here `self.v = torch.zeros(self.input_dim)`, and in the `get_input_output_tensors` function use it `loc = self.v`

fl4health/utils/data_generation.py
```diff
@@ -0,0 +1,295 @@
...
+    def generate(self) -> List[TensorDataset]:
+        """
+        Based on the class parameters, generate a list of synthetic TensorDatasets, one for each client.
+
+        Returns:
+            List[TensorDataset]: Synthetic datasets for each client.
+        """
+        client_tensors = self.generate_client_tensors()
+        client_datasets = [TensorDataset(X, Y) for X, Y in client_tensors]
+        return client_datasets
```
fatemetkl, line 96: A general question: how can each client get access to its own TensorDataset?
Do we need to implement or use another function to pass this to individual clients?

fatemetkl (approved):
LGTM! Just some minor comments/questions.

fl4health/utils/data_generation.py
emersodb: Will add!

fl4health/utils/data_generation.py
emersodb: I went a little bit further and just instantiated the multivariate normal in the init function to emphasis that it is also shared. Let me know what if that makes sense.

fl4health/utils/data_generation.py
emersodb: That's a good question. This wasn't implemented with an "integration" into the FL flow in mind. Similar to the Dirichlet allocation partitioner, this is being implemented to create/preprocess datasets. So in experiments, what I'll likely do is use this to generate a synthetic dataset for the number of clients I want to experiment with, save the datasets somewhere, and then for FL, I'll point each client to the right path to load the dataset up.

fl4health/utils/data_generation.py
fatemetkl: Looks great!

PR #239, opened 2024-09-17
Expanding ClientCheckpointModule to accept checkpointer sequences

emersodb:
Adding in the ability to pass multiple checkpointers to each of the pre and post aggregation components of the ClientCheckpointModule

# PR Type
Feature

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8689tgun7)

# Tests Added

Added a test that covers the case where pre-aggregation has a sequence of checkpointers and post aggregation only has one to test the mix. Also tested that the two different kinds of checkpointers exhibit the right behavior when used in the sequence.

### DIFF
--- a/tests/checkpointing/test_client_module.py
+++ b/tests/checkpointing/test_client_module.py
@@ -1,8 +1,9 @@
 from pathlib import Path
+from typing import List
 
 import torch
 
-from fl4health.checkpointing.checkpointer import BestLossTorchCheckpointer
+from fl4health.checkpointing.checkpointer import BestLossTorchCheckpointer, LatestTorchCheckpointer, TorchCheckpointer
 from fl4health.checkpointing.client_module import CheckpointMode, ClientCheckpointModule
 from fl4health.checkpointing.opacus_checkpointer import BestLossOpacusCheckpointer
 from fl4health.utils.privacy_utilities import convert_model_to_opacus_model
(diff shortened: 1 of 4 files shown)


tests/checkpointing/test_client_module.py
```diff
@@ -96,3 +96,74 @@ def test_client_checkpointer_module(tmp_path: Path) -> None:
...
+
+
+def test_path_duplication_check(tmp_path: Path) -> None:
+    checkpoint_dir = tmp_path.joinpath("resources")
+    checkpoint_dir.mkdir()
+    pre_aggregation_checkpointer: List[TorchCheckpointer] = [
+        BestLossTorchCheckpointer(str(checkpoint_dir), "pre_agg_best.pkl"),
+        LatestTorchCheckpointer(str(checkpoint_dir), "pre_agg_best.pkl"),
+    ]
+    post_aggregation_checkpointer = BestLossTorchCheckpointer(str(checkpoint_dir), "post_agg.pkl")
```
sanaAyrml, line 164: Can we add another small test here to check the error if we have same path for pre_aggregation_checkpointer and post_aggregation_checkpointer?

sanaAyrml (approved):
Thank you for the extension! It will greatly streamline our experimentation process, allowing us to avoid multiple runs for different checkpointing methods.

tests/checkpointing/test_client_module.py
emersodb: Yup. Will do now.

PR #242, opened 2024-09-19
Quick update for partitioner to also get a prior rather than sampling

sanaAyrml:
# PR Type
[Fix]

# Short Description

This update to the partitioner allows it to retrieve a prior distribution instead of generating a random one.

# Tests Added

Updated tests to support extra output.

### DIFF
--- a/fl4health/utils/partitioners.py
+++ b/fl4health/utils/partitioners.py
@@ -1,5 +1,5 @@
 import math
-from logging import INFO
+from logging import INFO, WARN
 from typing import Dict, List, Optional, Tuple, TypeVar, Union
 
 import numpy as np
@@ -56,20 +56,31 @@ class DirichletLabelBasedAllocation:
                 Otherwise the partitioner "gives up". Defaults to None.
             beta (Optional[float]): This controls the heterogeneity of the partition allocations. The smaller the beta,
               the more skewed the label assignments will be to different clients. It is mutually exclusive with given
-              prior distribution..
+              prior distribution.
             prior_distribution (Optional[Dict[T, np.ndarray]], optional): This is an optional input if you want to
               provide a prior distribution for the Dirichlet distribution. This is useful if you want to make sure that
               the partitioning of test data is similar to the partitioning of the training data. Defaults to None. It
               is mutually exclusive with the beta parameter and min_label_examples.
         """
-        assert beta is not None or prior_distribution is not None, "Either beta or prior must be provided"
-        assert beta is None or prior_distribution is None, "Either beta or prior must be provided, not both"
+        assert (beta is not None) ^ (
+            prior_distribution is not None
+        ), "Either beta or a prior distribution must be provided, but not both."
         self.number_of_partitions = number_of_partitions
         self.unique_labels = unique_labels
         self.n_unique_labels = len(unique_labels)
-        self.prior_distribution = prior_distribution
         self.beta = beta
         self.min_label_examples = min_label_examples if min_label_examples else 0
+        self.prior_distribution = prior_distribution
+        if self.prior_distribution is not None:
+            assert (
+                len(self.prior_distribution) == self.n_unique_labels
+            ), "The length of the prior must match the number of labels"
+            if self.min_label_examples > 0:
+                log(
+                    WARN,
+                    """A prior distribution has been provided for the partitioner,
+                    so min_label_examples will be ignored""",
+                )
 
     def partition_label_indices(
         self, label: T, label_indices: torch.Tensor
@@ -89,16 +100,13 @@ class DirichletLabelBasedAllocation:
             np.ndarray: The Dirichlet distribution used to partition the data points.
         """
         if self.prior_distribution is not None:
-            assert (
-                len(self.prior_distribution) == self.n_unique_labels
-            ), "The length of the prior must match the number of labels"
             partition_allocations = self.prior_distribution[label]
             log(
                 INFO,
-                f"Using a prior distribution of {self.prior_distribution[label]} to allocate the label ({str(label)})",
+                f"The allocation distribution for label ({str(label)}) is {partition_allocations} "
+                f"using the provided prior distribution",
             )
-        if self.beta is not None:
-            total_data_points_with_label = len(label_indices)
+        elif self.beta is not None:
             # These are the percentages of the label indices to be distributed for each partition
             partition_allocations = np.random.dirichlet(np.repeat(self.beta, self.number_of_partitions))
             log(
@@ -108,7 +116,9 @@ class DirichletLabelBasedAllocation:
                     f"using a beta of {self.beta}",
                 ),
             )
-
+        else:
+            raise ValueError("Either beta or a prior distribution must be provided.")
+        total_data_points_with_label = len(label_indices)
         num_samples_per_partition = [
             math.floor(probability * total_data_points_with_label) for probability in partition_allocations
         ]
@@ -145,7 +155,7 @@ class DirichletLabelBasedAllocation:
 
         Returns:
             List[D]: The partitioned datasets, length should correspond to self.number_of_partitions
-
+            Dict[T, np.ndarray]: The Dirichlet distribution used to partition the data points for each label.
         """
         targets = original_dataset.targets
         assert targets is not None, "A label-based partitioner requires targets but this dataset has no targets"
@@ -161,6 +171,7 @@ class DirichletLabelBasedAllocation:
                     self.partition_label_indices(label, label_indices)
                 )
                 # If the minimum number of labels is satisfied or if there is a prior distribution, we accept the
+                # partition. Otherwise, we retry.
                 if self.prior_distribution is not None or min_selected_labels >= self.min_label_examples:
                     partitioned_probabilities[label] = partitioned_probability
                     for i, indices_for_label_partition in enumerate(partitioned_indices_for_label):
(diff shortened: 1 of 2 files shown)


fl4health/utils/partitioners.py
```diff
@@ -43,22 +48,32 @@ def __init__(
...
             min_label_examples (Optional[int], optional): This is an optional input if you want to ensure a minimum
-                number of labels is present on each partition.
+                number of labels is present on each partition. If prior distribution is provided, this is ignored.
                 NOTE: This does not guarantee feasibility. That is, if you have a very small beta and request a large
                 minimum number here, you are unlikely to satisfy this request. In partitioning, if the minimum isn't
                 satisfied, we resample from the Dirichlet distribution. This is repeated some limited number of times.
                 Otherwise the partitioner "gives up". Defaults to None.
+            beta (Optional[float]): This controls the heterogeneity of the partition allocations. The smaller the beta,
+              the more skewed the label assignments will be to different clients. It is mutually exclusive with given
+              prior distribution..
```
emersodb, line 59: Two periods here 🙂

fl4health/utils/partitioners.py
```diff
@@ -43,22 +48,32 @@ def __init__(
...
+            prior_distribution (Optional[Dict[T, np.ndarray]], optional): This is an optional input if you want to
+              provide a prior distribution for the Dirichlet distribution. This is useful if you want to make sure that
+              the partitioning of test data is similar to the partitioning of the training data. Defaults to None. It
+              is mutually exclusive with the beta parameter and min_label_examples.
         """
+        assert beta is not None or prior_distribution is not None, "Either beta or prior must be provided"
```
emersodb, line 65: These two asserts can be combined to be a bit more concise as
```
assert (beta is not None) ^ (prior_distribution is not None), "Either beta or a prior distribution must be provided, but not both."
```

fl4health/utils/partitioners.py
```diff
@@ -43,22 +48,32 @@ def __init__(
...
+        assert beta is None or prior_distribution is None, "Either beta or prior must be provided, not both"
         self.number_of_partitions = number_of_partitions
         self.unique_labels = unique_labels
         self.n_unique_labels = len(unique_labels)
+        self.prior_distribution = prior_distribution
         self.beta = beta
         self.min_label_examples = min_label_examples if min_label_examples else 0
```
emersodb, line 72: We might want to add a warning for the user here that a non-zero `min_label_examples` will be ignored if there is a prior distribution. Something like
```
from logging import WARN
if self.min_label_examples > 0 and self.prior_distribution is not None:
    log(WARN, "A prior distribution has been provided for the partitioner, so min_label_examples will be ignored")
```

fl4health/utils/partitioners.py
```diff
@@ -99,9 +126,9 @@ def partition_label_indices(self, label: T, label_indices: torch.Tensor) -> Tupl
         partitioned_indices = list(torch.split(shuffled_indices, num_samples_per_partition))
 
         # Dropping the last partition as they are "excess" indices
-        return partitioned_indices[:-1], min_samples
+        return partitioned_indices[:-1], min_samples, partition_allocations
 
-    def partition_dataset(self, original_dataset: D, max_retries: int = 5) -> List[D]:
+    def partition_dataset(self, original_dataset: D, max_retries: int = 5) -> Tuple[List[D], Dict[T, np.ndarray]]:
```
emersodb, line 131: The documentation on the return here still needs to be updated to reflect the `Dict[T, np.ndarray]` that has been added I think.

fl4health/utils/partitioners.py
```diff
@@ -70,17 +85,29 @@ def partition_label_indices(self, label: T, label_indices: torch.Tensor) -> Tupl
...
-        # These are the percentages of the label indices to be distributed for each partition
-        partition_allocations = np.random.dirichlet(np.repeat(self.beta, self.number_of_partitions))
-        log(
-            INFO,
-            (
-                f"The allocation distribution for label ({str(label)}) is {partition_allocations} "
-                f"using a beta of {self.beta}",
-            ),
-        )
+        if self.prior_distribution is not None:
```
emersodb, line 91: Rather than doing this check each time we partition, we should probably move this to the init function?

fl4health/utils/partitioners.py
```diff
@@ -70,17 +85,29 @@ def partition_label_indices(self, label: T, label_indices: torch.Tensor) -> Tupl
...
+            assert (
+                len(self.prior_distribution) == self.n_unique_labels
+            ), "The length of the prior must match the number of labels"
+            partition_allocations = self.prior_distribution[label]
+            log(
+                INFO,
+                f"Using a prior distribution of {self.prior_distribution[label]} to allocate the label ({str(label)})",
```
emersodb, line 98: Super minor, but we already have `partition_allocations`, so we might as well use it for the logging here?

fl4health/utils/partitioners.py
```diff
@@ -70,17 +85,29 @@ def partition_label_indices(self, label: T, label_indices: torch.Tensor) -> Tupl
...
+            )
+        if self.beta is not None:
```
emersodb, line 100: Very minor, but just to be "on the safe side," in case we ever change something out the mutual exclusivity of beta and the prior in the init, perhaps we can do `elif` here?

fl4health/utils/partitioners.py
```diff
@@ -70,17 +85,29 @@ def partition_label_indices(self, label: T, label_indices: torch.Tensor) -> Tupl
...
+            total_data_points_with_label = len(label_indices)
```
emersodb, line 101: I don't think you want this guy to be here? That is, it's going to cause a problem if you have a prior I think?

fl4health/utils/partitioners.py
```diff
@@ -118,19 +145,24 @@ def partition_dataset(self, original_dataset: D, max_retries: int = 5) -> List[D
...
             List[D]: The partitioned datasets, length should correspond to self.number_of_partitions
+
         """
         targets = original_dataset.targets
         assert targets is not None, "A label-based partitioner requires targets but this dataset has no targets"
         partitioned_indices = [torch.Tensor([]).int() for _ in range(self.number_of_partitions)]
 
         partition_attempts = 0
-
+        partitioned_probabilities: Dict = {}
```
emersodb, line 155: You can describe this more specifically for mypy as `Dict[T, np.ndarray]` I think?

tests/utils/partitioners_test.py
emersodb, line 1: We probably want to add a test that considers providing a prior to the partitioner here I think?

emersodb (commented):
Overall, looks pretty nice. Just some minor comments and one bug fix I think we need to make.

fl4health/utils/partitioners.py
sanaAyrml: I have added a comment as
` np.ndarray: The Dirichlet distribution used to partition the data points.`
to reflect the changes. Do I need to add anything extra?

fl4health/utils/partitioners.py
sanaAyrml: Correct, I also updated wording to match the below logging.

fl4health/utils/partitioners.py
sanaAyrml: I am confused which one do you mean exactly?

emersodb: The scope of `total_data_points_with_label` is tied to this if statement but you use it in the code outside of this if to establish `num_samples_per_partition`. If you have a prior defined, then I'm pretty sure `total_data_points_with_label` will be None and when you try to use it later it will throw.

fl4health/utils/partitioners.py
emersodb: I think that's in a different functions documentation (if I'm reading everything correctly). Right now, the documentation for this function has the following:
```
        Returns:
            List[D]: The partitioned datasets, length should correspond to self.number_of_partitions

        """
```
which doesn't reflect the return type that we're using for annotation above.

fl4health/utils/partitioners.py
sanaAyrml: So whenever I do that I get following weird error from mypy ```
Invalid index type "T" for
"dict[T, ndarray[Any, Any]]"; expected type "T"
```
I assume it should be because of generic T type that causes this error.

File: tests/utils/partitioners_test.py
Comment (sanaAyrml) [reply tests/utils/partitioners_test.py:1]: Added a test 👍

File: fl4health/utils/partitioners.py [hunk 10]
```diff
@@ -90,15 +101,21 @@ def partition_label_indices(
...
+                    WARN,
+                    f"The provided prior distribution for label ({str(label)}) does not sum to 1. "
+                    "It will be normalized to sum to 1.",
+                )
+            partition_allocations = self.prior_distribution[label] / sum(self.prior_distribution[label])
             log(
                 INFO,
-                f"Using a prior distribution of {self.prior_distribution[label]} to allocate the label ({str(label)})",
+                f"The allocation distribution for label ({str(label)}) is {partition_allocations} "
+                f"using the provided prior distribution",
```
emersodb, line 116: I'm actually not sure why flake doesn't complain about this, but most of the time it gets fussy when using an f-string without an actual variable to dereference. Maybe drop the f here to be safe?

emersodb (approved):
Three very minor comments. Otherwise, good to go.

PR #247, opened 2024-10-03
Cleanup picai folder

jewelltaylor:
# PR Type
[**Documentation** | Other ]

# Short Description

Clickup Ticket(s): Link(s) [Update PICAI Research Folder](https://app.clickup.com/t/868a2ra5u)

The PICAI research folder needs to be updated. This includes updating the documentation to be more general and up to date. Fixing issues with images that aren't rendering. And a few other details.


research/picai/preprocessing/prepare_data.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 import SimpleITK as sitk
 
-from research.picai.preprocessing import (
+from research.picai.preprocessing.preprocessing import (
```
emersodb, line 10: preprocessing.preprocessing is a little awkward, but if `preprocessing.py` is the most logical name for the file, we can keep it 🙂

scarere: Maybe you could just import everything from the preprocessing file into the preprocessing subpackage in the __init__.py file?

scarere: Or alternatively `from .preprocessing impot ...`

jewelltaylor: Both seem like solid suggestions, I might go with a rename of preprocessing dir to data_utils in case we need to add data related files that are not necessarily preprocessing (ie augmentation) in the future

PR #248, opened 2024-10-07
Fix launch scripts

jewelltaylor:
# PR Type
[ **Fix** ]

# Short Description

Small PR to update the server and client python scripts to include arguments required for checkpointing. Also fix an issue where the client name was not being passed properly in slurm script.

# Tests Added


emersodb (approved):
Looks good to me...bash is so dumb 😂

PR #252, opened 2024-10-10
Changes to Support Expanded Experimentation with FedDG-GA

emersodb:
# PR Type
Feature/Experimentation.

# Short Description

The changes in this PR are targeted at enabling expanded experimentation with the FedDG-GA strategy for a wider set of FL approaches. Also included is a name change, `FedDgGaStrategy` -> `FedDgGa` to fit other strategy formats a bit better and renaming of the associated file from `feddg_ga_strategy.py` to `feddg_ga.py` as it is already under the strategies folder.

For FENDA+Ditto, there is also a bug fix moving from `SequentiallySplitExchangeBaseModel` to `SequentiallySplitModel`. In this setting we want to exchange the whole Ditto model, not just the feature extractor component. It wasn't causing a real bug, as a FullParameterExchanger was being used anyway, but the typing was dissonant.

Finally, I moved some "client agnostic" functionality out of basic client and into a utils file to help trim a few functions from the `BasicClient` class.

# Tests Added

Added a test for the new Fed DG-GA strategy that is compatible with adaptive constraint server-client pairs


fl4health/clients/basic_client.py
```diff
@@ -247,8 +240,13 @@ def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, N
         except ValueError:
             evaluate_after_fit = False
 
+        try:
```
emersodb, line 243: This will control whether the contents of the loss dictionary should be packed into the metrics for communication with the server or not. Packing the losses makes the full set of additional losses available to the server, which is essentially for FedDG-GA to work for an expanded set of FL techniques. However, it also means that the server would be aware of these which is, perhaps, advantageous in other settings as well.

fl4health/clients/basic_client.py
```diff
@@ -326,21 +326,6 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
             metrics,
         )
 
-    def evaluate_after_fit(self) -> Tuple[float, Dict[str, Scalar]]:
```
emersodb, line 329: This is being torched in favor of just using validate with `pack_losses_with_val_metrics = True`. We were packing the loss here specifically to facilitate FedDG-GA.

fl4health/clients/basic_client.py
```diff
@@ -536,69 +527,6 @@ def get_client_specific_reports(self) -> Dict[str, Any]:
...
-            return {key: value.to(self.device) for key, value in data.items()}
-        else:
-            raise TypeError(
-                "data must be of type torch.Tensor or Dict[str, torch.Tensor]. \
-                    If definition of TorchInputType or TorchTargetType has \
-                    changed this method might need to be updated or split into \
-                    two"
-            )
-
-    def is_empty_batch(self, input: Union[torch.Tensor, Dict[str, torch.Tensor]]) -> bool:
```
emersodb, line 571: This is a pretty generic function. Moved it to `utils/client.py`. Let me know if you disagree with the change.

fl4health/clients/basic_client.py
```diff
@@ -1034,24 +977,6 @@ def set_optimizer(self, config: Config) -> None:
         assert not isinstance(optimizer, dict)
         self.optimizers = {"global": optimizer}
 
-    def clone_and_freeze_model(self, model: nn.Module) -> nn.Module:
```
emersodb, line 1037: This is a pretty generic function. Moved it to utils/client.py. Let me know if you disagree with the change.

fl4health/clients/basic_client.py
```diff
@@ -1241,35 +1166,6 @@ def update_before_epoch(self, epoch: int) -> None:
         """
         pass
 
-    def maybe_progress_bar(self, iterable: Iterable) -> Iterable:
```
emersodb, line 1244: This is a pretty generic function. Moved it to utils/client.py. Let me know if you disagree with the change.

fl4health/clients/fenda_ditto_client.py
```diff
@@ -9,7 +9,7 @@
 from fl4health.checkpointing.client_module import ClientCheckpointModule
 from fl4health.clients.ditto_client import DittoClient
 from fl4health.model_bases.fenda_base import FendaModel
-from fl4health.model_bases.sequential_split_models import SequentiallySplitExchangeBaseModel
+from fl4health.model_bases.sequential_split_models import SequentiallySplitModel
```
emersodb, line 12: While it didn't actually change the flow of the example, since it uses a `FullParameterExchanger` this type is misleading, given that it's mean to facilitate only exchanging the feature extraction module.

fl4health/strategies/feddg_ga.py
```diff
@@ -23,7 +30,7 @@ class FairnessMetricType(Enum):
     """Defines the basic types for fairness metrics, their default names and their default signals"""
 
     ACCURACY = "val - prediction - accuracy"
-    LOSS = "val - loss"
```
emersodb, line 26: Changing to checkpoint, as this is the name of the vanilla loss for basic clients during validation.

fl4health/strategies/feddg_ga.py
```diff
@@ -220,18 +231,41 @@ def configure_fit(
...
+    ) -> List[Tuple[ClientProxy, EvaluateIns]]:
+        assert isinstance(
+            client_manager, FixedSamplingClientManager
+        ), f"Client manager is not of type FixedSamplingClientManager: {type(client_manager)}"
+
+        client_evaluate_ins = super().configure_evaluate(server_round, parameters, client_manager)
+
+        assert self.on_evaluate_config_fn is not None, "on_fit_config_fn must be specified"
+        config = self.on_evaluate_config_fn(server_round)
+        assert "pack_losses_with_val_metrics" in config, "pack_losses_with_val_metrics must be present in config"
```
emersodb, line 264: During validation, we need the server to have access to the losses. So `pack_losses_with_val_metrics` must be present and true for FedDG-GA to work properly.

fl4health/strategies/feddg_ga.py
```diff
@@ -252,9 +286,19 @@ def aggregate_fit(
...
+        if not self.accept_failures and failures:
+            return None, {}
+
+        # Aggregate custom metrics if aggregation fn was provided
+        metrics_aggregated = {}
+        if self.fit_metrics_aggregation_fn:
+            fit_metrics = [(res.num_examples, res.metrics) for _, res in results]
+            metrics_aggregated = self.fit_metrics_aggregation_fn(fit_metrics)
+        elif server_round == 1:  # Only log this warning once
+            log(WARNING, "No fit_metrics_aggregation_fn provided")
```
emersodb, line 301: Rather than relying on the original aggregation approach and discarding the weights aggregation, which is a bit of a wasted set of calculations, just do the metrics aggregation here, since that is all we want anyway.

fl4health/strategies/feddg_ga.py
```diff
@@ -290,10 +334,9 @@ def aggregate_evaluate(
         self.evaluation_metrics = {}
         for client_proxy, eval_res in results:
             cid = client_proxy.cid
+            # make sure that the metrics has the desired loss key
+            assert FairnessMetricType.LOSS.value in eval_res.metrics
             self.evaluation_metrics[cid] = eval_res.metrics
-            # adding the loss to the metrics
-            val_loss_key = FairnessMetricType.LOSS.value
-            self.evaluation_metrics[cid][val_loss_key] = eval_res.loss
```
emersodb, line 296: No need to force the loss into the dictionary here. It should be packed with the metrics. If it's not, something else has gone wrong.

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
+from logging import INFO, WARNING
```
emersodb, line 1: This class is needed in order to facilitate FedDG-GA for clients that also require adaptive constraint considerations. These are clients such as Ditto, MR-MTL, and FedProx. This strategy needs to coordinate the exchange of the loss weights in addition to the Generalization Adjustments for aggregation

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+        Returns:
+            List[Tuple[int, float]]: A list of the training losses produced by client training
+        """
+        train_losses_and_counts: List[Tuple[int, float]] = []
+        for _, fit_res in results:
+            sample_count = fit_res.num_examples
+            updated_weights, train_loss = self.parameter_packer.unpack_parameters(
+                parameters_to_ndarrays(fit_res.parameters)
+            )
+            # Modify the parameters in-place to just be the model weights.
```
emersodb, line 231: Since the training losses are packed with the weights, we need to extract the losses and jam the weights back into the FitRes parameters object so that FedDG-GA can happen unabated.

tests/smoke_tests/load_from_checkpoint_example/README.md
emersodb, line 1: Deleted this as it seemed to be a carbon copy of the BasicExample readme? Perhaps that's not true. Just let me know

fl4health/clients/basic_client.py
```diff
@@ -248,8 +241,13 @@ def process_config(self, config: Config) -> Tuple[Union[int, None], Union[int, N
         except ValueError:
             evaluate_after_fit = False
 
+        try:
+            pack_losses_with_val_metrics = narrow_dict_type(config, "pack_losses_with_val_metrics", bool)
```
jewelltaylor, line 245: Is it worth having an INFO log indicating that val metrics and backed with losses?

fl4health/clients/basic_client.py
```diff
@@ -536,69 +527,6 @@ def get_client_specific_reports(self) -> Dict[str, Any]:
         """
         return {}
 
-    def _move_data_to_device(
```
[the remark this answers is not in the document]
jewelltaylor: Nice!

fl4health/strategies/feddg_ga.py
```diff
@@ -220,18 +231,41 @@ def configure_fit(
...
         if self.num_rounds is None:
-            assert self.on_fit_config_fn is not None, "on_fit_config_fn must be specified"
-            config = self.on_fit_config_fn(server_round)
-            assert "evaluate_after_fit" in config, "evaluate_after_fit must be present in config and set to True"
-            assert config["evaluate_after_fit"] is True, "evaluate_after_fit must be set to True"
-            assert "n_server_rounds" in config, "n_server_rounds must be specified"
-            assert isinstance(config["n_server_rounds"], int), "n_server_rounds is not an integer"
             self.num_rounds = config["n_server_rounds"]
+        else:
+            assert config["n_server_rounds"] == self.num_rounds, "n_server_rounds has changed from the original value"
```
jewelltaylor, line 249: minor: might be nice to include the values of `config["n_server_rounds"]` and `n_server_rounds` in the assertion

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+                Function used to configure validation. Defaults to None.
+            accept_failures : bool, optional
+                Whether or not accept rounds containing failures. Defaults to True.
+            initial_parameters : Parameters, optional
+                Initial global model parameters.
+            fit_metrics_aggregation_fn : Optional[MetricsAggregationFn]
+                Metrics aggregation function, optional.
+            evaluate_metrics_aggregation_fn : Optional[MetricsAggregationFn]
+                Metrics aggregation function, optional.
+            adapt_loss_weight (bool, optional): Determines whether the value of mu is adaptively modified by
```
jewelltaylor, line 84: nit: Should we clarify what mu is or just refer to it as loss weight? mu isn't used isnt used in the FedGdGa paper so it may be a bit confusing

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+                FairnessMetricType or set to use a custom metric. Optional, default is
+                FairnessMetric(FairnessMetricType.LOSS).
+            weight_step_size : float
+                The step size to determine the magnitude of change for the adjustment weight. It has to be
+                0 < weight_step_size < 1. Optional, default is 0.2.
+        """
+        if fraction_fit != 1.0 or fraction_evaluate != 1.0:
+            log(
+                WARNING,
+                "fraction_fit or fraction_evaluate are not 1.0. The behaviour of FedDG-GA is unknown in those cases.",
```
jewelltaylor, line 106: Should we just throw an AssertionError if the behaviour is undefined? Or is there scenarios where we want to this anyway

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+                Optional[Tuple[float, Dict[str, Scalar]]],
+            ]
+        ] = None,
+        on_fit_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        on_evaluate_config_fn: Optional[Callable[[int], Dict[str, Scalar]]] = None,
+        accept_failures: bool = True,
+        initial_parameters: Parameters,
+        fit_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        evaluate_metrics_aggregation_fn: Optional[MetricsAggregationFn] = None,
+        initial_loss_weight: float = 1.0,
```
jewelltaylor, line 36: I think this is missing from the docstring in the constructor

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
```diff
@@ -0,0 +1,268 @@
...
+            evaluate_metrics_aggregation_fn=evaluate_metrics_aggregation_fn,
+        )
+
+        if fairness_metric is None:
+            self.fairness_metric = FairnessMetric(FairnessMetricType.LOSS)
+        else:
+            self.fairness_metric = fairness_metric
+
+        self.weight_step_size = weight_step_size
+        assert 0 < self.weight_step_size < 1, f"weight_step_size has to be between 0 and 1 ({self.weight_step_size})"
```
jewelltaylor, line 143: nit: should we also do assertions on self.initial_loss_weight, does it also have to be between 0 and 1?

fl4health/utils/client.py
```diff
@@ -0,0 +1,130 @@
...
+    # Prefixing the loss value keys with the mode from which they are generated
+    if logging_mode is LoggingMode.VALIDATION:
+        metrics.update({f"{MetricPrefix.VAL_PREFIX.value} {key}": loss_val for key, loss_val in loss_dict.items()})
+    else:
+        metrics.update({f"{MetricPrefix.TEST_PREFIX.value} {key}": loss_val for key, loss_val in loss_dict.items()})
+
+
+def move_data_to_device(data: T, device: torch.device) -> T:
+    """
+    _summary_
```
jewelltaylor, line 29: nit: I think you forgot to finish this

fl4health/utils/client.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        raise TypeError(
+            "data must be of type torch.Tensor or Dict[str, torch.Tensor]. \
+                If definition of TorchInputType or TorchTargetType has \
+                changed this method might need to be updated or split into \
+                two"
+        )
+
+
+def is_empty_batch(input: TorchInputType) -> bool:
```
jewelltaylor, line 57: nit: We are checking if the batch is empty but we are also checking if the input is valid (ie for dictionaries, verifying all tensors have same length. Is it worth a slight renaming of the function?

fl4health/utils/client.py
```diff
@@ -0,0 +1,130 @@
...
+        Iterable: _description_
+    """
+    if not display_progress_bar:
+        return iterable
+    else:
+        # Create a clean looking tqdm instance that matches the flwr logging
+        kwargs: Any = {
+            "leave": True,
+            "ascii": " >=",
+            # "desc": f"{LOG_COLORS['INFO']}INFO{LOG_COLORS['RESET']} ",
```
jewelltaylor, line 125: Is this supposed to be here? If so, maybe some context on why its commented out but we are still keeping it

jewelltaylor (commented):
Everything looks pretty much good to go for me! I will wait till you have a chance to take a look at the minor comments I made and take one last pass tomorrow

fl4health/clients/basic_client.py
emersodb: Yup. I think so. Will add it.

fl4health/strategies/feddg_ga.py
emersodb: Yup, makes sense!

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
emersodb: Yup, good call.

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
emersodb: Yes, will do. It shouldn't really be called mu.

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
emersodb: The method can technically work in these cases. However, the behavior isn't considered in the original paper and there are a few edge case type things that aren't well specified. I think your suggestion is a good one. Let's throw an error so the user doesn't do anything they really don't want to do.

fl4health/strategies/feddg_ga_with_adaptive_constraint.py
emersodb: So this is actually un-related to the penalty weight parameter, which can be any value we want it to be technically. This is for the way in which the generalization adjustment weights are modified, which is confusing. I'll rename this value to specify that.

fl4health/utils/client.py
emersodb: Indeed I did. Fixed now.

fl4health/utils/client.py
emersodb: Yeah I think that makes sense. Will rename.

fl4health/utils/client.py
emersodb: hmmm...good catch. I'm not sure why this was commented out. It's also commented out in the original function being migrated here. I'll drop it for now. We have it in the git history if we need to retrieve it.

jewelltaylor (approved):
Changes look good to me!

PR #253, opened 2024-10-10
Adding fit end and evaluate end times in the client metrics

lotif:
# PR Type
Fix

# Short Description

Clickup Ticket(s): NA

Adding fit end and evaluate end times in the client metrics so FLorist can track it properly.

# Tests Added

Fixed existing tests accordingly.


emersodb (approved):
Good to go.

PR #254, opened 2024-10-10
Reporting restructure

scarere:
# PR Type
**Feature**

# Short Description

Clickup Ticket(s): [Add Base Reporter Class](https://app.clickup.com/t/868a2ww5r)

This PR restructures and unifies reporting. What was previously the metrics reporter is now JsonReporter. Both the JsonReporter and WandBReporter inherit from the same base class.

Servers and clients now accept any number of reporters and send data to them frequently (During setup, at the end of each round, each epoch and each batch). It is up to the reporter classes internal logic to determine how frequently to report data.

There is no longer any default reporters. Users will not have to pass None to the wandb reporter argument of clients and servers in order to turn wandb reporting off. Additionally if users want to generate the metrics json file frequently used for pytests, they must explicitly pass a JsonReporter to their server or client instance. Creating a JsonReporter instance with all its default arguments exactly mimics the behaviour of what was previously the metrics reporter (Except some of the keys have been changed). There is no longer any need to call dump to write the metrics, this has been moved into the server/client shutdown.

I also upgraded the wandb dependency.

I think there is some work to be done in terms of deciding how to structure what metrics and what to use for the key names. It be great if there was a bit of a more predictable pattern.

# Tests Added

Had to modify a few tests to reflect the modified structure of the output metrics json. Also found two issues with the tests.

The MKMMD loss tests only work if your machine does not have a GPU. My temporary fix for this was to set devide to CPU. They still run in a few seconds. If GPU acceleration is needed then someone needs to go through and debug them. Created a ticket [here](https://app.clickup.com/t/868a6uduu)

Tests that made use of both freeze_time and patch could not be run individually in some instances. It seems like somehow their imports were reliant on other tests in the test suite. This makes debugging tests a bit annoying because you have to run the entire test suite just to check whether or not you fixed one specific test that was failing. I created a ticket [here](https://app.clickup.com/t/868a6raj3)


fl4health/clients/ditto_client.py
```diff
@@ -112,7 +119,10 @@ def get_parameters(self, config: Config) -> NDArrays:
             NDArrays: GLOBAL model weights to be sent to the server for aggregation
         """
         if not self.initialized:
-            log(INFO, "Setting up client and providing full model parameters to the server for initialization")
```
jewelltaylor, line 115: Not a big deal at all but it looks like a lot of formatting related changes crept into the PR, may want to check the settings of your formatter

fl4health/clients/evaluate_client.py
```diff
@@ -88,7 +88,8 @@ def setup_client(self, config: Config) -> None:
         self.criterion = self.get_criterion(config)
         self.parameter_exchanger = self.get_parameter_exchanger(config)
 
-        self.metrics_reporter.add_to_metrics({"type": "client", "initialized": datetime.datetime.now()})
+        for r in self.reporters:
+            r.report({"host_type": "client", "initialized": str(datetime.datetime.now())})
```
jewelltaylor, line 92: nit: Is it worth having some sort of ReporterManager that takes in reporting metrics and routes it to each one of the reporters so you can minimize the amount of for loops you have in the code (ie one for each time you have to report or initialize anything)

fl4health/reporting/base_reporter.py
```diff
@@ -0,0 +1,51 @@
+"""Base Class for Reporters.
+
+Super simple for now but keeping it in a seperate file in case we ass more base methods.
```
jewelltaylor, line 3: nit: ass -> add

fl4health/reporting/base_reporter.py
```diff
@@ -0,0 +1,51 @@
...
+    def report(
+        self,
+        data: dict,
+        round: int | None = None,
+        epoch: int | None = None,
+        batch: int | None = None,
+    ) -> None:
+        """A method called by clients or servers to send data to the reporter.
+
+        The report method is called reduncantly by the clients every step, every epoch,
```
jewelltaylor, line 19: nit: reduncantly -> redundantly

jewelltaylor: The report method is called by the client/server at frequent intervals (ie step, epoch, round) and sometimes outside of a FL round (for high level summary data). It is the responsibility of the Reporter to determine when and what to report.

jewelltaylor: Just to not only be specific to client

fl4health/reporting/base_reporter.py
```diff
@@ -0,0 +1,51 @@
...
+        It is up to the reporter to determine when and what to report.
+
+        Args:
+            data (dict): The data to maybe report from the server or client.
+            round (int | None, optional): The current FL round. If None, this indicates
+                that the method was called outside of a round (eg. for summary
+                information). Defaults to None.
+            epoch (int | None, optional): The current epoch. If None then this method
+                was not called within the scope of an epoch. Defaults to None.
+            batch (int | None, optional): The current batch. If None then this method
```
jewelltaylor, line 30: Am I missing where we would report metrics on the step level? I assume batch != local steps based on description Or is there a reason we aren't tracking at the step level that I am missing

fl4health/reporting/base_reporter.py
```diff
@@ -0,0 +1,51 @@
...
+                was called outside the scope of a training or evaluation step (eg. at
+                the end of an epoch or round) Defaults to None.
+        """
+        raise NotImplementedError
+
+    def initialize(self, **kwargs: Any) -> None:
```
jewelltaylor, line 36: Should these last two methods raise NotImplementedError or are we assuming by default they do nothing unless overridden?

fl4health/reporting/json_reporter.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        self.run_id = run_id
+
+        self.output_folder = Path(output_folder)
+        self.metrics: dict[str, Any] = {}
+
+        self.output_folder.mkdir(exist_ok=True)
+        assert self.output_folder.is_dir(), f"Output folder '{self.output_folder}' is not a valid directory."
+
+    def initialize(self, **kwargs: Any) -> None:
```
jewelltaylor, line 49: We are at least assuming kwargs is a dict with string keys the way we are using it. If we imagine there is any use case where this would not be a dict with string keys, we can keep it as any otherwise maybe we do dict[str, Any]

fl4health/server/evaluate_server.py
```diff
@@ -92,33 +92,39 @@ def fit(self, num_rounds: int, timeout: Optional[float]) -> Tuple[History, float
...
+            r.report(
+                {
+                    "fit_elapsed_time": str(start_time - end_time),
+                    "fit_start": str(start_time),
+                    "fit_end": str(end_time),
+                    "num_rounds": num_rounds,
+                    "host_type": "server",
+                }
+            )
+            print("REPORTED")
```
jewelltaylor, line 115: nit: Think you forgot to remove this print statement?

tests/server/test_base_server.py
```diff
@@ -124,18 +134,22 @@ def test_metrics_reporter_fit_round(mock_fit_round: Mock) -> None:
...
-                "fit_start": datetime.datetime(2012, 12, 12, 12, 12, 12),
-                "metrics_aggregated": test_metrics_aggregated,
-                "fit_end": datetime.datetime(2012, 12, 12, 12, 12, 12),
+                "fit_round_start": str(datetime.datetime(2012, 12, 12, 12, 12, 12)),
+                "fit_metrics": test_metrics_aggregated,
+                "fit_round_end": str(datetime.datetime(2012, 12, 12, 12, 12, 12)),
             },
         },
     }
+    print(reporter.metrics)
```
jewelltaylor, line 150: nit: Think you forgot to remove this print statement

jewelltaylor (commented):
All in all, the changes look great! I have a few minor suggestions, feel free to reply if any of them don't make sense. I will take one more pass through once you fix the merge conflicts and have a chance to go through the PR again. Then we can get this merged in early this week

fl4health/clients/ditto_client.py
scarere: Yeah my bad I prefer to have my formatter set at 90 characters instead of 120 but I'll turn off the autoformat on save setting so that I don't mess with other peoples files.

fl4health/clients/evaluate_client.py
scarere: I thought about this and kind of was 50/50 on it. My initial logic for not including one was that I thought it would be more lines of code than just manually doing it every time. Not sure if that assumption is still true. Happy to add one, or to add equivalent methods to basic client. What do you think?

fl4health/reporting/base_reporter.py
scarere: lol

fl4health/reporting/base_reporter.py
scarere: No batch == local steps. I agree the documentation is a bit unclear and needs updating here. The reason I used batch was because step means something else for the wandb reporters, however I agree its inconsistent with the terminology used throughout basic client. Happy to just change it to step if you think that is better

fl4health/reporting/base_reporter.py
scarere: My intention was to assume they do nothing unless overridden. In most cases I felt that they would not be needed and didn't want to force subclasses to override them just to pass

fl4health/reporting/json_reporter.py
scarere: Can't think of any cases now where we would want the keys to be something other than strings, was just trying to keep it general. But i'm happy to specify dict[str, Any] and let someone else change it if they need non-string keys

emersodb:
I'm not sure why these aren't failing the pre-commit checks, but it looks like there are a few places where the `MetricsReporter` class still exists on this branch. Namely:
- fl4health/clients/model_merge_client.py
- fl4health/server/model_merge_server.py
- fl4health/server/adaptive_constraint_servers/ditto_server.py
- fl4health/server/adaptive_constraint_servers/mrmtl_server.py
It also appears to be mentioned in the comments of 
- tests/smoke_tests/run_smoke_test.py

tests/losses/test_mkmmd_loss.py
```diff
@@ -35,7 +35,13 @@
...
 )
 
-DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+# Basically all the tests failed if DEVICE was "cuda:0"
+# Likely the only reason these passed were because the github actions server does not
+# have a gpu so DEVICE would default to cpu
+# I started th debugging, but this is outside the scope of my PR and the tests run
+# super quickly on cpu anyways
+# DEVICE = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+DEVICE = torch.device("cpu")
```
emersodb, line 44: This has been fixed so feel free to merge and then drop this comment 🙂

tests/test_utils/assert_metrics_dict.py
```diff
@@ -0,0 +1,55 @@
+from typing import Any, Optional
+
+from pytest import approx
+
+DEFAULT_TOLERANCE = 0.0005
+
+
+def assert_metrics_dict(metrics_to_assert: dict[str, Any], metrics_saved: dict[str, Any]) -> list[str]:
```
emersodb, line 8: Maybe add a docstring here to help orient the user on what this is doing etc?

tests/test_utils/assert_metrics_dict.py
```diff
@@ -0,0 +1,55 @@
...
+    errors = []
+
+    def _assert(value: Any, saved_value: Any) -> Optional[str]:
```
emersodb, line 11: Any reason to make this an inline function rather than popping it out of here into it's own separate function above this method?

emersodb (commented):
Overall, really like the changes. Just a few suggestions and questions associated with the various components.

fl4health/reporting/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .json_reporter import JsonReporter
+from .wandb_reporter import WandBReporter
+
+# Must add unused imports to __all__ so that flake8 knows whats going on
```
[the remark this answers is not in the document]
scarere: Basically flake8 will raise an error during the pre-commit CI if there are any unused imports. However often in init files you want to add unused imports to make certain modules available in a different namespace. Therefore to let flake know whats going on, you have to add them to the builtin __all__ variable, then flake8 recognizes what you are trying to do and will not raise an error. By adding the modules to the __all__ variable, they will automatically be imported when someone does `from fl4health.reporting import *`.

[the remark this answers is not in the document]
emersodb: That's helpful. Thanks for the explanation. I guess the follow up question is where do we have these 
unused imports that flake is misidentifying? I'm probably just missing where that would be happening, but I didn't see any when I read through.

fl4health/reporting/wandb_reporter.py
```diff
@@ -0,0 +1,123 @@
...
+    EPOCH = "epoch"
+    BATCH = "batch"
+
+
+# TODO: Add ability to parse data types and save certain data types in specific ways
+# (eg. Artifacts, Tables, etc.)
+
+
+class WandBReporter(BaseReporter):
+    def __init__(self, step_type: StepType | str = StepType.ROUND, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
scarere: In general I agree but in this instance I thought it'd be fine since all the kwargs go straight to the wandb.init function detailed [here](https://docs.wandb.ai/ref/python/init/). I want to expose all the arguments but didn't want to unnecessarily clutter the wandbreporter init method

fl4health/clients/basic_client.py
```diff
@@ -114,13 +114,12 @@ def __init__(
...
         self.initial_weights: Optional[NDArrays] = None
 
-        self.wandb_reporter: Optional[ClientWandBReporter] = None
         self.total_steps: int = 0  # Need to track total_steps across rounds for WANDB reporting
 
         # Attributes to be initialized in setup_client
         self.parameter_exchanger: ParameterExchanger
         self.model: nn.Module
-        self.optimizers: Dict[str, torch.optim.Optimizer]
+        self.optimizers: dict[str, torch.optim.Optimizer]
```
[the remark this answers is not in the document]
scarere: yep, I believe typing.Dict will be deprecated in the future

fl4health/clients/basic_client.py
```diff
@@ -261,12 +262,13 @@ def fit(self, parameters: NDArrays, config: Config) -> Tuple[NDArrays, int, Dict
...
 
         Returns:
-            Tuple[NDArrays, int, Dict[str, Scalar]]: The parameters following the local training along with the
+            Tuple[NDArrays, int, dict[str, Scalar]]: The parameters following the local training along with the
             number of samples in the local training dataset and the computed metrics throughout the fit.
 
         Raises:
             ValueError: If local_steps or local_epochs is not specified in config.
         """
+        round_start_time = datetime.datetime.now()
```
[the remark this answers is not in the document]
scarere: Not exactly sure how this would look. The reports manager is called frequently every epoch, every round and every step, so it would be tough for it to track where it is in the training loop. But maybe I am missing something here

fl4health/losses/mkmmd_loss.py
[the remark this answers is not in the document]
scarere: my formatter is on 120 now

fl4health/reporting/json_reporter.py
```diff
@@ -0,0 +1,90 @@
...
+from logging import INFO
+from pathlib import Path
+from typing import Any
+
+from flwr.common.logger import log
+
+from fl4health.reporting.base_reporter import BaseReporter
+
+
+class DateTimeEncoder(json.JSONEncoder):
```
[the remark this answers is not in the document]
scarere: It was an artifact from the old metrics reporter. I changed all the reporting to convert the datetimes to strings before reporting them instead of only when dumping. I'm not sure if there was a reason to have the actual datetime objects in the reports dictionary but didn't want to have to add a check in the wandb reporter looking for invalid types and converting them.

fl4health/reporting/json_reporter.py
```diff
@@ -0,0 +1,90 @@
...
+                will be saved in a file named {output_folder}/{run_id}.json. Optional,
+                default is "metrics".
+        """
+        self.run_id = run_id
+
+        self.output_folder = Path(output_folder)
+        self.metrics: dict[str, Any] = {}
+
+        self.output_folder.mkdir(exist_ok=True)
+        assert self.output_folder.is_dir(), f"Output folder '{self.output_folder}' is not a valid directory."
```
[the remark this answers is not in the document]
scarere: I have no idea, I copy pasted this from the original metrics reporter. Can check and delete if it is unnecessary

fl4health/reporting/json_reporter.py
```diff
@@ -0,0 +1,90 @@
...
+    def report(
+        self,
+        data: dict[str, Any],
+        round: int | None = None,
+        epoch: int | None = None,
+        batch: int | None = None,
+    ) -> None:
+        if round is None:  # Reports outside of a fit round
+            self.metrics.update(data)
+        # Ensure we don't report for each epoch or step
```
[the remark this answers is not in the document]
scarere: Basically the json reporter reports only once a round by default. Therefore if epoch or step are not None, that means the reporter has been sent information at the epoch or step level, which we want to ignore. I can update the documentation. The docstring for report in base reporter describes this in more detail

fl4health/reporting/wandb_reporter.py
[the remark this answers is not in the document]
emersodb: Yeah I think you're right. Perhaps we can just limit ourselves to the ones we absolutely expect to be there and the rest can be in kwargs. Perhaps we can also put a note in the documentation to where a user might find the args for the WandB object?

fl4health/clients/basic_client.py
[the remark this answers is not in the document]
emersodb: Yeah I see what you mean. I guess I was just trying to pull this responsibility out of basic client but perhaps its not possible.

fl4health/clients/basic_client.py
```diff
@@ -1314,7 +1328,7 @@ def load_client_state(self) -> None:
 
         narrow_dict_type_and_set_attribute(self, ckpt, "client_name", "client_name", str)
         narrow_dict_type_and_set_attribute(self, ckpt, "total_steps", "total_steps", int)
-        narrow_dict_type_and_set_attribute(self, ckpt, "metrics_reporter", "metrics_reporter", MetricsReporter)
+        narrow_dict_type_and_set_attribute(self, ckpt, "reporters", "reporters", list[BaseReporter])
```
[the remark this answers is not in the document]
emersodb: Yeah, this was a bad suggestion 😂. Neither list[...] nor Sequence[...] work ha

fl4health/reporting/json_reporter.py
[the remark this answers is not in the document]
emersodb: You can also just reference that documentation here if you think that makes sense.

tests/test_utils/assert_metrics_dict.py
scarere: I didn't write this test. A lot of of the code in this pr was previously in the main branch, I'm not sure why it's showing it as having been added by me

scarere: Lol not sure if you'll get my last comment but I actually did write this. I forgot because I just copied it from the smoke tests. My bad. But yeah it was an inline function because I just copy pasted the method from the smoke tests

scarere: I will clean it up

fl4health/reporting/base_reporter.py
```diff
@@ -0,0 +1,51 @@
...
+        """Method for initializing reporters with client/server information
+
+        This method is called once by the client or server during initialization.
```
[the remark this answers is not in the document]
scarere: This is enforced in the wandb reporter. I will similarly add this to the json reporter as well. Does it make more sense to add it to the base class and have the subclasses call the parent method, or to just let subclasses create an initialized flag on their own?

fl4health/reporting/json_reporter.py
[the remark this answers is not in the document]
scarere: I'm going to delete this class for now, but if there are any reasons the datetime values can't be strings then let me know. I will also change the typing on the data argument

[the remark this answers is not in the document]
scarere: The reason is that I believe wandb does not support arbitrary objects. They must be scalars or wandb supported objects

fl4health/reporting/base_reporter.py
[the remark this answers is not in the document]
emersodb: I was thinking we could put it in the base class. However, I'm also okay with leaving it to the subclasses if you'd rather do that.

fl4health/reporting/json_reporter.py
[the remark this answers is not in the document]
emersodb: Sounds good to me.

fl4health/reporting/json_reporter.py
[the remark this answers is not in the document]
scarere: I think it does so we should be good to delete the assertion. I can only test on linux which raises an error if there are forward slashed om the pathname which is the only disallowed character. Windows has some other disallowed characters but I think were safe to delete the assertion for now

fl4health/server/base_server.py
```diff
@@ -32,42 +30,56 @@ def __init__(
...
 
-        if metrics_reporter is not None:
-            self.metrics_reporter = metrics_reporter
-        else:
-            self.metrics_reporter = MetricsReporter()
+        self.reporters = [] if reporters is None else list(reporters)
+        for r in self.reporters:
+            r.initialize(id=self.server_name)
+
+    def report_centralized_eval(self, history: History, num_rounds: int) -> None:
```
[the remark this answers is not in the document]
scarere: This isn't really state tracking so much as it is parsing the server specific history object. Since the centralised metrics can not be returned by fit or eval methods (they are the outputs of an eval function passed to the strategy) they can only be obtained by either parsing this history object, calling the strategy.eval function manually (which would cause it to run twice) or overriding the flwr.server.fit parent method entirely. I opted to parse the history object. The parsing would be a lot simpler if they hadn't packages the metrics as list of 2d tuples with the metric value itself and the server round.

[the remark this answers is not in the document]
scarere: given that the server runs the eval every round, packaging the round explicitly seems unnecessary, but it is what it is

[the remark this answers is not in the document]
scarere: Anyways since this functionality is specific to the server class, i don't think it can be moved into the report manager. Resolving for now, but open to discussion if you have any suggestions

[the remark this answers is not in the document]
emersodb: Yeah I think this is a good argument against it. Sounds good.

tests/clients/test_evaluate_client.py
```diff
@@ -117,12 +117,17 @@ def test_metrics_reporter_evaluate() -> None:
     reporter = JsonReporter()
     evaluate_client = MockEvaluateClient(loss=test_loss, metrics=test_metrics, reporters=[reporter])
     evaluate_client.evaluate([], {})
+    print(reporter.metrics)
```
emersodb, line 120: We can toss this print statement I think?

emersodb (approved):
Looks good to go. We have a small left over print statement I suggest we drop, but other than that a really nice refactor.

tests/clients/test_evaluate_client.py
scarere: good catch

PR #255, opened 2024-10-10
Fixing the tests for Mk-MMD loss such that they still pass properly on a GPU and CPU

emersodb:
# PR Type
Fix

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/868a6uduu)

The Tests for Mk-MMD loss were not properly handling device allocation if run on a GPU. This PR fixes that issue along with a small device bug in the loss function when a tensor of gammas were manually provided rather than created by default.

# Tests Added

N/A


jewelltaylor (approved):
LGTM!

PR #262, opened 2024-10-25
Nnunet server patch

jewelltaylor:
# PR Type

# Short Description
Update NnunetServer and FlServer to load checkpointed model into server_model attribute as it is needed when the _hydrate_model_for_checkpointing is called. Made this change before I launched the FL4Health Experiments on PICAI but forgot to push until now.


emersodb (approved):
Good to go.

PR #264, opened 2024-10-25
Turning follow imports on in mypy for better coverage

emersodb:
# PR Type
Feature

# Short Description

Fixing an issue where we had mypy follow imports turned off across the board. Moved to just ignoring libraries with bad typing or no stubs.

Prior to this, we could have invalid imports that were not picked up by mypy. These will now be properly picked up and throw errors. In the process, we need to ignore underlying external library issues.

# Tests Added

None. Did fix several bugs that were revealed by this change.


jewelltaylor (approved):
LGTM!

PR #266, opened 2024-10-28
try except block for dataset.json in case nnunet_raw is unavailable

scarere:
# PR Type

# Short Description

My nnunet_raw server is down but i have the nnuner_preprocessed locally. In theory I should be able to run the nnunet client since the dataset is already preprocessed, but currentl the client loads the dataset.json from the nnunet_raw directory by default. I added a try except block to try and load the dataset json from the preprocessed folder as a backup.

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -5,7 +5,7 @@ import pickle
 import time
 import warnings
 from contextlib import redirect_stdout
-from logging import DEBUG, INFO, WARNING
+from logging import DEBUG, ERROR, INFO, WARNING
 from os.path import exists, join
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -157,7 +157,20 @@ def __init__(
         # Some nnunet client specific attributes
         self.dataset_id: int = dataset_id
         self.dataset_name = convert_id_to_dataset_name(self.dataset_id)
-        self.dataset_json = load_json(join(nnUNet_raw, self.dataset_name, "dataset.json"))
+        try:
+            self.dataset_json = load_json(join(nnUNet_raw, self.dataset_name, "dataset.json"))
+        except Exception as e:
+            self.dataset_json = load_json(join(nnUNet_preprocessed, self.dataset_name, "dataset.json"))
+            e1 = e
```
emersodb, line 164: Maybe a dumb question, but why not just do
`except Exception as e1:`
in the catch above, rather than doing this set?

emersodb: hmm...perhaps the of e is limited? It's just a touch awkward feeling, but perhaps this is the "canonical" way to do this.

emersodb (approved):
Looks alright to me. However, I might suggest taking 2 minutes to make sure there isn't a "cleaner" way to do the e1, e2 logging.

fl4health/clients/nnunet_client.py
scarere: the linters were complaining as I believe the as exists only within the scope of the except. However in testing I realized we can't actually have two except blocks with the same exception

emersodb: I like the refactor. Thanks for looking into it.

PR #268, opened 2024-10-29
Modify LR Scheduler + Test

jewelltaylor:
# PR Type

# Short Description
Modfiy LR scheduler to have fixed values over defined window size. 

# Tests Added

Added test to verify the scheduler anneals the learning rate as expected and produces the appropriate log message when exceed the max_steps it was initialized with.

### DIFF
--- a/fl4health/utils/nnunet_utils.py
+++ b/fl4health/utils/nnunet_utils.py
@@ -6,6 +6,7 @@ import warnings
 from enum import Enum
 from importlib import reload
 from logging import DEBUG, INFO, WARN, Logger
+from math import ceil
 from typing import Any, Callable, Dict, List, Sequence, Tuple, Union, no_type_check
 
 import numpy as np
@@ -438,7 +439,7 @@ class PolyLRSchedulerWrapper(_LRScheduler):
         self.exponent = exponent
         self.steps_per_lr = steps_per_lr
         # Number of windows with constant LR across training
-        self.num_windows = round(max_steps / self.steps_per_lr)
+        self.num_windows = ceil(max_steps / self.steps_per_lr)
         self._step_count: int
         super().__init__(optimizer, -1, False)
 
--- a/fl4health/utils/nnunet_utils.py
+++ b/fl4health/utils/nnunet_utils.py
@@ -453,8 +453,11 @@ class PolyLRSchedulerWrapper(_LRScheduler):
         Returns:
             Sequence[float]: A uniform sequence of LR for each of the parameter groups in the optimizer.
         """
-        if self._step_count > self.max_steps:
-            log(WARN, f"Current LR step of {self._step_count} exceeds Max Steps of {self.max_steps}")
+        if self._step_count == self.max_steps + 1:
+            log(
+                WARN,
+                f"Current LR step of {self._step_count} reached Max Steps of {self.max_steps}. LR will remain fixed.",
+            )
 
         curr_step = min(self._step_count, self.max_steps)
         curr_window = int(curr_step / self.steps_per_lr)
--- a/fl4health/utils/nnunet_utils.py
+++ b/fl4health/utils/nnunet_utils.py
@@ -5,7 +5,7 @@ import sys
 import warnings
 from enum import Enum
 from importlib import reload
-from logging import DEBUG, INFO, Logger
+from logging import DEBUG, INFO, WARN, Logger
 from typing import Any, Callable, Dict, List, Sequence, Tuple, Union, no_type_check
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


fl4health/utils/nnunet_utils.py
```diff
@@ -414,19 +415,53 @@ def flush(self) -> None:
...
     @no_type_check
     def get_lr(self) -> Sequence[float]:
+        """
+        Get the current LR of the scheduler.
+
+        Returns:
+            Sequence[float]: A uniform sequence of LR for each of the parameter groups in the optimizer.
+        """
+        if self._step_count > self.max_steps:
+            log(WARN, f"Current LR step of {self._step_count} exceeds Max Steps of {self.max_steps}")
```
emersodb, line 457: Perhaps in the warning here, we can say what we'll "do" when this occurs. For example, "Learning rate will remain fixed" or the like?

emersodb: The warning is also going to repeat every step I think? Perhaps we can only produce the warning "at decay time"? Perhaps this won't be too much of a problem. Just thought it might be a lot of logs.

tests/utils/nnnunet_utils_test.py
```diff
@@ -0,0 +1,45 @@
...
+    steps_per_lr = 100
+    initial_lr = 0.5
+
+    model = MnistNetWithBnAndFrozen()
+    opt = SGD(model.parameters(), lr=initial_lr)
+    lr_scheduler = PolyLRSchedulerWrapper(
+        optimizer=opt, max_steps=max_steps, initial_lr=initial_lr, exponent=exponent, steps_per_lr=steps_per_lr
+    )
+
+    assert lr_scheduler.num_windows == round(max_steps / steps_per_lr)
```
emersodb, line 23: Just a suggestion, but if the calculation is simple, I tend to inject it explicitly here rather than sort of duplicating the code just in case my code is somehow hiding something wrong. That is, here I would just do
```
assert lr_scheduler.num_windows == 10.0
```

tests/utils/nnnunet_utils_test.py
```diff
@@ -0,0 +1,45 @@
...
+    assert lr_scheduler.initial_lr == initial_lr
+
+    prev_lr = initial_lr
+    for step in range(1, max_steps + 1):
```
emersodb, line 27: I'm sure this calculation is fast, but maybe we can reduce both max_steps and steps_per_lr by an order of magnitude (100, 10) just to speed up this loop a bit?

tests/utils/nnnunet_utils_test.py
```diff
@@ -0,0 +1,45 @@
...
+        curr_lr = lr_scheduler.get_lr()[0]
+
+        if step % steps_per_lr == 0:
+            assert curr_lr != prev_lr
+        else:
+            assert curr_lr == prev_lr
+
+        prev_lr = curr_lr
+
+        if step < max_steps:
```
emersodb, line 37: I might be wrong, but I think because of our loop bounds, this will always be true. So perhaps we just drop the if? If it's not always true, we can just adjust the range bounds so it is true and then the next step pops the warning and still drop the if I think

emersodb (approved):
It looks good. None of my comments are deal breakers, but if you think they are useful feel free to incorporate them before merging.

tests/utils/nnnunet_utils_test.py
jewelltaylor: Since the loop is until max_steps + 1, on the last iteration this condition will be False. I don't want to change bound down to max_steps because at the beginning of each iteration I get the current lr and check whether its correct (same or different as prev lr depending on step number). So I want to  perform this check on the last iteration but not perform the step as it would bring us to max_steps. For this, after the loop I perform an additional step and test that a warning is logged. Let me know if I am missing something or making a silly mistake :)

emersodb: Yup, I see. That works for me.

fl4health/utils/nnunet_utils.py
```diff
@@ -414,19 +415,56 @@ def flush(self) -> None:
...
+        """
+        Learning rate (LR) scheduler with polynomial decay across fixed windows of size steps_per_lr.
+
+        Args:
+            optimizer (Optimizer): The optimizer to apply LR scheduler to.
+            initial_lr (float): The initial learning rate of the optimizer.
+            max_steps (int): The maximum total number of steps across all FL rounds.
+            exponent (float): Controls how quickly LR descreases over time. Higher values
+                lead to more rapdid descent.
+            steps_per_lr (int): The number of steps per LR before decaying.
```
scarere, line 434: Maybe add a bit more of an explanation as to what this means as it might be ambigious without context.

fl4health/utils/nnunet_utils.py
```diff
@@ -414,19 +415,56 @@ def flush(self) -> None:
...
+            log(
+                WARN,
+                f"Current LR step of {self._step_count} reached Max Steps of {self.max_steps}. LR will remain fixed.",
+            )
+
         curr_step = min(self._step_count, self.max_steps)
-        new_lr = self.initial_lr * (1 - curr_step / self.max_steps) ** self.exponent
+        curr_window = int(curr_step / self.steps_per_lr)
+
+        new_lr = self.initial_lr * (1 - curr_window / self.num_windows) ** self.exponent
```
scarere, line 465: Although I know this will return the same lr within a window, would it make sense to have it within the if block below and then return the result of `self.get_last_lr` otherwise?

fl4health/utils/nnunet_utils.py
scarere: Also maybe add a note that the default value is 250 because that is what's used as a default by nnunet

scarere (approved):
Couple minor comments but overall looks good to me

fl4health/utils/nnunet_utils.py
jewelltaylor: Good idea!

fl4health/utils/nnunet_utils.py
jewelltaylor: I think I might keep it as is since the amount of compute to get the lr is trivial and I don't trust the LRScheduler class after seeing the internals hahah

scarere: Honestly fair, I don't trust that method much either.

PR #273, opened 2024-10-30
modified progress bar formatting to inherit from flwr formatting

scarere:
# PR Type

# Short Description

I had hardcoded progress bar formatting before. Now it will inherit the same formatting used by FLWR dynamically so that we don't need to track changes made to flower log formatting.


emersodb (approved):
I'm going to trust that you know what this does, since I definitely don't 😂

PR #274, opened 2024-10-31
Modify curr step of LR Scheduler to start at 0 instead of 1

jewelltaylor:
# PR Type

# Short Description

Since the step count of the LR scheduler starts at 1 (imposed by pytorch), the final step will always log that it exceeds the max step count. This PR simply subtracts 1 when calculating the current step to avoid this. 

# Tests Added

Update LR scheduler test accordingly.

### DIFF
--- a/fl4health/utils/nnunet_utils.py
+++ b/fl4health/utils/nnunet_utils.py
@@ -462,7 +462,7 @@ class PolyLRSchedulerWrapper(_LRScheduler):
                 f"Current LR step of {self._step_count} reached Max Steps of {self.max_steps}. LR will remain fixed.",
             )
 
-        # Subtract 1 from step count since it starts at 1
+        # Subtract 1 from step count since it starts at 1 (imposed by PyTorch)
         curr_step = min(self._step_count - 1, self.max_steps)
         curr_window = int(curr_step / self.steps_per_lr)
 
--- a/fl4health/utils/nnunet_utils.py
+++ b/fl4health/utils/nnunet_utils.py
@@ -455,14 +455,15 @@ class PolyLRSchedulerWrapper(_LRScheduler):
         Returns:
             Sequence[float]: A uniform sequence of LR for each of the parameter groups in the optimizer.
         """
-        if self._step_count == self.max_steps + 1:
+
+        if self._step_count - 1 == self.max_steps + 1:
             log(
                 WARN,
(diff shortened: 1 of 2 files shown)


fl4health/utils/nnunet_utils.py
```diff
@@ -461,12 +461,13 @@ def get_lr(self) -> Sequence[float]:
                 f"Current LR step of {self._step_count} reached Max Steps of {self.max_steps}. LR will remain fixed.",
             )
 
+        # Subtract 1 from step count since it starts at 1
```
emersodb, line 464: I might be missing something, but are we actually subtracting one here?

jewelltaylor: Oops lol ... updated and added assert that the log message doesn't trigger early and only after the number of steps exceeds max steps

scarere: Seconded. Also can you add a comment noting that pytorch imposes that step_count starts at 1.

scarere: Also it seems like step count might actually start at 2 since pytorch calls step on initialization of the LR scheduler. Not sure why they do this.

scarere: Maybe I'm missing somehing, but it seems like the self._initialize_step method of the pytorch base class performs a step on init before training. So by the time step is called after the first training step, self._step_count will equal 2

emersodb: If true...that's really silly 😂

emersodb: mmm that makes sense. Perhaps their counter starts at zero and bumps from there. Still a weird design choice.

scarere: They start it at zero, and call step on init, but within the step function, self._step_count is incremented before calling self.get_lr. That being said maybe there is something else going on under the hood that I'm missing. I trust that your test confirms that we have the correct LR at each step.

scarere (approved):
LGTM

fl4health/utils/nnunet_utils.py
jewelltaylor: @scarere what a nightmare ... let me know if it pops up on your side, its not exhaustive testing but still would be surprised if a magic +1 gets added in under certain conditions

PR #279, opened 2024-11-07
Lowercasing torch device variables throughout library when not global

emersodb:
# PR Type
Other

# Short Description

Clickup Ticket(s): N/A

Just a small syntax cleanup PR that lowercases the torch device variable in the library throughout when it isn't being used as a global variable. This is a bit more standard practice.

# Tests Added

N/A. This is strictly a name change. So shouldn't change functionality.


jewelltaylor (approved):
LGTM!

PR #280, opened 2024-11-08
First pass at optionally shutting down server (and remaining clients) on failures

emersodb:
# PR Type
Feature/Fix

# Short Description

Clickup Ticket(s): N/A

There is an odd default behavior in flower servers wherein, if all clients die due to an exception for example, the server process essentially becomes a "zombie." That is, in spite of receiving failure messages from clients, its process persists indefinitely waiting for client messages to continue training. 

This can be quite troublesome for debugging, testing, or just experimentation because the persisting server can interfere with future runs if not manually killed (via htop or other process monitor). This is because it will still be listening on the default port and may intercept messages from clients that are spun up in a new process kicked off by the user.

At first blush, we might use the server config timeout to help address this issue. While this is possible, it means that you need to have a fairly reasonable estimate of how long training will take on the client sides, which a user may not have. So I implemented an option in our base server to at least say that the server doesn't tolerate failures in the FL process. If a failure is received, the server will send kill signals to any remaining clients and terminate its processes with an error.

**Note:** I didn't disable `accept_failures` in our examples. However, if we think it makes sense to do so, at least in the places where we are training with 100% of available clients, I'm happy to add that option.

# Tests Added

I tested this on our Basic Example by throwing an exception right at the start of training on each client. See attachment for what it looks like on the server side.
[server.txt](https://github.com/user-attachments/files/17680100/server.txt)


fl4health/server/base_server.py
```diff
@@ -386,13 +415,12 @@ def __init__(
...
+                evaluation from clients or not. If set to False, this will cause the server to shutdown all clients
+                and throw an exception. Defaults to True.
         """
         super().__init__(
-            client_manager,
-            strategy,
-            reporters,
-            checkpointer,
-            server_name=server_name,
+            client_manager, strategy, reporters, checkpointer, server_name=server_name, accept_failures=accept_failures
```
scarere, line 423: Will this affect how this server should do checkpointing? Currently it saves the server state after the evaluate round. So if the fit round has no failures, but the evaluate round does, we lose the parameters after the successful fit round. Is this the behaviour we want or should we move the per round checkpointing to occer after the fit round but before the evaluate round?

fl4health/server/scaffold_server.py
```diff
@@ -22,37 +22,39 @@ def __init__(
...
-                reporters which the server should send data to before and after each
-                round.
-            warm_start (bool, optional): Whether or not to initialize control variates
-                of each client as local gradients. The clients will perform a training
-                pass (without updating the weights) in order to provide a "warm"
-                estimate of the SCAFFOLD control variates. If false, variates are
-                initialized to 0. Defaults to False.
+            checkpointer (Optional[TorchCheckpointer], optional): To be provided if the server should perform server
+                side checkpointing based on some criteria. If none, then no server-side checkpointing is performed.
+                Defaults to None.
```
scarere, line 40: lol sorry the formatting to a shorter line length was definitely my fault

scarere (commented):
Looks good, a couple questions but other than that I think this is a much needed feature

fl4health/server/base_server.py
```diff
@@ -345,6 +364,15 @@ def evaluate_round(
 
         return eval_round_results
 
+    def _terminate_after_unacceptable_failures(self, timeout: Optional[float]) -> None:
+        assert not self.accept_failures
+        # First we shutdown all clients involved in the FL training/evaluation if they can be.
+        self.disconnect_all_clients(timeout=timeout)
```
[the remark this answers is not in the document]
emersodb: From what I can tell they don't. In the base server code they just send out this signal and let the program finish as the termination. However, it is worth calling our own shutdown method before throwing the exception. So I'll add that.

fl4health/server/base_server.py
```diff
@@ -131,9 +132,18 @@ def fit_round(
...
         if fit_round_results is not None:
-            _, metrics, _ = fit_round_results
+            _, metrics, (_, failures) = fit_round_results
             self.reports_manager.report({"fit_metrics": metrics}, server_round)
 
+            num_failures = len(failures)
+            if num_failures > 0 and not self.accept_failures:
+                log(
+                    ERROR,
+                    f"There were {num_failures} failures in the fitting process. This will result in termination of "
```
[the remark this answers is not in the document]
scarere: yeah i wish there was a way to identify failures that are the exact same.

fl4health/server/base_server.py
emersodb: That's a good question. My initial thought is that it's alright if we lose a single FL round due to failures on the evaluation side. My reasoning would be that even though we finished a train cycle, we never actually got to see what the "results" of that cycle and aggregation were for the clients. So re-starting from those trained parameters is sort of "blind." That is, we don't "know" how those parameters actually did in that round. Just my initial thoughts.

fl4health/server/base_server.py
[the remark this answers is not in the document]
scarere: without using regex lol

fl4health/server/scaffold_server.py
emersodb: No worries. Just moving it back where I see it 🙂

fl4health/server/base_server.py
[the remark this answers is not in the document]
emersodb: True...but knowing which clients failed with which errors could also be important. So I think we just kind of live with the possibility that there were a lot of failures.

fl4health/server/base_server.py
scarere: Yeah that makes sense. I was thinking the same thing which is why I was unsure. I think John is the one who primarily uses the checkpointing server so maybe he'll have an opinion

scarere: I'm fine with whatever though

emersodb: Sounds good. We'll see what @jewelltaylor thinks 🙂

fl4health/server/base_server.py
```diff
@@ -369,10 +362,32 @@ def _terminate_after_unacceptable_failures(self, timeout: Optional[float]) -> No
...
+        log(
+            ERROR,
+            f"There were {len(failures)} failures in the fitting process. This will result in termination of "
+            "the FL process",
+        )
+        for failure in failures:
+            if isinstance(failure, BaseException):
+                log(
+                    ERROR,
+                    "An exception was returned instead of any failed results. As such the client ID is unknown. "
```
scarere, line 380: This might be a bit overkill so feel free to ignore this suggestion. But could you not technically determine the client ids of the clients that returned exceptions by process of elimination? If you have multiple exceptions you won't know which exception goes with which client, but you could at least log a list of clients that failed.

emersodb: Yup. That's a fair point. I'll log a list of clients that failed. We won't know which corresponds to which exception, but it's reasonable to at least gather them.

emersodb: So it turns out this is harder than it seems at first. Basically, if they return only an exception, they are immediately removed from the client manager and they aren't necessarily connected to the server yet when the first fit_round starts (for example). So the representation of these CIDs in the client manager is a bit asynchronous, which makes these reporting really difficult. I'm going to just punt on it for now and we can try to be more verbose later.

fl4health/server/base_server.py
jewelltaylor: Agree with you two, we should only consider the round complete after evaluation

fl4health/server/base_server.py
```diff
@@ -663,7 +708,7 @@ def _get_initial_parameters(self, server_round: int, timeout: Optional[float]) -
             log(INFO, "Received initial parameters from one random client")
         else:
             log(
-                WARN,
```
emersodb, line 666: Moved to WARNING so we can drop a duplicate import.

jewelltaylor (approved):
LGTM! It will be nice to not have to deal the zombie processes :)

PR #281, opened 2024-11-08
Adding a touch of logging for checkpointing errors

emersodb:
# PR Type
"Fix"

# Short Description

In doing some debugging for the numerical fluctuations fix, I encountered a small issue where torch.save was failing, but the exception wasn't ending up in the logs. This made my life sort of hard and I thought I'd just fix it up in a short PR.

# Tests Added

None


tests/utils/sampler_test.py
```diff
@@ -177,7 +177,7 @@ def test_dirichlet_sampler_without_hash_key() -> None:
 
     _, p_val_1 = chisquare(f_obs=new_samples_per_class_1, f_exp=samples_per_class)
     _, p_val_2 = chisquare(f_obs=new_samples_per_class_2, f_exp=samples_per_class)
-    _, p_val_3 = chisquare(f_obs=new_samples_per_class_1, f_exp=new_samples_per_class_2)
+    _, p_val_3 = chisquare(f_obs=new_samples_per_class_2, f_exp=new_samples_per_class_1)
```
emersodb, line 180: I noticed a div by zero warning was popping out from this test. It's caused by the fact that the `new_samples_per_class_2` distribution has one class with 0 elements. `new_samples_per_class_1` doesn't have this issue. So making it the reference distribution and the other the test distribution eliminates the error and should be effectively the same statistical test.

fl4health/checkpointing/checkpointer.py
```diff
@@ -95,7 +95,12 @@ def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: Dict[str, Sca
                 f"{self.comparison_str} Best score ({self.best_score})",
             )
             self.best_score = comparison_score
-            torch.save(model, self.best_checkpoint_path)
+            try:
+                log(INFO, f"Saving checkpoint as {str(self.best_checkpoint_path)}")
+                torch.save(model, self.best_checkpoint_path)
+            except Exception as e:
+                log(ERROR, f"Encountered the following error while saving the checkpoint: {e}")
+                raise e
```
emersodb, line 103: In experimentation, I was encountering an issue where saving the checkpoint would fail, but it wasn't surfacing properly in the logs. So this explicitly places it there.

jewelltaylor (approved):
LGTM!

PR #283, opened 2024-11-11
More Masked Layers for FedPM

yc7z:
# PR Type
[Feature]

Added masked implementations for transposed convolution, layer normalization, and batch normalization layers.

# Tests Added

Tests in `test_masked_layers` have been added to test the functionality of the new masked layers.


fl4health/reporting/wandb_reporter.py
```diff
@@ -110,6 +110,13 @@ def start_run(self, wandb_init_kwargs: dict[str, Any]) -> None:
             id=self.id,
             **wandb_init_kwargs  # Other less commonly used kwargs
         )
+
+        if isinstance(wandb_run, wandb.wandb_run.Run):
```
emersodb, line 114: Was this causing an issue somewhere that you saw? Just curious if this is somehow being missed by our mypy or something else.

fl4health/model_bases/masked_layers/masked_normalization_layers.py
```diff
@@ -0,0 +1,297 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch import Tensor
+from torch.nn.modules.batchnorm import _BatchNorm
+from torch.nn.parameter import Parameter
+
+from fl4health.utils.functions import bernoulli_sample
+
+_shape_t = Union[int, List[int], torch.Size]
```
emersodb, line 12: Perhaps we can name this something slightly more verbose? Maybe TorchShape?

fl4health/model_bases/masked_layers/masked_normalization_layers.py
```diff
@@ -0,0 +1,297 @@
...
+        Finally, the binary masks are applied to the weight and the bias. During training,
+        gradients with respect to the score tensors are computed and used to update the score tensors.
+
+        When elementwise_affine is False, nn.LayerNorm does not have weight or bias.
+        Under this condition, both score tensors are None and MaskedLayerNorm acts in the same way as nn.LayerNorm.
+
+        Note: the scores are not assumed to be bounded between 0 and 1.
+
+        Args:
+            normalized_shape (int or list or torch.Size): input shape from an expected input.
```
emersodb, line 41: Super minor, but I'd update the type for `normalized_shape` to be whatever you decide the type alias should called.

fl4health/model_bases/masked_layers/masked_linear.py
```diff
@@ -0,0 +1,92 @@
...
+        self.in_features = in_features
+        self.out_features = out_features
+        self.weight.requires_grad = False
+        self.weight_scores = Parameter(torch.randn_like(self.weight), requires_grad=True)
+        if bias:
+            assert self.bias is not None
+            self.bias.requires_grad = False
+            self.bias_scores = Parameter(torch.randn_like(self.bias), requires_grad=True)
+        else:
+            self.register_parameter("bias", None)
```
emersodb, line 56: If bias is None, do we need to register it as such here? For example, we don't do this for the normalization layers below.

fl4health/model_bases/masked_layers/masked_normalization_layers.py
```diff
@@ -0,0 +1,297 @@
...
+                weight_score: learnable scores for the weights. Has the same shape as weight. When applied
+                    to the default initial values of self.weight (i.e., all ones), this is equivalent to
+                    randomly dropping out certain features.
+                bias_score: learnable scores for the bias. Has the same shape as bias. When applied to
+                    the default initial values of self.bias (i.e., all zeros), it does not have any actual
+                    effect. Thus, bias_score only influences training when MaskedLayerNorm is created
+                    from some pretrained nn.LayerNorm module whose bias is not all zeros.
+        """
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__(num_features, eps, momentum, affine, track_running_stats, **factory_kwargs)
```
emersodb, line 184: Genuine question, can we not simply pass `device=device, dtype=dtype` at the end of the `super().__init__`? I'm not sure, but that would mean we don't have to do through the creation of the dictionary rigmarole 🙂

fl4health/model_bases/masked_layers/masked_normalization_layers.py
```diff
@@ -0,0 +1,297 @@
...
+            masked_batch_norm_module.bias = Parameter(batch_norm_module.bias.clone().detach(), requires_grad=False)
+            masked_batch_norm_module.bias_scores = Parameter(
+                torch.randn_like(batch_norm_module.weight), requires_grad=True
+            )
+        return masked_batch_norm_module
+
+
+class MaskedBatchNorm1d(_MaskedBatchNorm):
+    """
+    Applies (masked) Batch Normalization over a 2D or 3D input.
```
emersodb, line 270: Maybe we can mention that the 2D is batch x input input and 3D is batch x input x channels?

emersodb (approved):
Overall looks really nice. Some pretty minor comments!

fl4health/model_bases/masked_layers/masked_conv.py
```diff
@@ -441,23 +357,425 @@ def from_pretrained(cls, conv_module: nn.Conv3d) -> "MaskedConv3d":
...
+        )
+        self.weight.requires_grad = False
+        self.weight_scores = Parameter(torch.randn_like(self.weight), requires_grad=True)
+        if bias:
+            assert self.bias is not None
+            self.bias.requires_grad = False
+            self.bias_scores = Parameter(torch.randn_like(self.bias), requires_grad=True)
+        else:
+            self.register_parameter("bias_scores", None)
+        self.reset_parameters()
```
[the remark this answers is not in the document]
yc7z: This function is used to initialize the weight and bias tensor following the Kaiming uniform distribution since both the weight and bias are initialized via `torch.empty` in the parent class. But I now noticed `self.reset_parameters()` is already being called in the parent class so I've removed them here.

fl4health/reporting/wandb_reporter.py
yc7z: It was mypy that complained about this on my end. I assumed we want to make sure that the initialization returns a `Run` type hence the change here. Is this change okay?

fl4health/model_bases/masked_layers/masked_conv.py
```diff
@@ -441,23 +357,425 @@ def from_pretrained(cls, conv_module: nn.Conv3d) -> "MaskedConv3d":
...
+            masked_bias = bias_mask * self.bias
+        else:
+            masked_bias = None
+
+        return F.conv_transpose1d(
+            input, masked_weight, masked_bias, self.stride, self.padding, output_padding, self.groups, self.dilation
+        )
+
+    @classmethod
+    def from_pretrained(cls, conv_module: nn.ConvTranspose1d) -> "MaskedConvTranspose1d":
```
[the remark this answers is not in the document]
yc7z: That's good to know! I've switched over to this too.

fl4health/model_bases/masked_layers/masked_linear.py
yc7z: You are right. It is already registered as None in the parent class `nn.Linear`.

fl4health/model_bases/masked_layers/masked_normalization_layers.py
yc7z: Yes we definitely can 😂

fl4health/reporting/wandb_reporter.py
emersodb: Yes for sure. I'm just not sure why I'm not getting that popping up in mypy on my end. I think we recently upgraded mypy versions. Maybe just quickly check that you have the most recent one in your environment and if it fixes the mypy compliant by upgrading if you don't? Otherwise, I'm good to keep this around.

emersodb: Nice. Mystery solved 🙂

PR #285, opened 2024-11-12
Update deep mmd client and loss implementation

sanaAyrml:
# PR Type
[Fix]

This is a small fix for Deep MMD loss implementation to both impose more augmentation for training deep kernel and also not let them randomness change just because of Deep MMD loss initialization. Such a pin down would let the initial model for ditto and ditto deep mmd loss be same in case if we have same seed. Otherwise due to kernel initialization in Deep MMD loss we might end up having different results, even if we set deep_mmd_loss_weight zero.

# Tests Added

No tests added.


fl4health/losses/deep_mmd_loss.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         hidden_size: int = 10,
         output_size: int = 50,
         lr: float = 0.001,
-        training: bool = True,
+        training: bool = False,
```
emersodb, line 50: With this change, you'll want to update the init documentation default to False.

A more general question is what we expect to happen if the user sets `training = True` in creating the loss? I haven't thought this fully through, but it seems like the user would expect that to mean that the trainable parameters were, in fact, trainable. 

If we want the loss to always start with training off and that we (or the user) will be responsible for turning it on (and later off) at the right times, then we can probably just drop this argument entirely and set `self.training = False` explicitly to avoid any mismatch.

fl4health/losses/deep_mmd_loss.py
```diff
@@ -240,7 +242,10 @@ def train_kernel(self, X: torch.Tensor, Y: torch.Tensor) -> None:
         self.sigma_phi_opt.requires_grad = True
         self.epsilon_opt.requires_grad = True
 
-        features = torch.cat([X, Y], 0)
+        indices = torch.randperm(Y.size(0))
```
emersodb, line 245: Maybe just add a small comment as to what we're doing here? That is, we're shuffling the examples to ensure they aren't always presented in the same order for training.

fl4health/utils/random.py
```diff
@@ -60,6 +60,32 @@ def unset_all_random_seeds() -> None:
     torch.use_deterministic_algorithms(False)
 
 
+def save_random_state() -> Tuple[Tuple[Any, ...], Dict[str, Any], torch.Tensor]:
+    """
```
emersodb, line 64: I like the documentation, but it's not quite in the Google Docstring format here. That is, the return and its type aren't documented.

fl4health/utils/random.py
```diff
@@ -60,6 +60,32 @@ def unset_all_random_seeds() -> None:
...
+    random_state = random.getstate()
+    numpy_state = np.random.get_state()
+    torch_state = torch.get_rng_state()
+    return random_state, numpy_state, torch_state
+
+
+def restore_random_state(
+    random_state: Tuple[Any, ...], numpy_state: Dict[str, Any], torch_state: torch.Tensor
+) -> None:
+    """
```
emersodb, line 78: Similar comment as above around a slight augmentation to the docstring 🙂

fl4health/utils/random.py
```diff
@@ -60,6 +60,32 @@ def unset_all_random_seeds() -> None:
...
+    Restore the state of the random number generators for Python, NumPy, and PyTorch. This will allow you to restore
+    the state of the random number generators to a previously saved state.
+    """
+    log(INFO, "Restoring random state.")
+    random.setstate(random_state)
+    np.random.set_state(numpy_state)
+    torch.set_rng_state(torch_state)
+    return
```
emersodb, line 86: This return isn't necessary I don't think.

emersodb (commented):
Overall looks good. Just a few small comments/questions.

fl4health/losses/deep_mmd_loss.py
sanaAyrml: yes. Exactly that's why I want to have this off so people need to turn on training to activate it.

On another note training basically activates training of kernel multiple epochs, and then compute loss, however if it is off, it only computes loss.

I agree that we can omit it from input arguments.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -28,6 +28,8 @@ def __init__(
         checkpointer: Optional[ClientCheckpointModule] = None,
         deep_mmd_loss_weight: float = 10.0,
         feature_extraction_layers_with_size: Optional[Dict[str, int]] = None,
+        beta_global_update_interval: int = 20,
```
emersodb, line 31: Maybe we should call this something slightly different, since the trainable parameters in the DeepMMD loss aren't necessarily "betas"?

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -190,8 +287,9 @@ def compute_training_loss(
                 additional losses indexed by name. Additional losses includes each loss component and the global model
                 loss tensor.
         """
-        for layer_loss_module in self.deep_mmd_losses.values():
-            assert layer_loss_module.training
+        if self.beta_global_update_interval == -1:
```
emersodb, line 290: Do we only want to run this assertion when it's -1? If it's non-zero, it seems like maybe we'd want to do this assertion at some point too?

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -103,9 +114,95 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        self.local_feature_extractor.enable_accumulating_features()
+        self.initial_global_feature_extractor.enable_accumulating_features()
+
+        # Save the initial state of the local model to restore it after the buffer is populated,
+        # however as initial global model is already cloned and frozen, we don't need to save its state.
+        initial_state_local_model = local_model.training
+
+        # Set local model to evaluation mode, as we don't want to create a computational graph
+        # for the local model when populating the local buffer with features to compute optimal
+        # betas for the MK-MMD loss
```
emersodb, line 172: It looks like this comment is perhaps outdated?

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
emersodb: We might want to scrub mention of betas throughout, since it's used in a few other variable names too.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -103,9 +114,95 @@ def update_before_train(self, current_server_round: int) -> None:
...
+        if self.beta_global_update_interval > 0 and self._should_optimize_betas(step):
+            # Get the feature distribution of the local and initial global features with evaluation
+            # mode
+            local_distributions, initial_global_distributions = self.update_buffers(
+                self.model, self.initial_global_model
+            )
+            # Update betas for the Deep MMD loss based on gathered features during training
+            for layer, layer_deep_mmd_loss in self.deep_mmd_losses.items():
+                layer_deep_mmd_loss.training = True
+                _ = layer_deep_mmd_loss(local_distributions[layer], initial_global_distributions[layer])
```
emersodb, line 141: I don't think you need to "catch" this return value if you're not going to use it. You can just call the loss.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
```diff
@@ -103,9 +114,95 @@ def update_before_train(self, current_server_round: int) -> None:
...
+    def update_after_step(self, step: int, current_round: Optional[int] = None) -> None:
+        if self.beta_global_update_interval > 0 and self._should_optimize_betas(step):
+            # Get the feature distribution of the local and initial global features with evaluation
+            # mode
+            local_distributions, initial_global_distributions = self.update_buffers(
+                self.model, self.initial_global_model
+            )
+            # Update betas for the Deep MMD loss based on gathered features during training
+            for layer, layer_deep_mmd_loss in self.deep_mmd_losses.items():
+                layer_deep_mmd_loss.training = True
```
emersodb, line 140: Maybe a comment here that, with training being set to true, this computes **and** trains the loss

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: That is true, I should update its docstring too.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: Only if interval is -1 we want to activate training of Deep MMD kernel and keep it open, otherwise we will set it True update kernel in each interval and set it back again False, so I don't think we need to have this assertion somewhere else for other cases. Though we can have an assertion her for other cases to make sure it not in training mode.

fl4health/clients/deep_mmd_clients/ditto_deep_mmd_client.py
sanaAyrml: Agreed. This comment should be updated.

emersodb (approved):
Changes look good to me

PR #286, opened 2024-11-12
Moving relevant files to use the future annotations import for 'in-class' return values

emersodb:
# PR Type
Other

# Short Description

Clickup Ticket(s): [Link](https://app.clickup.com/t/8688wzkxm)

Really small PR moving us to a better practice with mypy when defining functions inside a class that return an instance of that class. Previously, there was a hacky way of doing this, but John J. found a better way a few months ago. Just getting around to propagating it.

# Tests Added

N/A


fatemetkl (approved):
LGTM! :)

PR #287, opened 2024-11-13
Small update to FLamby install instructions to avoid cluster failures.

emersodb:
# PR Type
Documentation

# Short Description

Albumentations upgraded to 1.4.21 on October 31st. They now rely on simsimd, which is not compatible with our cluster. So we pin to 1.4.20 to make stuff work.

### DIFF
--- a/research/flamby/README.md
+++ b/research/flamby/README.md
@@ -17,8 +17,10 @@ cd <fl4health_repository>
 pip install --upgrade pip poetry
 poetry install --with "dev, dev-local, test, codestyle"
 cd <FLamby_repository>
+pip install albumentations==1.4.20
 pip install -e ".[cam16, heart, isic2019, ixi, lidc, tcga]"
 ```
+__NOTE__: The forced install of albumentations 1.4.20 avoids an issue where versions > 1.4.20 rely on simsimd which does not compile on the Vector cluster
 __NOTE__: We avoid installing Fed-KITS2019, as it requires a fairly old version on nnUnet, which we no longer support in our library.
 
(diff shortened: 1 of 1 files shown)


research/flamby/README.md
```diff
@@ -17,8 +17,10 @@ cd <fl4health_repository>
 pip install --upgrade pip poetry
 poetry install --with "dev, dev-local, test, codestyle"
 cd <FLamby_repository>
+pip install albumentations==1.4.20
```
lotif, line 20: Shouldn't we add this to poetry's `pyproject.toml` instead?

emersodb: Good question. This is specific to the FLamby environment. We don't use albumentations in our library but FLamby does in theirs. There is an issue using the most recent version of the library though. So this forces a pin while still allowing the rest of FLamby's dependencies to be installed

lotif: I see. Is that the only dependency that is in flamby but not in fl4health?

emersodb: No. Unfortunately there are a lot more in FLamby that are necessary to support the different datasets.

lotif: I see, I was just wondering if there was a requirements file somewhere that you could put this but looking into it here it seems like there isn't, so all good.

lotif (approved):

research/flamby/README.md
emersodb: Your thought process is a good one. We actually used to have a requirements file for installing the requirements of FL4Health and FLamby together. The problem was that it ends up out of date fairly quickly, as it's not tied to our poetry or their repository. If you have better suggestions for maintaining the compatibility as a follow up PR, I'm open to it 🙂

PR #294, opened 2024-11-21
Server storing FL configs and Consolidating Base Server Functionality

emersodb:
# PR Type

Other: QoL improvements

# Short Description

Clickup Ticket(s): [Ticket 1](https://app.clickup.com/t/868avwx9u), [Ticket 2](https://app.clickup.com/t/8689xmgdh)

There are two tickets folded into this PR. They were sort of interconnected, where fixing one would have left some code weirdness lying around that I didn't want.

The first component forces the server to store a configuration file. The goal here is for this configuration file to essentially be the representation of the config used to create the `on_fit_config_fn` and `on_evaluation_config_fn` variables that are used to drive overall FL training and evaluation. The server sort of "forgets" about these in the current setup because they go directly to the strategy and are become inaccessible to the server. 

**NOTE:** There are a lot of file changes in this PR. Almost all of them are due to the fact that I am forcing the config to be provided to the server. So I had to migrate all of our examples etc.

The second component of this PR is merging the functionality of `FlServerWithCheckpointing` and `FlServerWithInitializer` into the `FlServer` class. For a while it made sense for these to be seperate functionalities, but it's actually not all that much more complex to just fold those functionalities into the base class and it makes inheritance much easier.

# Tests Added

A test was added to cover a checkpointing edge case, but, overall, this is a refactor. So the tests we have should be sufficient.


examples/dp_fed_examples/instance_level_dp/server.py
```diff
@@ -61,43 +56,6 @@ def fit_config(
     )
 
 
-class CifarInstanceLevelDPServerWithCheckpointing(InstanceLevelDpServer):
```
emersodb, line 64: All functionality implemented here is now neatly contained in the parent class of `InstanceLevelDpServer`

fl4health/servers/base_server.py
```diff
@@ -353,14 +394,11 @@ def evaluate_round(
                     "eval_round_end": str(end_time),
                     "eval_round_time_elapsed": round((end_time - start_time).total_seconds()),
                 }
-                dummy_params = Parameters([], "None")
```
emersodb, line 356: The stored config means that we no longer have to use the strategy to configure evaluation, which was a work-around for not having access to the config in the server.

fl4health/servers/base_server.py
```diff
@@ -44,30 +50,89 @@ def __init__(
...
+
+        self.current_round: int
+        self.history: History
+
         # Initialize reporters with server name information.
         self.reports_manager = ReportsManager(reporters)
         self.reports_manager.initialize(id=self.server_name)
+        self._log_fl_config()
+
+    def update_before_fit_round(self, server_round: int, timeout: Optional[float] = None) -> None:
```
emersodb, line 122: This function takes the place of the `initialize` functionality that was present in the FlServerWithIntiailizer class. It's called before every fit round starts. In the case of nnunet, it simply does the initialization process once (when server_round == 1) and then does nothing there after.

fl4health/servers/nnunet_server.py
```diff
@@ -138,43 +153,27 @@ def initialize_server_model(self) -> None:
...
-        Same as parent method except initialize hook method is called first
-        """
-        # Initialize the server
-        if not self.initialized:
-            self.initialize(server_round=0, timeout=timeout)
-
-        return FlServerWithCheckpointing.fit(self, num_rounds, timeout)
-
-    def initialize(self, server_round: int, timeout: Optional[float] = None) -> None:
+    def update_before_fit_round(self, server_round: int, timeout: Optional[float] = None) -> None:
```
emersodb, line 156: Rather than overridding an initialize function, we use a more general hook that is called every time the server calls `fit_round`. This function only does anything on the very first server round.

fl4health/servers/nnunet_server.py
```diff
@@ -138,43 +153,24 @@ def initialize_server_model(self) -> None:
...
-        Same as parent method except initialize hook method is called first
-        """
-        # Initialize the server
-        if not self.initialized:
-            self.initialize(server_round=0, timeout=timeout)
-
-        return FlServerWithCheckpointing.fit(self, num_rounds, timeout)
-
-    def initialize(self, server_round: int, timeout: Optional[float] = None) -> None:
+    def update_before_fit(self, num_rounds: int, timeout: Optional[float]) -> None:
```
emersodb, line 156: This function takes the place of the initialize functionality that was present in the FlServerWithIntiailizer class. It's called before any fitting starts (i.e. right at the start of FL).

fl4health/servers/nnunet_server.py
```diff
@@ -187,11 +183,11 @@ def initialize(self, server_round: int, timeout: Optional[float] = None) -> None
                 "Requesting initialization of global nnunet plans from one random client via get_properties",
             )
             random_client = self._client_manager.sample(1)[0]
-            ins = GetPropertiesIns(config=config)
-            properties_res = random_client.get_properties(ins=ins, timeout=timeout, group_id=server_round)
+            ins = GetPropertiesIns(config=self.fl_config | {"current_server_round": 0})
```
emersodb, line 186: The fl_config will have everything needed for this except for what server round we're currently on. So we just add it here. 0 is because no rounds have actually been done.

scarere (commented):
Looks good to me! Mostly had comments on stuff from previous PR's that we're just brought to light in this one and might be worth addressing

fl4health/servers/nnunet_server.py
```diff
@@ -206,20 +202,20 @@ def initialize(self, server_round: int, timeout: Optional[float] = None) -> None
 
             # If config contains nnunet_plans, server side initialization of plans
             # Else client side initialization with nnunet_plans from client
-            if config.get("nnunet_plans") is not None:
-                self.nnunet_plans_bytes = narrow_dict_type(config, "nnunet_plans", bytes)
+            if self.fl_config.get("nnunet_plans") is not None:
```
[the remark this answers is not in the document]
scarere: As an add on to this, I just realized that it seems the pre-init runs everytime regardless. The log statements are also innaccurate, they say we are requesting plans from the client regardless of whether or not this is the case

fl4health/servers/base_server.py
```diff
@@ -84,9 +148,103 @@ def report_centralized_eval(self, history: History, num_rounds: int) -> None:
...
+            if res_fed:
+                loss_fed, evaluate_metrics_fed, _ = res_fed
+                if loss_fed:
+                    self.history.add_loss_distributed(server_round=self.current_round, loss=loss_fed)
+                    self.history.add_metrics_distributed(server_round=self.current_round, metrics=evaluate_metrics_fed)
+
+            self.current_round += 1
+
+            # Save checkpoint after training and testing
+            self._hydrate_model_for_checkpointing()
```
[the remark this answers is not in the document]
emersodb: It's a good question. It's because of the state save part actually. Because I reordered the methods (shouldn't have done that for the diffing 😂), it's a bit hidden, but we had this method previously as well. Basically, we need to intervene to maybe load state at the beginning of a round, save state at the end of the rounds and to keep track of what round we're on etc.

I don't think I'm actually changing anything in this method. I just moved it and renamed the state saving functions to have the _ prefix to indicate an internal call.

fl4health/servers/base_server.py
```diff
@@ -370,249 +409,42 @@ def evaluate_round(
...
-                }
-            )
-        else:
-            # parent method includes call to report metrics to wandb if specified
-            history, elapsed_time = super().fit(num_rounds, timeout)
-
-        return history, elapsed_time
 
-    def fit_with_per_epoch_checkpointing(self, num_rounds: int, timeout: Optional[float]) -> Tuple[History, float]:
+    def _save_server_state(self) -> None:
```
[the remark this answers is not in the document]
emersodb: Yeah, I think we have some tickets that speak to this theme. There are some design things that need to be sorted out, as the two classes are sort of trying to accomplish different things. I'm taking the first steps towards unifying these things into one module. Perhaps the next natural step is to merge them together in a meaningful way.

fl4health/servers/nnunet_server.py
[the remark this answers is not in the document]
emersodb: Discussed together. Going to refactor for a better flow.

fl4health/servers/base_server.py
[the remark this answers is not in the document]
scarere: Ah yes i'm reminded again why this is necessary. Annoying that the whole method needs to overridden to just so that a single line can be changed to resume from a particular round.

fl4health/servers/nnunet_server.py
```diff
@@ -177,22 +177,25 @@ def update_before_fit(self, num_rounds: int, timeout: Optional[float]) -> None:
...
             log(INFO, "")
             log(INFO, "[PRE-INIT]")
             log(INFO, "Requesting properties from one random client via get_properties")
 
-            if self.fl_config.get("nnunet_plans") is None:
+            if not server_nnunet_plans_exist:
                 # If the nnUnet plans are not specified, we also need those plans from the client.
                 log(INFO, "Initialization of global nnunet plans will be sourced from this client")
-            else:
+            if self.per_round_checkpointer is not None:
```
scarere, line 198: The statement below on line 209 is not always True. We're stating that we received the global nnunet plans from the client, when in fact the server may have provided them. Same with the line 211, we may be querying for the properties only for the purpose of the state saving, not for the plans.

fl4health/servers/nnunet_server.py
```diff
@@ -212,7 +215,7 @@ def update_before_fit(self, num_rounds: int, timeout: Optional[float]) -> None:
 
             # If config contains nnunet_plans, server side initialization of plans
             # Else client side initialization with nnunet_plans from client
-            if self.fl_config.get("nnunet_plans") is not None:
+            if server_nnunet_plans_exist:
```
scarere, line 218: The code that wraps the on_fit_config function could go within the scope of this if statement. There's no need to update the config functions with the nnunet_plans if we didn't change the config and the nnunet_plans are already in there

fl4health/servers/nnunet_server.py
```diff
@@ -177,22 +177,25 @@ def update_before_fit(self, num_rounds: int, timeout: Optional[float]) -> None:
...
+
         # If the per_round_checkpointer has been specified and a state checkpoint exists, we load state
         # NOTE: Inherent assumption that if checkpoint exists for server that it also will exist for client.
         if self.per_round_checkpointer is not None and self.per_round_checkpointer.checkpoint_exists():
             self._load_server_state()
         # Otherwise, we're starting training from "scratch"
-        elif self.per_round_checkpointer is not None:
-            # If the state checkpointer is not None, then we want to do state checkpointing. So we need information
-            # from the clients in the form of get_properties.
+        elif self.per_round_checkpointer is not None or not server_nnunet_plans_exist:
```
scarere, line 187: if per round checkpointer is None, does the server model need to be initialized? It will be if server_nnunet_plans_exist is True

emersodb: Yes, that's a good catch. Apologies for missing that.

fl4health/servers/nnunet_server.py
emersodb: We talked about this briefly. We'll keep it where it is for now, since it might save someone who forgets the plans in the configs, but did have them in the `fl_config`.

fl4health/servers/client_level_dp_fed_avg_server.py
```diff
@@ -38,6 +40,10 @@ def __init__(
         Args:
             client_manager (ClientManager): Determines the mechanism by which clients are sampled by the server, if
                 they are to be sampled at all.
+            fl_config (Config): This should be the configuration that was used to setup the federated training.
+                In most cases it should be the "source of truth" for how FL training/evaluation should proceed. For
+                example, the config used to produce the on_fit_config_fn and on_evaluate_config_fn for the strategy.
+                NOTE: This config is DISTINCT from the Flwr server config, which is extremely minimal.
```
jewelltaylor, line 46: Maybe put the type (ServerConfig) as well and/or some info on the distinction between fl_config and "Flwr server config" just so people know its not a substitute and stores complimentary information (if thats the case)

fl4health/servers/adaptive_constraint_servers/mrmtl_server.py
```diff
@@ -12,6 +13,7 @@ class MrMtlServer(FlServer):
     def __init__(
         self,
         client_manager: ClientManager,
+        fl_config: Config,
```
jewelltaylor, line 16: I may be missing something but parent (FlServer) takes in a few other parameters like model? Is there a reason we are omitting those arguments in the subclasses like MrMtl and others

fl4health/servers/client_level_dp_fed_avg_server.py
emersodb: Yup. Definitely a good call

fl4health/servers/scaffold_server.py
```diff
@@ -19,6 +20,7 @@ class ScaffoldServer(FlServer):
     def __init__(
         self,
         client_manager: ClientManager,
+        fl_config: Config,
```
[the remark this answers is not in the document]
emersodb: Yeah, I think this was an oversight for us, as it didn't exist in our implementation before. However, this will be a moot point, as the checkpointing work in [this PR](https://github.com/VectorInstitute/FL4Health/pull/298) adds it properly to this server class.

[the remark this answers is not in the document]
jewelltaylor: Discussed and this is a moot point as its handled in the following PR

fl4health/servers/nnunet_server.py
jewelltaylor: Cool! Didn't know about | operator for dicts!

fl4health/strategies/basic_fedavg.py
```diff
@@ -365,7 +365,7 @@ def __init__(
             on_fit_config_fn (Optional[Callable[[int], Dict[str, Scalar]]], optional):
                 Function used to configure training by providing a configuration dictionary. Defaults to None.
             on_evaluate_config_fn (Optional[Callable[[int], Dict[str, Scalar]]], optional):
-                Function used to configure server-side central validation by providing a Config dictionary.
+                Function used to configure client-side validation by providing a Config dictionary.
```
jewelltaylor, line 368: Good catch

jewelltaylor (approved):
LGTM!

fl4health/servers/base_server.py
```diff
@@ -370,249 +409,42 @@ def evaluate_round(
...
-        # Throw an exception alerting the user to failures on the client-side causing termination
-        self.shutdown()
-        raise ValueError(
-            f"The server encountered failures from the clients and accept_failures is set to {self.accept_failures}"
-        )
+    def _log_fl_config(self) -> None:
+        log(INFO, "FL Configuration:") if self.fl_config else log(INFO, "FL Config is Empty")
+        for config_key, config_value in self.fl_config.items():
+            if not isinstance(config_value, bytes):
+                log(INFO, f"Key: {config_key} Value: {config_value!r}")
```
lotif, line 416: Interesting, I didn't know about `!r`

fl4health/servers/nnunet_server.py
lotif: Just be careful because `{}` will evaluate to `True` in this case. It's fine if that's not a concern.

emersodb: I maybe misreading your comment Marcelo, but in this case, the pipe isn't a binary operator but is rather a merge operator for the two dictionaries. So I think we're okay, but perhaps I'm missing something 🙂

PR #296, opened 2024-11-24
Custom nnunet trainer

jewelltaylor:
# PR Type

# Short Description

Small PR to add the ability to specify a custom nnunet trainer class.

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -8,7 +8,7 @@ from contextlib import redirect_stdout
 from logging import DEBUG, ERROR, INFO, WARNING
 from os.path import exists, join
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 
 import numpy as np
 import torch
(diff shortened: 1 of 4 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -141,6 +143,10 @@ def __init__(
                 provided. Defaults to None.
             reporters (Sequence[BaseReporter], optional): A sequence of FL4Health
                 reporters which the client should send data to.
+            nnunet_trainer_class (Type[nnUNetTrainer]): A nnUNetTrainer constructor.
```
emersodb, line 146: Maybe add a quick note here and on the server-side that the training classes have to match each other (client and server trainers should be the same)?

pyproject.toml
```diff
@@ -34,6 +34,7 @@ scikit-learn = "1.5.0" # Pin as it was causing issues with nnunet
 # Problematic grpcio versions cause issues, should be fixed in next flwr update
 # See https://github.com/adap/flower/pull/3853
 # https://github.com/grpc/grpc/issues/37162
+tornado = ">=6.4.2"
```
emersodb, line 37: Out of curiosity, why did this need upgrading?

emersodb (approved):
Looks good. Just two very minor comments.

pyproject.toml
jewelltaylor: pip audit flags 6.4.1 so had to upgrade to 6.4.2 or higher

emersodb: Sounds good!

PR #298, opened 2024-11-27
Consolidating model and state checkpointing on the client and server sides.

emersodb:
# PR Type
Refactor

# Short Description

Clickup Ticket(s): Work towards this ticket: https://app.clickup.com/t/8689xkbwm, though more work is likely necessary.

This PR is starting the work towards consolidating the model and state checkpointing on both the server and client side. On the server side, we're pulling the hydration responsibilities into the model checkpointing containers. That way the server need not "know" the process required to inject parameters into a model. 

It also means that there aren't several connected but disparate pieces that need to be provided to the servers/clients, you instantiate these models and send them to the server to make things work.

# Tests Added

Migrated existing tests to work with this new system.


fl4health/checkpointing/server_module.py
```diff
@@ -0,0 +1,580 @@
...
+
+CheckpointModuleInput = Union[TorchModuleCheckpointer, Sequence[TorchModuleCheckpointer]] | None
+
+
+class BaseServerCheckpointAndStateModule:
+    def __init__(
+        self,
+        model: nn.Module | None = None,
+        parameter_exchanger: ExchangerType | None = None,
+        model_checkpointers: CheckpointModuleInput = None,
```
scarere, line 30: So it seems the on thing preventing us from merging the model checkpointing and state checkpointing is the load state method. I'm wondering if there is any way around this so that we can fold all these classes and subclasses into one.

fl4health/checkpointing/server_module.py
```diff
@@ -0,0 +1,580 @@
...
+            if self.state_checkpointer.checkpoint_exists(state_checkpoint_name):
+                return self.state_checkpointer.load_checkpoint(state_checkpoint_name)
+            else:
+                log(INFO, "State checkpointer is defined but no state checkpoint exists.")
+                return None
+        else:
+            raise ValueError("Attempting to load state, but no state checkpointer is specified")
+
+
+class PackingServerCheckpointAndAndStateModule(BaseServerCheckpointAndStateModule):
```
scarere, line 204: If the only difference between all these subclasses is the parameter exchanger, why the need for the wrapper classes? Why not just let the user initialize the base class with a parameter exchanger?

fl4health/checkpointing/server_module.py
```diff
@@ -0,0 +1,580 @@
...
+                places. Defaults to None.
+            model_checkpointers (CheckpointModuleInput, optional): If defined, this checkpointer (or sequence of
+                checkpointers) is used to checkpoint models based on their defined scoring function. Defaults to None.
+            state_checkpointer (PerRoundStateCheckpointer | None, optional): If defined, this checkpointer will be
+                used to preserve FL training state to facilitate restarting training if interrupted. Generally, this
+                checkpointer will save much more than just the model being trained. Defaults to None.
+        """
+        if parameter_exchanger is not None:
+            assert isinstance(
+                parameter_exchanger, FullParameterExchangerWithPacking
```
scarere, line 238: Why the need for all the structure and classes with the parameter and packing exchangers? From what I can tell we're trying to allow users to pass additional information when pushing and pulling the parameters. They define a packer, which is passed to the packing exchanger subclass which now is also passed here. There just seems to be a lot of classes involved. Can we just have all the packer classes be subclasses of the parameter exchanger? Is there a way to reduce the number of classes and subclasses across the state checkpointers and module checkpointers?

fl4health/checkpointing/client_module.py
```diff
@@ -34,31 +40,37 @@ def __init__(
             pre_aggregation (CheckpointModuleInput, optional): If defined, this checkpointer (or sequence of
                 checkpointers) is used to checkpoint models based on their validation metrics/losses **BEFORE**
                 server-side aggregation. Defaults to None.
-            post_aggregation (CheckpointModuleInput, optional], optional): If defined, this checkpointer (or sequence
+            post_aggregation (CheckpointModuleInput, optional): If defined, this checkpointer (or sequence
```
scarere, line 43: Is there an instance where the post-aggregation client model is different from the global model?

fl4health/checkpointing/client_module.py
```diff
@@ -1,31 +1,37 @@
...
 
 import torch.nn as nn
 from flwr.common.logger import log
 from flwr.common.typing import Scalar
 
-from fl4health.checkpointing.checkpointer import TorchCheckpointer
+from fl4health.checkpointing.checkpointer import PerRoundStateCheckpointer, TorchModuleCheckpointer
 
-CheckpointModuleInput = Optional[Union[TorchCheckpointer, Sequence[TorchCheckpointer]]]
+CheckpointModuleInput = Union[TorchModuleCheckpointer, Sequence[TorchModuleCheckpointer]] | None
```
scarere, line 11: Is this type needed? Might it be clearer to just have the more explicit typing. If not maybe we can change this type name to something that communicates that there can be multiple checkpointers

fl4health/checkpointing/server_module.py
```diff
@@ -0,0 +1,580 @@
...
+from fl4health.parameter_exchange.parameter_exchanger_base import ExchangerType
+from fl4health.parameter_exchange.parameter_packer import (
+    ParameterPackerAdaptiveConstraint,
+    ParameterPackerWithClippingBit,
+    ParameterPackerWithControlVariates,
+    ParameterPackerWithLayerNames,
+    SparseCooParameterPacker,
+)
+
+CheckpointModuleInput = Union[TorchModuleCheckpointer, Sequence[TorchModuleCheckpointer]] | None
```
scarere, line 22: Same comment here with the typing, is it better to be more explicit by removing this type definition?

scarere (commented):
So as I understand it, we've defined two base types for checkpointing. Module (Torch) checkpointers and state checkpointers. Furthermore we defined client and server checkpoint  module base classes which pack together the module checkpointers (is the term module being reused here?) and the state checkpointers. I think as a long term goal for a future PR it would be great to unify the state and module checkpointers. However I understand that needing to know which checkpointer to use for loading the state complicates that. Although I know this wasn't part of this PR, I was looking to the packing exchangers and parameter exchangers and am a little confused as to why so much structure and subclassing is needed. Additionally with the server checkpointing modules, there seems to be a large number of subclasses where the primary difference is just the argument passed to the parent class. Is it possible to get rid of these?

emersodb:
> So as I understand it, we've defined two base types for checkpointing. Module (Torch) checkpointers and state checkpointers. Furthermore we defined client and server checkpoint module base classes which pack together the module checkpointers (is the term module being reused here?) and the state checkpointers. I think as a long term goal for a future PR it would be great to unify the state and module checkpointers. However I understand that needing to know which checkpointer to use for loading the state complicates that. Although I know this wasn't part of this PR, I was looking to the packing exchangers and parameter exchangers and am a little confused as to why so much structure and subclassing is needed. Additionally with the server checkpointing modules, there seems to be a large number of subclasses where the primary difference is just the argument passed to the parent class. Is it possible to get rid of these?

Shawn and I talked about the final question before the holiday break. The basic gist is that there is some necessary complexity associated with the exchangers but also the opportunity for simplification. We'll take a shot at this in future work. Similarly, the end state that Shawn describes of some merging of checkpointers and state modules etc. is desirable but to be deferred to future PRs

fl4health/checkpointing/client_module.py
emersodb: That's a good question. I'll change this to something that is a bit more explicit. We don't need this type but it's more convenient than writing a really long type 😂. Perhaps I can name it something like ` ModelCheckpointers`?

fl4health/checkpointing/client_module.py
emersodb: This is a good question. I assume by different, you mean has different model weights than the global model. The simplest example is for personalized federated learning. For a FENDA model, there are model weights that are aggregated and weights that are never sent to the server. You might checkpoint the model after local training where the shared weights haven't yet been aggregated by the server or you might save the model immediately after aggregation (where some of the weights are still "local") but before any additional local training begins. That would constitute post-aggregation checkpointing in our definition.

fl4health/checkpointing/server_module.py
emersodb: I think it's worthwhile to settle on a more expressive name. I'll migrate this over to something better.

fl4health/checkpointing/server_module.py
emersodb: Potentially. I think we can defer to a follow up PR so we can think about this unification carefully.

fl4health/checkpointing/server_module.py
emersodb: We talked about this briefly (after the comment). The idea is to try to make things "user proof" such that you can't end up with some misconfigurations etc. However, to do this it sometimes requires subclasses that might be undesirably heavy. We'll try to revisit these designs and think of a more compact way to achieve something similar in future work.

fl4health/checkpointing/checkpointer.py
```diff
@@ -38,13 +38,26 @@ def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: Dict[str, Sca
...
+        checkpointer
 
-    def load_best_checkpoint(self) -> nn.Module:
-        return torch.load(self.best_checkpoint_path)
+        Args:
+            path_to_checkpoint (str | None, optional): If provided, the checkpoint will be loaded from this path.
+                If not specified, the checkpointer will load from self.checkpoint_path. Defaults to None.
+
+        Returns:
+            nn.Module: _description_
```
jewelltaylor, line 53: nit: _description_ placeholder

fl4health/checkpointing/checkpointer.py
```diff
@@ -38,13 +38,26 @@ def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: Dict[str, Sca
         Raises:
             NotImplementedError: Must be implemented by the checkpointer
         """
-        raise NotImplementedError
+        raise NotImplementedError("maybe_checkpoint must be implemented by inheriting classes")
+
+    def load_checkpoint(self, path_to_checkpoint: str | None = None) -> nn.Module:
+        """
+        Checkpointer with the option to either specify a checkpoint path or fall back on the internal path of the
```
jewelltaylor, line 45: Maybe this will become more obvious as I go through the PR, but what is the case when we would pass a path to checkpoint instead of using the internal checkpoint_path? If you think its worth mentioning, perhaps we could include a small note in the documentation here

fl4health/checkpointing/checkpointer.py
emersodb: Good catch

fl4health/checkpointing/checkpointer.py
emersodb: Good question: Right now, this functionality is mostly useful for testing (used in our unit tests). However, a downstream case I could see this being useful is if you have a checkpointer that doesn't overwrite checkpoints (i.e. it rotates the internal path). If you wanted to load a past checkpoint, you'd need to specify it here. I'll add a note along that line here.

fl4health/checkpointing/server_module.py
```diff
@@ -0,0 +1,575 @@
...
+                order to allow FL to restart from where it left off.
+
+        Raises:
+            ValueError: Throws an error if other_state already has a key called 'model'
+            ValueError: Throws an error if this function is called, but no state checkpointer has been provided
+        """
+        if self.state_checkpointer is not None:
+            self._hydrate_model_for_checkpointing(server_parameters)
+            if "model" not in other_state:
+                other_state["model"] = self.model
```
jewelltaylor, line 170: nit: Should we create a new dict as opposed to modifying the dictionary that was passed

emersodb: Yeah I think that's a better idea! Will implement.

fl4health/checkpointing/server_module.py
jewelltaylor: Apologies if this was already discussed but can we make it "user proof" by design by passing the parameter_exchanger attribute from the server when we invoke methods in the ServerCheckpointAndAndStateModule that require packing and unpacking? (Since we know CheckpointAndAndStateModule is always called from a FlServer or subclass) That way we don't have to define a different CheckpointAndAndStateModule for every method that has a different parameter exchanger and only cases when different logic is needed in the packing and unpacking. 

Apologies if I am missing something obvious, I will blame it on brain fog from the holidays

jewelltaylor: This presumes we decouple the parameter exchanger from the CheckpointAndAndStateModule and pass the parameter exchanger as a separate argument as before

emersodb: Talk about this on Zoom. This is sort of what is already being done, but in essence we're isolating the exchanger to the only module that uses it at the moment. However, it's definitely clear that with some thought we can likely make this simpler. Just need to carve out some time to produce the better design. Will defer this to future work.

fl4health/servers/adaptive_constraint_servers/ditto_server.py
```diff
@@ -30,23 +33,42 @@ def __init__(
...
-                server-side checkpointing is performed. Multiple checkpointers can also be passed in a sequence to
-                checkpointer based on multiple criteria. Ensure checkpoint names are different for each checkpoint
-                or they will overwrite on another. Defaults to None.
-            reporters (Sequence[BaseReporter], optional): A sequence of FL4Health
-                reporters which the server should send data to before and after each round.
+            reporters (Sequence[BaseReporter], optional): A sequence of FL4Health reporters which the server should
+                send data to before and after each round. Defaults to None.
+            checkpoint_and_state_module (AdaptiveConstraintServerCheckpointAndStateModule | None, optional): This
+                module is used to handle both model checkpointing and state checkpointing. The former is aimed at
+                saving model artifacts to be used or evaluated after training. The later is used to preserve training
```
jewelltaylor, line 42: nit: The later => The latter

fl4health/servers/base_server.py
```diff
@@ -57,60 +49,38 @@ def __init__(
...
-                server round. Defaults to None.
-            on_init_parameters_config_fn (Optional[Callable[[int], Dict[str, Scalar]]], optional):
-                Function used to configure how one asks a client to provide parameters from which to initialize all
-                other clients by providing a Config dictionary. If this is none, then a blank config is sent with the
-                parameter request (which is default behavior for flower servers). Defaults to None.
-            server_name (Optional[str], optional): An optional string name to uniquely identify server.
-                Defaults to None.
+            checkpoint_and_state_module (BaseServerCheckpointAndStateModule | None, optional): This module is used
+                to handle both model checkpointing and state checkpointing. The former is aimed at saving model
+                artifacts to be used or evaluated after training. The later is used to preserve training state
```
jewelltaylor, line 56: nit: The later => The latter. This is present in each of the server docs for checkpoint_and_state_module

jewelltaylor (approved):
Looks good to me!

PR #303, opened 2024-12-11
Converting to masked layers bug fix

yc7z:
# PR Type
Fix

# Short Description

Fixed the bug for converting a model to the masked version by setting the mask scores parameters at the appropriate module level.

# Tests Added

Added a new test that uses `load_state_dict`.


emersodb (approved):
Changes look good to me...very subtle bug. Good catch!

PR #304, opened 2024-12-16
Add max_num_validation_steps member of config and client and related …

jewelltaylor:
# PR Type

# Short Description

[Add ability to specify maximum number of validation steps to perform](https://app.clickup.com/t/868bcpawz)

Add ability to specify maximum number of validation steps to perform. This functionality is useful in cases where the validation dataset size is large and we don't want to compute metrics over the entire validation set. This is immediately useful for the nnunet experiments because the [nnunet cli](https://github.com/MIC-DKFZ/nnUNet/blob/43349fa5f0680e8109a78dca7215c19e258c9dd7/nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L151) only performs 50 steps on the validation set.

# Tests Added

### DIFF
--- a/fl4health/clients/basic_client.py
+++ b/fl4health/clients/basic_client.py
@@ -126,6 +126,9 @@ class BasicClient(NumPyClient):
         self.num_val_samples: int
         self.num_test_samples: Optional[int] = None
         self.learning_rate: Optional[float] = None
+        # Config can contain max_num_validation_steps key, which determines an upper bound
+        # for the validation steps taken. If not specified, no upper bound will be enforced.
+        self.max_num_validation_steps: int | None = None
 
     def _maybe_checkpoint(self, loss: float, metrics: dict[str, Scalar], checkpoint_mode: CheckpointMode) -> None:
         """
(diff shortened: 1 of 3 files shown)


fl4health/clients/basic_client.py
```diff
@@ -748,7 +753,10 @@ def _validate_or_test(
         metric_manager.clear()
         loss_meter.clear()
         with torch.no_grad():
-            for input, target in maybe_progress_bar(loader, self.progress_bar):
+            for i, (input, target) in enumerate(maybe_progress_bar(loader, self.progress_bar)):
```
scarere, line 756: So this approach works great for the nnunet client where the dataloaders randomly sample from the set. But if we have a traditional dataloader that runs through the val set, then will setting a max number of steps here that doesn't cover the whole validation set mean that we're basically not using part of the validation set? If so do we care? Is it worth trying to add logic that ensures the whole validation set is used?

scarere (approved):

fl4health/clients/basic_client.py
jewelltaylor: I think it's a lever you would only want to turn on if you explicitly don't want to use the entire validation set (ex: for performance reasons). In all other instances, you would not specify max_num_validation_steps and get the expected behaviour of going through entire validation set

jewelltaylor: I will add some sort of log that warns that we are sampling from validation set when max_validation_steps is specified

PR #305, opened 2024-12-18
Nnunet amp integration

jewelltaylor:
# PR Type

# Short Description

Clickup Ticket(s): [Add Mixed Precision Training to NnuentClient](https://app.clickup.com/t/868beh5kf)

By default, if you have a CUDA GPU, nnUNetTrainer uses automatic mixed precision training. This casts down the weights and inputs/activation from fp32 to fp16 for certain sets of operations. Weights are kept in fp32 and downcasted. Gradients are compute in fp16 but accumulated in fp32. In order to realize automatic mixed precision training, we must wrap forward pass and loss computation in the autocast context. Additionally, the loss must be scaled to prevent underflow of the gradients which are computed in fp16. This is accomplished via the torch provided GradScaler.

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -15,6 +15,7 @@ import torch
 from flwr.common.logger import FLOWER_LOGGER, console_handler, log
 from flwr.common.typing import Config, Scalar
 from torch import nn
+from torch.cuda.amp import GradScaler
 from torch.nn.modules.loss import _Loss
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler
@@ -188,6 +189,9 @@ class NnunetClient(BasicClient):
         # Used to redirect stdout to logger
         self.stream2debug = StreamToLogger(FLOWER_LOGGER, DEBUG)
 
+        # Used to scale gradients if using mized precision training (true if device is cuda)
+        self.grad_scaler: GradScaler | None = GradScaler() if self.device.type == "cuda" else None
+
         # nnunet specific attributes to be initialized in setup_client
         self.nnunet_trainer_class = nnunet_trainer_class
         self.nnunet_trainer_class_kwargs = nnunet_trainer_class_kwargs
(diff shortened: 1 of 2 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -188,6 +189,9 @@ def __init__(
         # Used to redirect stdout to logger
         self.stream2debug = StreamToLogger(FLOWER_LOGGER, DEBUG)
 
+        # Used to scale gradients if using mized precision training (true if device is cuda)
```
scarere, line 192: nit: mixed

fl4health/clients/nnunet_client.py
```diff
@@ -203,6 +207,58 @@ def __init__(
                 log(INFO, "Switching pytorch model jit compile to OFF")
             os.environ["nnUNet_compile"] = str("false")
 
+    def train_step(self, input: TorchInputType, target: TorchTargetType) -> Tuple[TrainingLosses, TorchPredType]:
```
scarere, line 210: I'm not super familiar with mixed precision training so I'm going to trust that this does what its supposed to do.

fl4health/clients/nnunet_client.py
```diff
@@ -15,6 +15,7 @@
 from flwr.common.logger import FLOWER_LOGGER, console_handler, log
 from flwr.common.typing import Config, Scalar
 from torch import nn
+from torch.cuda.amp import GradScaler
```
scarere, line 18: Curious on your thoughts on whether mixed precision training should be specific to nnunet, or whether its a feature that should be extended to Basic Client (or maybe it warrants a new client subclass and the nnunet client either subclasses or mixes in the new subclass).

fl4health/clients/nnunet_client.py
scarere: Should we add a specific argument here rather than making mixed precision the default for cuda. Is there ever any reason not to use mixed precision?

fl4health/clients/nnunet_client.py
```diff
@@ -525,7 +582,14 @@ def predict(self, input: TorchInputType) -> Tuple[TorchPredType, Dict[str, torch
             unused by this subclass and therefore is always an empty dict
         """
         if isinstance(input, torch.Tensor):
-            output = self.model(input)
+            # If device type is cuda, nnUNet defaults to mixed precision forward pass
+            # https://github.com/MIC-DKFZ/nnUNet/blob/43349fa5f0680e8109a78dca7215c19e258c9dd7 \
+            # nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L993
+            if self.device.type == "cuda":
+                with torch.autocast(self.device.type, enabled=True):
```
scarere, line 589: This is cool, never seen it used before. I thought mixed precision training would be more involved

fl4health/clients/nnunet_client.py
```diff
@@ -577,7 +642,16 @@ def compute_loss_and_additional_losses(
                 f"Got {len(loss_preds)} predictions and {len(loss_targets)} targets."
             )
 
-        return self.criterion(loss_preds, loss_targets), None
+        # If device type is cuda, nnUNet defaults to compute loss in mixed precision
+        # https://github.com/MIC-DKFZ/nnUNet/blob/43349fa5f0680e8109a78dca7215c19e258c9dd7 \
+        # nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L993
+        if self.device.type == "cuda":
+            with torch.autocast(self.device.type, enabled=True):
+                loss = self.criterion(loss_preds, loss_targets), None
```
scarere, line 650: Out of curiosity, does autocast automatically cast variables back to the default precision when the scope is exited? For example are the loss values here going to remain fp16 or whatever they were cast to?

scarere (approved):
PR looks good to me. Most of my comments are just me asking questions about mixed precision training out of interest

fl4health/clients/nnunet_client.py
jewelltaylor: Yeah I think you are right!

fl4health/clients/nnunet_client.py
jewelltaylor: I will add it in the backlog as a potential addition that the user can optionally specify

PR #308, opened 2025-01-06
Upgrading peotry lock to include jinja update. 

emersodb:
# PR Type
Fix

# Short Description
Updating the poetry lock file. This includes moving to Jinja 3.1.5, which is to fix two vulnerabilities in that library

# Tests Added

N/A


pyproject.toml
```diff
@@ -29,6 +29,7 @@ grpcio = "^1.60.0,!=1.64.2,!=1.65.1,!=1.65.2,!=1.65.4"
 monai = "^1.3.0"
 nnunetv2 = "^2.3.1"
 wandb = "^0.18.0"
+acvl_utils = "0.2" # Pin as it was causing an issue with nnunet (ModuleNotFoundError: No module named 'blosc2')
```
emersodb, line 32: Was causing an issue with nnunet. See here: https://discourse.slicer.org/t/totalsegmentator-failed-to-compute-results-returned-non-zero-exit-status-1/38142/7 for others who were also having this issue.

scarere: I've noticed something similar on my end as well. I've been resolving it just by manually pip installing blosc2. Is it better to pin acvl_utils or should blosc2 be added as a dependency?

emersodb: mmm...that's a good question. Has everything been working alright for you with forcing a blosc2 install? If so, that might be a better option.

emersodb: Also, have you pinned the `blosc2` library to anything specific, or is it just a free dependency for now?

lotif: +1 to lock `blosc2` if that works

scarere: I didn't pin it to a version, just free installed it. I currently have version 3.0.0b4 and everything ran fine for my application. Can't confirm whether or not it will cause issues elsewhere

emersodb: Alright. So I tried this out. Unfortunately, `blosc2` 3.0.0 requires python 3.11 (we're 3.10 at the moment). So I have to install 2.7.1, which doesn't fix the issue that we're seeing with nnunet and the smoke tests. We'll eventually upgrade to python 3.11, but it may or may not be a big change. So I'm going to revert back to pinning `acvl_utils` for now and we'll fix it the right way when we move up to python 3.11

fl4health/parameter_exchange/parameter_packer.py
```diff
@@ -41,8 +41,8 @@ def unpack_parameters(self, packed_parameters: NDArrays) -> Tuple[NDArrays, floa
         # The last entry in the parameters list is assumed to be a clipping bound (even if we're evaluating)
         split_size = len(packed_parameters) - 1
         model_parameters = packed_parameters[:split_size]
-        clipping_bound = float(packed_parameters[split_size:][0])
-        return model_parameters, clipping_bound
+        clipping_bound = packed_parameters[split_size:][0]
+        return model_parameters, clipping_bound.item()
```
emersodb, line 45: This needed to be updated as it was throwing a deprecation error that was killing the smoke tests.

pyproject.toml
jewelltaylor: My acvl_utils resolved to 0.2 so I haven't ran into this yet. Seems like the right call is just to pin acvl_utils for now

jewelltaylor (approved):
LGTM!

lotif (approved):

PR #309, opened 2025-01-06
Rxrx1 research running scripts

sanaAyrml:
# PR Type
[Other]

# Short Description

Add running script for rxrx1 dataset. This dataset has natural feature heterogeneity among 4 clients.

# Tests Added

NA


research/rxrx1/fedavg/client.py
```diff
@@ -0,0 +1,174 @@
...
+        test_loader, _ = load_rxrx1_test_data(
+            data_path=self.data_path, client_num=self.client_number, batch_size=batch_size
+        )
+
+        return test_loader
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return torch.nn.CrossEntropyLoss()
+
+    def get_optimizer(self, config: Config) -> Optimizer:
```
emersodb, line 82: Here, we're using the AdamW optimizer, but elsewhere we're just using SGD. Do we want to be consistent?

research/rxrx1/fedavg/run_hp_sweep.sh
```diff
@@ -0,0 +1,57 @@
...
+# Notes:
+# 1) The bash command above should be run from the top level directory of the repository.
+###############################################
+
+SERVER_CONFIG_PATH=$1
+ARTIFACT_DIR=$2
+DATASET_DIR=$3
+VENV_PATH=$4
+
+LR_VALUES=( 0.00001 0.0001 0.001 0.01 0.1 )
```
emersodb, line 28: Not an issue per se, just a suggestion that if we see that a certain range of LR values are performing poorly consistently (i.e. we always get bad models for big LRs of 0.01 to 0.1), we can leave them off sweeps for other experiments. I've seen models essentially be trash when using 1e-1 LR. So I've stopped including it in sweeps 😂

research/rxrx1/evaluate_on_test.py
```diff
@@ -0,0 +1,728 @@
...
+    dataset_dir: str,
+    eval_write_path: str,
+    eval_best_pre_aggregation_local_models: bool,
+    eval_last_pre_aggregation_local_models: bool,
+    eval_best_post_aggregation_local_models: bool,
+    eval_last_post_aggregation_local_models: bool,
+    eval_best_global_model: bool,
+    eval_last_global_model: bool,
+    eval_over_aggregated_test_data: bool,
+    is_apfl: bool,
```
emersodb, line 40: Since we're not going to be experimenting with APFL, I think we can drop this guy?

research/rxrx1/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    return mean, std
+
+
+def write_measurement_results(eval_write_path: str, results: dict[str, float]) -> None:
+    with open(eval_write_path, "w") as f:
+        for key, metric_value in results.items():
+            f.write(f"{key}: {metric_value}\n")
+
+
+def evaluate_rxrx1_model(
```
emersodb, line 41: Since we're not planning to evaluate an apfl model, we can probably drop this option

research/rxrx1/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def load_eval_last_post_aggregation_local_model(run_folder_dir: str, client_number: int) -> nn.Module:
+    model_checkpoint_path = os.path.join(run_folder_dir, f"post_aggregation_client_{client_number}_last_model.pkl")
+    model = torch.load(model_checkpoint_path)
+    return model
+
+
+def evaluate_model_on_dataset(
+    model: nn.Module, dataset: DataLoader, metrics: Sequence[Metric], device: torch.device, is_apfl: bool
```
emersodb, line 78: Since we're not planning to evaluate apfl, we can probably drop this option.

research/rxrx1/evaluate_on_test.py
```diff
@@ -0,0 +1,728 @@
...
+        required=True,
+    )
+    parser.add_argument(
+        "--dataset_dir",
+        action="store",
+        type=str,
+        help="Path to the preprocessed Cifar10 Dataset (ex. path/to/cifar10)",
+        required=True,
+    )
+    parser.add_argument(
```
emersodb, line 633: I don't think this arg is used anywhere?

research/rxrx1/evaluate_on_test.py
```diff
@@ -0,0 +1,728 @@
...
+        action="store",
+        type=str,
+        help="Path to saved model artifacts to be evaluated",
+        required=True,
+    )
+    parser.add_argument(
+        "--dataset_dir",
+        action="store",
+        type=str,
+        help="Path to the preprocessed Cifar10 Dataset (ex. path/to/cifar10)",
```
emersodb, line 630: This helper function should probably be updated :-)

research/rxrx1/data/dataset.py
```diff
@@ -0,0 +1,66 @@
...
+class Rxrx1Dataset(Dataset):
+    def __init__(
+        self,
+        metadata: pd.DataFrame,
+        root: Path,
+        dataset_type: str,
+        transform: Callable | None = None,
+        cache_images: bool = False,
+    ):
+        """
```
emersodb, line 22: As part of this descriptor, maybe we can have a small blurb about how one retrieves this dataset? That is, a url from which to download it (or however else it should be obtained) along with any processing steps that need to happen?

emersodb (commented):
Overall looks nearly ready to go. Most of my comments are minor. Only question of import that we want to consider is the optimizer setup (and consistency for FedAvg vs. the other approaches).

research/rxrx1/data/data_utils.py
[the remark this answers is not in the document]
sanaAyrml: I basically looked at other research scripts and found that picai has a data folder, so I followed the same practice. But I agree that we can move it into `fl4health/datasets` directory.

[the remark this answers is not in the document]
emersodb: Sounds good!

research/rxrx1/data/dataset.py
```diff
@@ -0,0 +1,66 @@
...
+            metadata (DataFrame): A DataFrame containing image metadata.
+            root (str): Root directory for the image data.
+            dataset_type (str): 'train' or 'test' to specify dataset type.
+            transform (callable, optional): Optional transformation to apply to the images.
+        """
+        self.metadata = metadata[metadata["dataset"] == dataset_type]
+        self.root = root
+        self.transform = transform if transform else None
+
+        self.label_map = {label: idx for idx, label in enumerate(sorted(self.metadata["sirna_id"].unique()))}
```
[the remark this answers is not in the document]
sanaAyrml: Why do we want to have this reverse map? original labels are also saved in metadata, so we can retrieve from there, I assume.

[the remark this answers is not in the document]
emersodb: It's just a suggestions to make this line from the utils file simpler and more efficient
```
original_label = next((k for k, v in label_map.items() if v == label), "Unknown")
```
if we have a reverse map, then this can be done with a get rather than an iterator. It doesn't seem like we use the label_map except once. So I thought the reverse map would be more useful.

research/rxrx1/ditto/client.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+        return test_loader
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return torch.nn.CrossEntropyLoss()
+
+    def get_optimizer(self, config: Config) -> dict[str, Optimizer]:
+        # Following the implementation in pFL-Bench : A Comprehensive Benchmark for Personalized
+        # Federated Learning (https://arxiv.org/pdf/2405.17724) for cifar10 dataset we use SGD optimizer
```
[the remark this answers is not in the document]
sanaAyrml: It is not clear from their paper, which optimizer they used. I checked their code as well however the implementation for this dataset was not complete.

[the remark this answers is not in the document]
emersodb: lol...of course not 😂. Maybe quickly check, but if you can't verify, let's just pick one (I think you mentioned AdamW, which is good)

research/rxrx1/fedavg/run_hp_sweep.sh
sanaAyrml: I agree, mostly 0.001 or 0.00001 is the one that working,

research/rxrx1/data/dataset.py
sanaAyrml: Sure I will add a README to dataset section

research/rxrx1/evaluate_on_test.py
sanaAyrml: Deleted it

research/rxrx1/data/dataset.py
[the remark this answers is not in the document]
sanaAyrml: Oh I see your point now. sure I can change it.

research/rxrx1/central/train.py
```diff
@@ -0,0 +1,113 @@
...
+from torch.utils.data import DataLoader
+from torchvision import models
+
+from fl4health.datasets.rxrx1.load_data import load_rxrx1_data
+from fl4health.utils.dataset import TensorDataset
+from fl4health.utils.metrics import Accuracy, MetricManager
+from research.rxrx1.single_node_trainer import SingleNodeTrainer
+
+NUM_CLIENTS = 4
+EPOCHS = 100
```
emersodb, line 18: One other thing we might want to consider is that we're not doing "way more" training steps for central vs. FL. That is, we can approximately tally up the number of batches processed during an entire FL run (summed across clients) and make sure the number of epochs here gets us to that "ballpark" number of steps.

research/rxrx1/central/train.py
```diff
@@ -0,0 +1,113 @@
...
+        action="store",
+        type=str,
+        help="Path to save artifacts such as logs and model checkpoints",
+        required=True,
+    )
+    parser.add_argument(
+        "--dataset_dir",
+        action="store",
+        type=str,
+        help="Path to the preprocessed FedIsic2019 Dataset (ex. path/to/fedisic2019)",
```
emersodb, line 83: Probably update to rxrx1 🙂

emersodb (commented):
Changes all look pretty good. Just a few follow on comments and maybe some speculation on the OOM cause (certainly not definitive though)

fl4health/datasets/rxrx1/load_data.py
```diff
@@ -0,0 +1,171 @@
...
+
+    dataset, _ = construct_rxrx1_tensor_dataset(data, data_path, client_num, "train")
+
+    train_indices, val_indices = create_splits(dataset, seed=seed, train_fraction=train_val_split)
+    train_set = copy.deepcopy(dataset)
+    train_set.data = train_set.data[train_indices]
+    assert train_set.targets is not None
+    train_set.targets = train_set.targets[train_indices]
+
+    validation_set = copy.deepcopy(dataset)
```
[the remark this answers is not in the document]
sanaAyrml: I added this here because I kinda remember previously we figured out if we don't do deep copy change on train set would effect original dataset and the output would not be as expected. Please correct me if I'm wrong.

fl4health/datasets/rxrx1/load_data.py
```diff
@@ -0,0 +1,171 @@
...
+    num_workers: int = 0,
+) -> tuple[DataLoader, DataLoader, dict[str, int]]:
+
+    # Read the CSV file
+    data = pd.read_csv(f"{data_path}/clients/meta_data_{client_num+1}.csv")
+
+    dataset, _ = construct_rxrx1_tensor_dataset(data, data_path, client_num, "train")
+
+    train_indices, val_indices = create_splits(dataset, seed=seed, train_fraction=train_val_split)
+    train_set = copy.deepcopy(dataset)
```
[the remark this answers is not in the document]
sanaAyrml: I think we talked to pass this optimization, so I am resolving this comment,

fl4health/datasets/rxrx1/preprocess.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns:
+        torch.Tensor: The loaded image tensor.
+    """
+    experiment = row["experiment"]
+    plate = row["plate"]
+    well = row["well"]
+    site = row["site"]
+
+    images = []
+    for channel in range(1, 4):
```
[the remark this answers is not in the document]
sanaAyrml: I agree!

research/rxrx1/central/train.py
```diff
@@ -0,0 +1,113 @@
+import argparse
```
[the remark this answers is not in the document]
sanaAyrml: Ignoring this commit as issue is already resolved.

research/rxrx1/central/train.py
sanaAyrml: Yeah, I will take this into account in designing FL set up.

emersodb (approved):
Good to go!

PR #312, opened 2025-01-13
Keras vulnerability ignore until there is a fix

emersodb:
# PR Type
Fix

# Short Description

Keras vulnerability ignore until there is a fix

# Tests Added

NA

### DIFF
--- a/.github/workflows/static_code_checks.yaml
+++ b/.github/workflows/static_code_checks.yaml
@@ -43,7 +43,9 @@ jobs:
           virtual-environment: .venv/
           # Ignoring vulnerability in cryptography
           # Fix is 43.0.1 but flwr 1.9 depends on < 43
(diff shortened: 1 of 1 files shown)


nerdai (approved):
Lacking context for sure on this. I think this is okay for now? I guess alternatives are to rip out Kera's from the deps, and do delay imports if necessary (throwing an error message if needed)?

Not sure how serious these vulnerabilities are.

emersodb:
> Lacking context for sure on this. I think this is okay for now? I guess alternatives are to rip out Kera's from the deps, and do delay imports if necessary (throwing an error message if needed)?
> 
> Not sure how serious these vulnerabilities are.

This is definitely okay for now. We don't heavily use Keras and it's unlikely to cause any issues.

PR #314, opened 2025-01-14
[CU-868byqhk5] Fix smoke tests for GPU

### LINKED ISSUE #217: Smoke Tests Fail due to tolerances on my machine
For some reason smoke tests fail locally due to results being outside tolerances. Tried starting from a fresh environment and the issue persists. Not urgent but worth looking into eventually. Creating an issue so that I don't forget about it.

### System Info
13th Gen Intel® Core™ i9-13900K × 32
Ubuntu 22.04.4 LTS
NVIDIA GeForce RTX 4090 
Driver Version: 555.42.06
Cuda Version: 12.1
Python Version: 3.9.19

### Steps to recreate
```bash
python3.9 -m venv test_env
source test_env/bin/activate
cd ../../FL4Health
git checkout main
git pull
pip install --upgrade pip poetry
poetry install --with "codestyle, dev, dev-local, test, picai"
python -m tests.smoke_tests.run_smoke_test
```

### Output
```bash
$ python -m tests.smoke_tests.run_smoke_test
2024-08-08 15:59:27 INFO     Running smoke tests with parameters:
2024-08-08 15:59:27 INFO     	Server : examples.fedprox_example.server
2024-08-08 15:59:27 INFO     	Client : examples.fedprox_example.client
2024-08-08 15:59:27 INFO     	Config : tests/smoke_tests/fedprox_config.yaml
2024-08-08 15:59:27 INFO     	Dataset: examples/datasets/mnist_data/
2024-08-08 15:59:27 INFO     Preloading MNIST dataset...
2024-08-08 15:59:27 INFO     Preloading MNIST dataset...
INFO :      Data directory: examples/datasets/mnist_data
2024-08-08 15:59:27 INFO     Data directory: examples/datasets/mnist_data
Downloading http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
Failed to download (trying next):
HTTP Error 403: Forbidden

Downloading https://ossci-datasets.s3.amazonaws.com/mnist/train-images-idx3-ubyte.gz
Downloading https://ossci-datasets.s3.amazonaws.com/mnist/train-images-idx3-ubyte.gz to examples/datasets/mnist_data/MNIST/raw/train-images-idx3-ubyte.gz
100%|████████████████████████████████████████████████████| 9912422/9912422 [00:00<00:00, 22336309.08it/s]
Extracting examples/datasets/mnist_data/MNIST/raw/train-images-idx3-ubyte.gz to examples/datasets/mnist_data/MNIST/raw

Downloading http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz
Failed to download (trying next):
HTTP Error 403: Forbidden

Downloading https://ossci-datasets.s3.amazonaws.com/mnist/train-labels-idx1-ubyte.gz
Downloading https://ossci-datasets.s3.amazonaws.com/mnist/train-labels-idx1-ubyte.gz to examples/datasets/mnist_data/MNIST/raw/train-labels-idx1-ubyte.gz
100%|█████████████████████████████████████████████████████████| 28881/28881 [00:00<00:00, 1481673.44it/s]
Extracting examples/datasets/mnist_data/MNIST/raw/train-labels-idx1-ubyte.gz to examples/datasets/mnist_data/MNIST/raw

Downloading http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz
Failed to download (trying next):
HTTP Error 403: Forbidden

Downloading https://ossci-datasets.s3.amazonaws.com/mnist/t10k-images-idx3-ubyte.gz
Downloading https://ossci-datasets.s3.amazonaws.com/mnist/t10k-images-idx3-ubyte.gz to examples/datasets/mnist_data/MNIST/raw/t10k-images-idx3-ubyte.gz
100%|████████████████████████████████████████████████████| 1648877/1648877 [00:00<00:00, 13375956.93it/s]
Extracting examples/datasets/mnist_data/MNIST/raw/t10k-images-idx3-ubyte.gz to examples/datasets/mnist_data/MNIST/raw

Downloading http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
Failed to download (trying next):
HTTP Error 403: Forbidden

Downloading https://ossci-datasets.s3.amazonaws.com/mnist/t10k-labels-idx1-ubyte.gz
Downloading https://ossci-datasets.s3.amazonaws.com/mnist/t10k-labels-idx1-ubyte.gz to examples/datasets/mnist_data/MNIST/raw/t10k-labels-idx1-ubyte.gz
100%|██████████████████████████████████████████████████████████| 4542/4542 [00:00<00:00, 12683441.26it/s]
Extracting examples/datasets/mnist_data/MNIST/raw/t10k-labels-idx1-ubyte.gz to examples/datasets/mnist_data/MNIST/raw

2024-08-08 15:59:33 INFO     Finished preloading MNIST dataset
2024-08-08 15:59:33 INFO     Starting server...
2024-08-08 15:59:36 INFO     Server started
2024-08-08 15:59:36 INFO     Starting client 0
2024-08-08 15:59:36 INFO     Starting client 1
2024-08-08 15:59:36 INFO     Collecting output for Client 0...
2024-08-08 15:59:46 INFO     Output collected for Client 0
2024-08-08 15:59:46 INFO     Collecting output for Client 1...
2024-08-08 15:59:46 INFO     Output collected for Client 1
2024-08-08 15:59:46 INFO     All clients finished execution
2024-08-08 15:59:46 INFO     Collecting output for Server...
2024-08-08 15:59:49 INFO     Output collected for Server
2024-08-08 15:59:49 INFO     Server has finished execution
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/shawn/Code/FL4Health/tests/smoke_tests/run_smoke_test.py", line 462, in <module>
    loop.run_until_complete(
  File "/usr/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/shawn/Code/FL4Health/tests/smoke_tests/run_smoke_test.py", line 260, in run_smoke_test
    assert len(server_errors) == 0, f"Server metrics check failed. Errors: {server_errors}"
AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - prediction - accuracy' (0.4656666666666667) does not match the requested value (0.4627) within requested tolerance (0.0005).", "Saved value for metric 'loss_aggregated' (2.022878646850586) does not match the requested value (2.0268) within requested tolerance (0.0005).", "Saved value for metric 'val - prediction - accuracy' (0.43377777777777776) does not match the requested value (0.4323) within requested tolerance (0.0005).", "Saved value for metric 'loss_aggregated' (1.7308796644210815) does not match the requested value (1.7356) within requested tolerance (0.0005).", "Saved value for metric 'val - prediction - accuracy' (0.5417777777777778) does not match the requested value (0.5362) within requested tolerance (0.0005).", "Saved value for metric 'loss_aggregated' (1.4460455179214478) does not match the requested value (1.4518) within requested tolerance (0.0005)."]
```

nerdai:
# PR Type
[Fixes #217 | Smoke Test Re-Factor]

# Short Description

This PR started out of a need to fix smoke tests whenever running them locally with a GPU (cuda-enabled device). In addition to that, some other (minor) improvements have been made and are listed below:

- Re-factor smoke-tests to use `pytest-asyncio`. This allows us to test each smoke test individually (and leverage pytest features)
- Expose `tolerance` param in `run_smoke_test` in order to specify a tolerance level to use in assertions. Note, that if a `custom_tolerance` field is supplied in the `xxx_metrics.json` then this takes precedence over the tolerance param supplied in `run_smoke_test`.
    - In the smoke tests, the tolerance level is a fixture that is determined on whether a gpu/cuda-enabled device is available.
- Minor improvements to logic for `run_smoke_test` where async tasks were being run in a for loop and thus sequentially blocking one another. Using `asyncio.gather()` instead should allow for better cooperation between these tasks, and is the prevailing choice in these cases.

NOTE: skipping `test_apfl` and `test_feddg_ga` on local tests for now since these are not passing but we understand these as not to be implementation errors but rather due to random fluctuations.

Clickup Ticket(s): Ticket was linked automatically.

# Tests Added
N/A

N/A (though I have some thoughts around these smoke tests that we can discuss when time is right to do so.)


vector-admin:
Task linked: [CU-868byqhk5 [Fix] Smoke tests discrepancy when running on GPUs](https://app.clickup.com/t/868byqhk5)

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -0,0 +1,296 @@
...
+    format="%(asctime)s %(levelname)-8s %(message)s",
+    level=logging.INFO,
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
+logger = logging.getLogger()
+
+
+@pytest.mark.smoketest
+@pytest.mark.asyncio()
+async def test_basic_server_client_cifar(test_checkpoint_dirname: str) -> None:
```
nerdai, line 17: Using tempdir `test_checkpoint_dirname` still fails when running on local. I run into assertion error with server metrics. It makes me wonder if some state is being pulled from "./" by default. 🤔

nerdai:
```
FAILED tests/smoke_tests/test_smoke_tests.py::test_basic_server_client_cifar - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - prediction - accuracy' (0.0983) does not match the requested value (0.0942) within requested tolerance (0.0005).",...
```

nerdai: I resolved this by setting the DEFAULT_TOLERANCE level to something very conservative, namely `0.1`. @emersodb let me know if you think this is too conservative, but this should solve Shawn's issue as well 🤞🏼

emersodb: I don't think this particular smoke test should have significant local fluctuations if I recall correctly. Does this only happen when you move to using the temporary directory? If so then I think we need to look into this more.

A default tolerance of 0.1 is a bit too high for me. For things like accuracy that is a really large fluctuation.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -209,25 +214,29 @@ async def run_smoke_test(
         if seed is not None:
             client_args.extend(["--seed", str(seed)])
 
-        client_process = await asyncio.create_subprocess_exec(
-            "python",
-            *client_args,
-            stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.STDOUT,
+        task = asyncio.create_task(
+            asyncio.create_subprocess_exec(
```
emersodb, line 218: This is potentially a dumb question but if these run in parallel are we going to run into communication clashes for each of the tests?

nerdai: If in there are multiple clients running on same port, then yea you will run into this issue. This wouldn't be an issue for tests, tho because the loop is scope per test.

So if we have multiple clients spun up in a smoke test, then we should just need to set them to different ports: maybe `f"808{i}"`?

nerdai: I didn't see all but I thought most of these smoke tests were one server / one client? Or I probably misunderstood something.

emersodb: I think most will have at least 2 clients. You can see the number in the `n_clients` variable in the configs

tests/smoke_tests/test_smoke_tests.py
nerdai: Yeah, I figured it might be too high. This only happens when I change the directory to something else other than `"./"`, which probably is something we should address now in this PR.

Just not sure if I know enough to even begin to look where this might be coming from...

Would you mind trying to replicate the tol issue when setting checkpoint dir to something else: even something hardcoded `"/.checkpoints"` brings this error for me.

tests/smoke_tests/run_smoke_test.py
nerdai: Does flwr have some automatic logic (similar to jupyter notebooks) of launching on another port if its already being used by another process? Just trying to figure out why these tests would still pass.

nerdai: But yea if you're asking if there is communication problems across the test, there shouldn't be. As these tests are not technically ran in parallel, but sequentially. If we did want to run them in parallel allocating blocks of tests to various threads, then it should be fine as well since each thread would have its own asyncio event loop plus the tests would still run sequentially within each thread

(But we're not here yet, but maybe one day)

tests/smoke_tests/test_smoke_tests.py
emersodb: Yeah let me look into it for a bit and see if I can understand what's happening.

nerdai:
```
================================================================================================== short test summary info ===================================================================================================
FAILED tests/smoke_tests/test_smoke_tests.py::test_scaffold - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - prediction - accuracy' (0.4141111111111111) does not match the requested value (0.4062) within requested tolerance (0.005)."]
FAILED tests/smoke_tests/test_smoke_tests.py::test_apfl - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - loss - aggregated' (1.5720984935760498) does not match the requested value (1.5862) within requested tolerance (0.005).", "Saved val...
FAILED tests/smoke_tests/test_smoke_tests.py::test_feddg_ga - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - loss - aggregated' (1.579448938369751) does not match the requested value (1.5862) within requested tolerance (0.005).", "Saved valu...
============================================================================ 3 failed, 21 passed, 326 deselected, 1 warning in 4015.66s (1:06:55) ============================================================================
(fl4-health) vafajardo@gpu079:~/projects/FL4Health$ 
```

These are the only tests that fail now for me on local and on our cluster (consistency... at least). The two `apfl` and `feddg_ga` are the ones you mentioned before, but I think you had `scaffold` pass locally?

In any case, the GPU/device check and tol level worked – the above snippet is from cluster where cuda is available and so tolerance is 0.005.

fl4health/clients/nnunet_client.py
```diff
@@ -262,7 +262,14 @@ def get_data_loaders(self, config: Config) -> tuple[DataLoader, DataLoader]:
             # NOTE: The type: ignore here is to skip issues where a local operating system is not compatible
             # with sched_getaffinity (older versions of MacOS, for example). The code still won't run but mypy won't
             # complain. Workarounds like using os.cpu_count(), while not exactly the same, are possible.
-            self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)  # type: ignore
+            try:
+                self.n_dataload_proc = min(12, len(os.sched_getaffinity(0)) - 1)  # type: ignore
+            except AttributeError:
```
emersodb, line 267: I think you've already confirmed this, but wanted to make sure that AttributeError is the right exception to catch when the os doesn't have `sched_getaffinity` for the specific underlying OS?

tests/smoke_tests/conftest.py
```diff
@@ -0,0 +1,20 @@
+import tempfile
+from typing import Iterator
+
+import pytest
+import torch
+
+
+@pytest.fixture(scope="session")
+def test_checkpoint_dirname() -> Iterator[str]:
```
emersodb, line 9: Is doing this different than using the `tmp_path: Path` argument that's built into pytest? We've used it before (for example in `test_save_and_load_best_loss_checkpoint`) when we wanted to a temporary directory to save and load stuff from in during the scope of the test function. It's definitely possible that that isn't usable here and therefore we need to define our own. Just curious.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -20,6 +21,12 @@
 )
 logger = logging.getLogger()
 
+DEFAULT_TOLERANCE = 0.0005
```
emersodb, line 24: I feel like I'm missing something, but given your definition of a fixture associated with tolerance, should we be doing away with this default tolerance value and forcing any smoke test to consume the fixture variable?

tests/smoke_tests/run_smoke_test.py
```diff
@@ -20,6 +21,12 @@
...
+
+
+def postprocess_logs(logs: str) -> str:
```
emersodb, line 27: Maybe we can document this for posterity and anyone reading this who might be confused? Something like 
```
This function removes a spurious 'error' that is sometimes thrown when running smoke tests on certain machines from the tcp_posix.cc library. It has been heavily investigated and doesn't not appear to actually affect FL processes.
```

tests/smoke_tests/run_smoke_test.py
```diff
@@ -260,11 +273,12 @@ async def run_smoke_test(
...
 
-    server_errors = _assert_metrics(MetricType.SERVER, server_metrics)
+    server_errors = _assert_metrics(MetricType.SERVER, server_metrics, tolerance)
     assert len(server_errors) == 0, f"Server metrics check failed. Errors: {server_errors}"
 
     # client assertions
     client_errors = []
+    full_client_outputs = [postprocess_logs(el) for el in full_client_outputs]
     for i in range(len(full_client_outputs)):
         assert "error" not in full_client_outputs[i].lower(), (
```
emersodb, line 283: Very minor (and this isn't something you changed or missed), but we access `full_client_outputs[i]` multiple times in this loop. Probably cleaner to access it once, save it, and then use that variable in the later access steps.

emersodb (commented):
Just a couple of small comments 🙂

tests/smoke_tests/conftest.py
nerdai: Yeah, I tried using that pytest `tmp_path` before resorting to `tempfile` builtin but it wasn't working out for me. Perhaps I was battling something else at the time though. I was unsure of what processes were writing to tmp dir and if these processes had access. 🤷‍♂️

Going to try again quickly with `tmp_path`

nerdai: Looks like it's good to go. Removing the custom fixture and using `tmp_path` built in 👍🏼

tests/smoke_tests/run_smoke_test.py
nerdai: This is the fallback option. If you wanted to invoke `run_smoke_test` not in a pytest testing environment. Say for example if we had `main` still defined, or if someone just wanted to spin one up for some reason.

It's not necessary for us at all to have it as you mention since in our smoke tests we're specifying a tolerance every time. I'm good to remove this if you like :)

emersodb: I'm good to leave it up to you.

tests/smoke_tests/run_smoke_test.py
nerdai: good call.. done!

tests/smoke_tests/run_smoke_test.py
nerdai: Yeah, I thought of this as well. Didn't think it was a huge deal, but I can clean it up to pass once thru this list. I'll use an enumerate in the loop. I was opting for syntax cleanliness, but I'm happy to "nitpick" on the efficiency here.

nerdai:
> > ```
> > ================================================================================================== short test summary info ===================================================================================================
> > FAILED tests/smoke_tests/test_smoke_tests.py::test_scaffold - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - prediction - accuracy' (0.4141111111111111) does not match the requested value (0.4062) within requested tolerance (0.005)."]
> > FAILED tests/smoke_tests/test_smoke_tests.py::test_apfl - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - loss - aggregated' (1.5720984935760498) does not match the requested value (1.5862) within requested tolerance (0.005).", "Saved val...
> > FAILED tests/smoke_tests/test_smoke_tests.py::test_feddg_ga - AssertionError: Server metrics check failed. Errors: ["Saved value for metric 'val - loss - aggregated' (1.579448938369751) does not match the requested value (1.5862) within requested tolerance (0.005).", "Saved valu...
> > ============================================================================ 3 failed, 21 passed, 326 deselected, 1 warning in 4015.66s (1:06:55) ============================================================================
> > (fl4-health) vafajardo@gpu079:~/projects/FL4Health$ 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These are the only tests that fail now for me on local and on our cluster (consistency... at least). The two `apfl` and `feddg_ga` are the ones you mentioned before, but I think you had `scaffold` pass locally?
> > In any case, the GPU/device check and tol level worked – the above snippet is from cluster where cuda is available and so tolerance is 0.005.
> 
> Sounds good. I'm going to investigate the SCAFFOLD failures on GPU just to ensure they aren't something that we missed when the algorithm runs on GPU. I'll also double check APFL and FedDG-GA on GPU, but we already know they have local fluctuations on CPUs for whatever reason. So I think it's likely just an numerical error type thing. Won't hold up merging this PR for that. I'll correct whatever I find, if anything, in a follow up.

Sounds good I've marked them as "skipped" when executing these tests on local since they're not working for us at this moment. If after you're investigation you're able to get it resolved, we can remove the skip mark.

fl4health/clients/nnunet_client.py
nerdai: Yeah `AttributeError` was the error I was getting. It makes sense tho since its trying to access an attribute that doesn't exist. 👍🏼

tests/smoke_tests/run_smoke_test.py
nerdai: Sgtm. I'm just going to leave the default there. I hope it's not confusing to folks. It's just a param with a default value at end of the day.

emersodb (approved):
@nerdai: I think this looks good to go. However, do you mind updating the description to reflect the actual changes we settled on? That will just make it easier to search for in the event we ever need to.

PR #316, opened 2025-01-16
Smoke test bug fix for Andrei with the state checkpointers

emersodb:
# PR Type
Fix

# Short Description

When switching off of the default directory "./" for the state checkpointer smoke tests, Andrei uncovered a bug in the way we were checking for the existence of a checkpoint and loading the intermediate state.

# Tests Added

Smoke tests now pass without the default directory specified.

### DIFF
--- a/fl4health/checkpointing/checkpointer.py
+++ b/fl4health/checkpointing/checkpointer.py
@@ -303,17 +303,18 @@ class PerRoundStateCheckpointer:
             dict[str, Any]: A dictionary representing the checkpointed state, as loaded by torch.load.
         """
 
+        assert self.checkpoint_exists(checkpoint_name)
         checkpoint_path = os.path.join(self.checkpoint_dir, checkpoint_name)
         log(INFO, f"Loading state from checkpoint at {checkpoint_path}")
-        assert self.checkpoint_exists(checkpoint_path)
 
         return torch.load(checkpoint_path)
 
-    def checkpoint_exists(self, checkpoint_path: str) -> bool:
+    def checkpoint_exists(self, checkpoint_name: str) -> bool:
         """
         Checks if a checkpoint exists at the checkpoint_path constructed at initialization.
 
         Returns:
             bool: Whether or not a checkpoint exists.
         """
+        checkpoint_path = os.path.join(self.checkpoint_dir, checkpoint_name)
         return os.path.exists(checkpoint_path)
--- a/fl4health/checkpointing/checkpointer.py
+++ b/fl4health/checkpointing/checkpointer.py
@@ -311,7 +311,7 @@ class PerRoundStateCheckpointer:
 
     def checkpoint_exists(self, checkpoint_name: str) -> bool:
         """
-        Checks if a checkpoint exists at the checkpoint_path constructed at initialization.
+        Checks if a checkpoint exists at the checkpoint_dir + checkpoint_name constructed at initialization.
 
         Returns:
             bool: Whether or not a checkpoint exists.
--- a/fl4health/checkpointing/checkpointer.py
+++ b/fl4health/checkpointing/checkpointer.py
@@ -311,7 +311,7 @@ class PerRoundStateCheckpointer:
 
     def checkpoint_exists(self, checkpoint_name: str) -> bool:
         """
-        Checks if a checkpoint exists at the checkpoint_dir + checkpoint_name constructed at initialization.
+        Checks if a checkpoint exists at the checkpoint_dir constructed at initialization + checkpoint_name.
 
         Returns:
             bool: Whether or not a checkpoint exists.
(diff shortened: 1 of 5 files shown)


vector-admin:
Task linked: [CU-868byqhk5 [Fix] Smoke tests discrepancy when running on GPUs](https://app.clickup.com/t/868byqhk5)

fl4health/checkpointing/checkpointer.py
```diff
@@ -303,17 +303,18 @@ def load_checkpoint(self, checkpoint_name: str) -> dict[str, Any]:
...
         checkpoint_path = os.path.join(self.checkpoint_dir, checkpoint_name)
         log(INFO, f"Loading state from checkpoint at {checkpoint_path}")
-        assert self.checkpoint_exists(checkpoint_path)
 
         return torch.load(checkpoint_path)
 
-    def checkpoint_exists(self, checkpoint_path: str) -> bool:
+    def checkpoint_exists(self, checkpoint_name: str) -> bool:
         """
         Checks if a checkpoint exists at the checkpoint_path constructed at initialization.
```
nerdai, line 314: nit: should this `checkpoint_path` be `checkpoint_name` to match the modified param name.

Also is this internal method and/or do we care if that this is breaking? That is, if I have code that uses this fn with named param `checkpoint_path` then if I update my `fl4health` my existing script will now fail.

Our next release should be at least a `minor` bump then.

fl4health/checkpointing/checkpointer.py
```diff
@@ -303,17 +303,18 @@ def load_checkpoint(self, checkpoint_name: str) -> dict[str, Any]:
...
 
         Returns:
             bool: Whether or not a checkpoint exists.
         """
+        checkpoint_path = os.path.join(self.checkpoint_dir, checkpoint_name)
```
nerdai, line 319: noice!

tests/smoke_tests/run_smoke_test.py
```diff
@@ -551,7 +551,7 @@ class MetricType(Enum):
 
 
 DEFAULT_METRICS_FOLDER = Path("metrics")
-DEFAULT_TOLERANCE = 0.1
+DEFAULT_TOLERANCE = 0.0005
```
nerdai, line 554: sigh of relief!

nerdai (approved):
Generally LGTM!

Left one comment about a potential breaking change but not sure if we care? If we do, then we should maybe add `kwargs` to catch the removed `checkpoint_path` and then handle it or throw some error/warning

fl4health/checkpointing/checkpointer.py
emersodb: I think we care that it's breaking but it's also not something we want to "back-support" since it's a serious issue with the functionality to leave it.

emersodb: Good point on the documentation update

nerdai: Ah okay, in that case, maybe we should at least catch the modified param and throw an error. Something like:

```python
def checkpoint_exists(self, checkpoint_name: str, **kwargs) -> bool:
    if `checkpoint_path` in kwargs:
          msg = ... # some good msg
          raise ValueError(msg)  # or some custom error if have those in our library
```

emersodb: I see what you mean...I'm good to add that.

nerdai: Sounds good. It's probably a little overkill for our stage, but we could just build the habit for these sorts of things.

We can leave a note that we'll leave this in here for "x" period of time. After that period we're probably safe to remove it

nerdai (approved):
LGTM! Noice!

fl4health/checkpointing/checkpointer.py
```diff
@@ -303,17 +303,23 @@ def load_checkpoint(self, checkpoint_name: str) -> dict[str, Any]:
...
         """
-        Checks if a checkpoint exists at the checkpoint_path constructed at initialization.
+        Checks if a checkpoint exists at the checkpoint_dir constructed at initialization + checkpoint_name.
 
         Returns:
             bool: Whether or not a checkpoint exists.
         """
+        if "checkpoint_path" in kwargs:
+            raise ValueError(
+                "Previously this checkpoint supported sending a path, but it requires now requires a "
```
nerdai, line 321: ohh minor nit: I think `require` appears twice by accident here.

emersodb: Yup. Got it.

scarere:
Just going to add a loose comment here that overhauling some of the checkpointing functionality is in our backlog. Especially in the nnunet client where some bugs were introduced.

emersodb:
> Just going to add a loose comment here that overhauling some of the checkpointing functionality is in our backlog. Especially in the nnunet client where some bugs were introduced.

Yup...that's in the todos. This fix is just to correct the current implementation of fault tolerance when using a non-default directory.

PR #318, opened 2025-01-23
Move assert statements out of `run_smoke_test` and into the actual test (for graceful shutdown in case of failure)

nerdai:
# PR Type
Fix

# Short Description

**THIS PR BROUGHT THE SMOKE**  💨 💨 💨
image

This PR addresses an issue with our smoke tests where if one fails (the usual culprit being `test_client_level_dp_breast_cancer`) then it torpedos all of the remaining tests that follow it. Moreover, the error logs of the job looked super scary!

To address this issue, this PR:

1. Ensures that each test explicitly fails if an `Exception` is encountered during the test.
2. Aims to ensure that any lingering cancelled tasks in the event of unexpected test failure (due to Timeout or another reason) are cleared. The reason why we saw the domino effect is because if these cancelled tasks are not cleared, then they will be the next time an event loop starts running. In our case the next test would run the event loop but the hangover cancelled tasks would take precedence and then "stop" the loop and thus cause the next test to crash.

In addition to the above:
- We also move the asserts used in `run_smoke_test` to the actual testing module `test_smoke_test.py` and "within" the test themselves directly. This is mostly for hygiene to follow typical pytest conventions which makes it easier to see why a test might fail and explicitly make it fail if an exception is raised.
- Also the progress of these smoke tests can now actually be viewed in the logs of job as its running.
image

NOTE:
If `test_client_level_dp_breast_cancer` fails, it will still fail the overall smoke test job which mean's we'll need to run all of the tests even passing ones again. In another PR we could mark this test with something like "flaky" and then create a separate job for flaky tests and so re-running these would be less time consuming.

# Tests Added

N/A

### DIFF
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - main
-  workflow_dispatch:
 
 jobs:
   test:
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:  # for testing with github runners and stability
 
 jobs:
   test:
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -7,7 +7,6 @@ on:
   pull_request:
     branches:
       - main
-  workflow_dispatch:  # for testing with github runners and stability
 
 jobs:
   test:
--- a/.github/workflows/smoke_tests.yaml
+++ b/.github/workflows/smoke_tests.yaml
@@ -47,4 +47,4 @@ jobs:
       - name: Run Script
         run: |
           source .venv/bin/activate
-          pytest -m "smoketest"
+          pytest -m "smoketest" -v
--- a/tests/smoke_tests/run_smoke_test.py
+++ b/tests/smoke_tests/run_smoke_test.py
@@ -1,5 +1,4 @@
 import asyncio
-import datetime
 import json
 import logging
 import re
@@ -555,15 +554,12 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
     process_name: str,
     timeout: int = 300,  # timeout for the whole process to complete
 ) -> str:
-    logger.info(f"Collecting output for {process_name}...")
-    full_output = ""
-    try:
-        if not (process.stdout is not None):
-            raise SmokeTestExecutionError("Process stdout is None")
-        start_time = datetime.datetime.now()
+
+    async def get_output_from_stdout(stream_reader: asyncio.streams.StreamReader) -> tuple[str, int | None]:
+        full_output = ""
         while True:
-            # giving a smaller timeout here just in case it hangs for a long time waiting for a single log line
-            output_in_bytes = await asyncio.wait_for(process.stdout.readline(), timeout=timeout)
+            output_in_bytes = await stream_reader.readline()
+            await asyncio.sleep(0)  # give control back to loop manager
             output = output_in_bytes.decode().replace("\\n", "\n")
             logger.debug(f"{process_name} output: {output}")
             full_output += output
@@ -572,12 +568,15 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
             if output == "" and return_code is not None:
                 break
 
-            elapsed_time = datetime.datetime.now() - start_time
-            if elapsed_time.seconds > timeout:
-                raise Exception(f"Timeout limit of {timeout}s exceeded waiting for {process_name} to finish execution")
+        return full_output, return_code
 
+    logger.info(f"Collecting output for {process_name}...")
+
+    try:
+        if not (process.stdout is not None):
+            raise SmokeTestExecutionError("Process stdout is None")
+        full_output, return_code = await asyncio.wait_for(get_output_from_stdout(process.stdout), timeout=timeout)
     except Exception as ex:
-        logger.error(f"{process_name} output:\n{full_output}")
         logger.exception(f"Error collecting {process_name} log messages:")
         raise ex
 
--- a/tests/smoke_tests/run_smoke_test.py
+++ b/tests/smoke_tests/run_smoke_test.py
@@ -24,6 +24,15 @@ logger = logging.getLogger()
 DEFAULT_TOLERANCE = 0.0005
 
 
+# Custom Errors
+class SmokeTestAssertError(Exception):
+    pass
+
+
+class SmokeTestExecutionError(Exception):
+    pass
+
+
 def postprocess_logs(logs: str) -> str:
     """Postprocess logs to remove spurious Errors.
 
@@ -195,7 +204,8 @@ async def run_smoke_test(
     output_found = False
     while not output_found:
         try:
-            assert server_process.stdout is not None, "Server's process stdout is None"
+            if not (server_process.stdout is not None):
+                raise SmokeTestExecutionError("Server's process stdout is None")
             server_output_in_bytes = await asyncio.wait_for(server_process.stdout.readline(), 20)
             server_output = server_output_in_bytes.decode()
             logger.debug(f"Server output: {server_output}")
@@ -205,16 +215,16 @@ async def run_smoke_test(
             break
 
         return_code = server_process.returncode
-        assert return_code is None or (return_code is not None and return_code == 0), (
-            f"Full output:\n{full_server_output}\n" f"[ASSERT ERROR] Server exited with code {return_code}."
-        )
+        if not (return_code is None or (return_code is not None and return_code == 0)):
+            msg = f"Full output:\n{full_server_output}\n" f"[ASSERT ERROR] Server exited with code {return_code}."
+            raise SmokeTestAssertError(msg)
 
         if any(startup_message in server_output for startup_message in startup_messages):
             output_found = True
 
-    assert output_found, (
-        f"Full output:\n{full_server_output}\n" f"[ASSERT_ERROR] Startup log message not found in server output."
-    )
+    if not output_found:
+        msg = f"Full output:\n{full_server_output}\n" f"[ASSERT_ERROR] Startup log message not found in server output."
+        raise SmokeTestAssertError(msg)
 
     logger.info("Server started")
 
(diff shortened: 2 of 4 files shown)


tests/smoke_tests/run_smoke_test.py
```diff
@@ -50,7 +59,7 @@ async def run_smoke_test(
     client_metrics: dict[str, Any] | None = None,
     # assertion params
     tolerance: float = DEFAULT_TOLERANCE,
-) -> None:
+) -> tuple[list[str], list[str]]:
```
nerdai, line 62: This now returns `server_errors` and `client_errors` to the caller.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -201,16 +215,16 @@ async def run_smoke_test(
             break
 
         return_code = server_process.returncode
-        assert return_code is None or (return_code is not None and return_code == 0), (
```
nerdai, line 204: In order to remove these `assert` that avoided me dangerously doing Boolean algebra, I employ the following pattern. We can change this if we want...

```python
# old
assert <cond>, <fail_msg>

# new
if not <cond>:
    raise SmokeTestError(fail_msg)
```

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -98,87 +162,156 @@ async def test_feddg_ga(tolerance: float) -> None:
...
+                server_python_path="examples.dp_fed_examples.client_level_dp_weighted.server",
+                client_python_path="examples.dp_fed_examples.client_level_dp_weighted.client",
+                config_path="tests/smoke_tests/client_level_dp_weighted_config.yaml",
+                dataset_path="examples/datasets/breast_cancer_data/hospital_0.csv",
+                skip_assert_client_fl_rounds=True,
+                tolerance=tolerance,
+            )
+            task = asyncio.create_task(coro)
+            await task
+        except SmokeTestTimeoutError as e:
```
nerdai, line 233: This culprit test will only attempt to retry if `SmokeTestTimeoutError` is raised. Otherwise the fail is due to another, potentially real error and we fail the test in this case.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -240,74 +259,96 @@ async def run_smoke_test(
...
             )
+            raise SmokeTestAssertError(msg)
+
+        if assert_evaluation_logs:
+            if not ("Client Evaluation Local Model Metrics" in full_client_output):
+                msg = (
+                    f"Full client output:\n{full_client_output}\n"
+                    f"[ASSERT ERROR] 'Client Evaluation Local Model Metrics' message not found for client {i}."
+                )
+                raise SmokeTestAssertError(msg)
```
nerdai, line 339: Tbh, I am not sure if I like the name `SmokeTestAssertError` it makes me feel like its asserting on a good output against an expected value.

I think these are probably better as `SmokeTestExecutionError`? I was just following our naming with [ASSERT ERROR] but I think this convention is confusing.

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -1,21 +1,45 @@
...
+
 
 @pytest.mark.smoketest
-@pytest.mark.asyncio()
 async def test_basic_server_client_cifar(tolerance: float, tmp_path: Path) -> None:
     checkpoint_dir = tmp_path / "checkpoints"
     checkpoint_dir.mkdir(exist_ok=True)
-
-    await run_fault_tolerance_smoke_test(
+    coro = run_fault_tolerance_smoke_test(
```
emersodb, line 42: Super minor, but maybe we expand this to `coroutine`?

tests/smoke_tests/run_smoke_test.py
```diff
@@ -521,14 +567,12 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
...
-    logger.info(f"Collecting output for {process_name}...")
-    full_output = ""
-    try:
-        assert process.stdout
-        start_time = datetime.datetime.now()
+
+    async def get_output_from_stdout(stream_reader: asyncio.streams.StreamReader) -> tuple[str, int | None]:
+        full_output = ""
         while True:
-            # giving a smaller timeout here just in case it hangs for a long time waiting for a single log line
```
nerdai, line 530: I cleaned this up, but note that we weren't actually giving this "inner task" a smaller timeout.

Instead I outsource this logic to a contained method `get_output_from_stdout()` which reads the stream until completion. This whole process of reading from the stream is what I assign a timeout for. (No more need for manual computation if timeout was reached)

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -27,36 +51,60 @@ async def test_basic_server_client_cifar(tolerance: float, tmp_path: Path) -> No
         intermediate_checkpoint_dir=checkpoint_dir.as_posix(),
         tolerance=tolerance,
     )
+    task = asyncio.create_task(coro)
```
emersodb, line 54: Since we use this code snippet in a bunch of places, is it possible to encapsulate it in a small function?

emersodb (approved):
Some really minor comments. Otherwise, this looks awesome.

tests/smoke_tests/run_smoke_test.py
```diff
@@ -521,14 +563,12 @@ async def _wait_for_process_to_finish_and_retrieve_logs(
...
 ) -> str:
-    logger.info(f"Collecting output for {process_name}...")
-    full_output = ""
-    try:
-        assert process.stdout
-        start_time = datetime.datetime.now()
+
+    async def get_output_from_stdout(stream_reader: asyncio.streams.StreamReader) -> tuple[str, int | None]:
+        full_output = ""
         while True:
```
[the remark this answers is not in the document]
nerdai: Ohh, it's actually still required. It wasn't immediately obvious to me that it was needed because the coro is produced by `process.stdout.readline()` which doesn't seem like it needs a loop.

However, `process.stdout` is actually a `StreamReader` and we need to iteratively read from the stream until we reach the end of the stream. So the `while True` makes sense, so I kept it, I just moved it in a func called `get_output_from_stdout` for better readability but also so that I can wrap a timeout on it.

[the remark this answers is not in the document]
nerdai: In other words, the overall functionality hasn't changed, but I just polished how we invoke it.

.github/workflows/smoke_tests.yaml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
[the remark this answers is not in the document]
nerdai: Oh yeah, I can remove this. It's for manually triggering workflows. For example, with this I can trigger a workflow from cli:

```sh
gh workflow run "Smoke Tests" --ref nerdai/smoke-tests-explicitly-assert
```

That's how I hammer our github action runners with these smoke tests.

tests/smoke_tests/test_smoke_tests.py
nerdai: sure! no probs

tests/smoke_tests/test_smoke_tests.py
nerdai: Yeah, that's probably a good idea. I thought it was nice to leave the tests each explicitly see the task and waiting it, but I think you're right. I just added a helper fn for this to reduce repeatability.

tests/smoke_tests/run_smoke_test.py
nerdai: I'm lazy now. I'm going to leave it as is lol.

.github/workflows/smoke_tests.yaml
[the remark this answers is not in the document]
nerdai: removed it just in case... only needed it for development.

tests/smoke_tests/run_smoke_test.py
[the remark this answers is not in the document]
emersodb: Sounds good. Perhaps we can still leave a comment summarizing the above so we don't forget?

emersodb (approved):
Good to go!

tests/smoke_tests/run_smoke_test.py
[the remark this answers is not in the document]
nerdai: Yup, I can add a quick comment for this.

nerdai:
```
collecting ... collected 350 items / 326 deselected / 24 selected
tests/smoke_tests/test_smoke_tests.py::test_basic_server_client_cifar PASSED [  4%]
tests/smoke_tests/test_smoke_tests.py::test_nnunet_config_2d PASSED      [  8%]
tests/smoke_tests/test_smoke_tests.py::test_nnunet_config_3d PASSED      [ 12%]
tests/smoke_tests/test_smoke_tests.py::test_scaffold PASSED              [ 16%]
tests/smoke_tests/test_smoke_tests.py::test_apfl PASSED                  [ 20%]
tests/smoke_tests/test_smoke_tests.py::test_feddg_ga PASSED              [ [25](https://github.com/VectorInstitute/FL4Health/actions/runs/12958022255/job/36147520295#step:9:26)%]
tests/smoke_tests/test_smoke_tests.py::test_basic PASSED                 [ 29%]
tests/smoke_tests/test_smoke_tests.py::test_client_level_dp_cifar PASSED [ 33%]
tests/smoke_tests/test_smoke_tests.py::test_client_level_dp_breast_cancer FAILED [ 37%]
tests/smoke_tests/test_smoke_tests.py::test_instance_level_dp_cifar PASSED [ 41%]
tests/smoke_tests/test_smoke_tests.py::test_dp_scaffold PASSED           [ 45%]
tests/smoke_tests/test_smoke_tests.py::test_fedbn PASSED                 [ 50%]
tests/smoke_tests/test_smoke_tests.py::test_fed_eval PASSED              [ 54%]
tests/smoke_tests/test_smoke_tests.py::test_fedper_mnist PASSED          [ 58%]
tests/smoke_tests/test_smoke_tests.py::test_fedper_cifar PASSED          [ 62%]
tests/smoke_tests/test_smoke_tests.py::test_ditto_mnist PASSED           [ 66%]
tests/smoke_tests/test_smoke_tests.py::test_mr_mtl_mnist PASSED          [ 70%]
tests/smoke_tests/test_smoke_tests.py::test_fenda PASSED                 [ 75%]
tests/smoke_tests/test_smoke_tests.py::test_fenda_ditto PASSED           [ 79%]
tests/smoke_tests/test_smoke_tests.py::test_perfcl PASSED                [ 83%]
tests/smoke_tests/test_smoke_tests.py::test_fl_plus_local PASSED         [ 87%]
tests/smoke_tests/test_smoke_tests.py::test_moon PASSED                  [ 91%]
tests/smoke_tests/test_smoke_tests.py::test_ensemble PASSED              [ 95%]
tests/smoke_tests/test_smoke_tests.py::test_flash PASSED                 [100%]
=================================== FAILURES ===================================
______________________ test_client_level_dp_breast_cancer ______________________
```

Smoke tests are not being torpedoed anymore with graceful shutdown and cleaning up tasks. In another PR we may want to address this flaky test or at least give it a different mark/ci-job so that we don't have to re-run all other smoke tests. Retry logic doesn't seem to be working for some reason -- could look into that after...

PR #325, opened 2025-01-28
Fix local plans

scarere:
# Fix

Clickup Ticket(s):
- [Fix nnunet warnings and log spam](https://app.clickup.com/t/868c8gne1)
- [Replace channel properties in local nnunet plans with local fingerprint](https://app.clickup.com/t/868bugrfy)

The nnunet client must modify the global plans and create a local copy. Previously the only fields being updated or checked for edge case errors were the dataset_name, plans_name, data_identifiers and batch_size. This PR improves the local plans file and tailors it more appropriately to the client's local dataset. Changes include

- Updating additional fields containing information about the datasets median image size both before and after resampling
- Updating the foreground_intensity_properties_per_channel. Before we were using stats from the source dataset for normalization and now we will use local stats
- Updating the batch size check. Previously we were using a batch size limit that was based on an estimate of the source dataset's size, now we are accurately setting the 5% upper bound for batch size

Additionally this PR also fixes a bunch of log errors and deprecation warnings that were introduced in previous PR's. Most annoyingly an error where all logs were printed to the console twice. This was present for all clients, not just the nnunet client.

### DIFF
--- a/fl4health/checkpointing/opacus_checkpointer.py
+++ b/fl4health/checkpointing/opacus_checkpointer.py
@@ -1,13 +1,21 @@
 import pickle
-from logging import INFO
+from logging import INFO, getLogger
 from typing import Any
+
 
 import torch.nn as nn
 from flwr.common.logger import log
 from flwr.common.typing import Scalar
+
 from opacus import GradSampleModule
 
 from fl4health.checkpointing.checkpointer import FunctionTorchModuleCheckpointer
+
+# It seems like the opacus package adds a handler to the root logger which causes double logs.
+# By clearing handlers on the root logger we prevent these double logs whilst also allowing the flwr logger to
+# propagate its logs to the root logger.
+root_logger = getLogger()
+root_logger.handlers.clear()
 
 
 class OpacusCheckpointer(FunctionTorchModuleCheckpointer):
@@ -27,9 +35,9 @@ class OpacusCheckpointer(FunctionTorchModuleCheckpointer):
             loss (float): Loss value associated with the model to be used in checkpointing decisions.
             metrics (dict[str, Scalar]): Metrics associated with the model to be used in checkpointing decisions.
         """
-        assert isinstance(
-            model, GradSampleModule
-        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        assert isinstance(model, GradSampleModule), (
+            f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        )
         comparison_score = self.checkpoint_score_function(loss, metrics)
         if self._should_checkpoint(comparison_score):
             log(
@@ -119,9 +127,9 @@ class LatestOpacusCheckpointer(OpacusCheckpointer):
         super().__init__(checkpoint_dir, checkpoint_name, latest_score_function, False)
 
     def maybe_checkpoint(self, model: GradSampleModule, loss: float, _: dict[str, Scalar]) -> None:
-        assert isinstance(
-            model, GradSampleModule
-        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        assert isinstance(model, GradSampleModule), (
+            f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        )
         # Always checkpoint the latest model
         log(INFO, "Saving latest checkpoint with LatestTorchCheckpointer")
         self._extract_and_save_state(model)
@@ -149,9 +157,9 @@ class BestLossOpacusCheckpointer(OpacusCheckpointer):
         )
 
     def maybe_checkpoint(self, model: GradSampleModule, loss: float, metrics: dict[str, Scalar]) -> None:
-        assert isinstance(
-            model, GradSampleModule
-        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        assert isinstance(model, GradSampleModule), (
+            f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
+        )
         # First we use the provided scoring function to produce a score
         comparison_score = self.checkpoint_score_function(loss, metrics)
         if self._should_checkpoint(comparison_score):
--- a/fl4health/checkpointing/opacus_checkpointer.py
+++ b/fl4health/checkpointing/opacus_checkpointer.py
@@ -2,11 +2,9 @@ import pickle
 from logging import INFO, getLogger
 from typing import Any
 
-
 import torch.nn as nn
 from flwr.common.logger import log
 from flwr.common.typing import Scalar
-
 from opacus import GradSampleModule
 
 from fl4health.checkpointing.checkpointer import FunctionTorchModuleCheckpointer
@@ -35,9 +33,9 @@ class OpacusCheckpointer(FunctionTorchModuleCheckpointer):
             loss (float): Loss value associated with the model to be used in checkpointing decisions.
             metrics (dict[str, Scalar]): Metrics associated with the model to be used in checkpointing decisions.
         """
-        assert isinstance(model, GradSampleModule), (
-            f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
-        )
+        assert isinstance(
+            model, GradSampleModule
+        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
         comparison_score = self.checkpoint_score_function(loss, metrics)
         if self._should_checkpoint(comparison_score):
             log(
@@ -127,9 +125,9 @@ class LatestOpacusCheckpointer(OpacusCheckpointer):
         super().__init__(checkpoint_dir, checkpoint_name, latest_score_function, False)
 
     def maybe_checkpoint(self, model: GradSampleModule, loss: float, _: dict[str, Scalar]) -> None:
-        assert isinstance(model, GradSampleModule), (
-            f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
-        )
+        assert isinstance(
+            model, GradSampleModule
+        ), f"Model is of type: {type(model)}. This checkpointer need only be used to checkpoint Opacus modules"
         # Always checkpoint the latest model
         log(INFO, "Saving latest checkpoint with LatestTorchCheckpointer")
         self._extract_and_save_state(model)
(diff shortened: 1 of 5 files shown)


fl4health/checkpointing/opacus_checkpointer.py
```diff
@@ -9,6 +9,12 @@
 
 from fl4health.checkpointing.checkpointer import FunctionTorchModuleCheckpointer
 
+# It seems like the opacus package adds a handler to the root logger which causes double logs.
+# By clearing handlers on the root logger we prevent these double logs whilst also allowing the flwr logger to
+# propagate its logs to the root logger.
+root_logger = getLogger()
+root_logger.handlers.clear()
```
emersodb, line 16: I don't know why logging is so complicated sometimes 😂. Thanks for the comment and the fixup

fl4health/clients/nnunet_client.py
```diff
@@ -360,61 +364,93 @@ def get_lr_scheduler(self, optimizer_key: str, config: Config) -> _LRScheduler:
...
+              rather than using class attributes.
         """
-        # Get the nnunet plans specified by the server
+        # Get the source nnunet plans specified by the server
         plans = pickle.loads(narrow_dict_type(config, "nnunet_plans", bytes))
 
         # Change plans name.
         if self.plans_name is None:
-            self.plans_name = "FL_source-" + plans["plans_name"]
+            self.plans_name = "FL-" + plans["plans_name"] + f"-{self.dataset_id}local"
```
emersodb, line 396: Super minor, but I think this could all be one f-string as 
```
self.plans_name = f"FL-{plans['plans_name']}-{self.dataset_id}local"
```
note the use of the single quote on the dictionary to avoid f-string complaining 🙂

fl4health/clients/nnunet_client.py
```diff
@@ -360,61 +364,93 @@ def get_lr_scheduler(self, optimizer_key: str, config: Config) -> _LRScheduler:
...
 
+        # Load the dataset fingerprint for the local client dataset
+        fp_path = Path(nnUNet_preprocessed) / self.dataset_name / "dataset_fingerprint.json"
+        assert fp_path.exists(), "Could not find the dataset fingerprint file"
+        fp = load_json(fp_path)
+
+        # Change the foreground intensity properties per channel
+        plans["foreground_intensity_properties_per_channel"] = fp["foreground_intensity_properties_per_channel"]
+
+        # Compute the median image size and spacing of the local client dataset
```
emersodb, line 412: Super minor but encapsulating these steps and the ones below in two functions with descriptive names would likely be a bit easier to read.

fl4health/clients/nnunet_client.py
```diff
@@ -360,61 +364,93 @@ def get_lr_scheduler(self, optimizer_key: str, config: Config) -> _LRScheduler:
...
             self.data_identifier = self.plans_name
 
-        # Get maximum number of voxels for a batch based on dataset size
-        # TODO: This function assumes the median image size of the local dataset is the
-        # same as the one from which the plans file was created. Better way to do it is
-        # to pass fingerprint as a param and compute median image size manually.
-        n_cases = self.dataset_json["numTraining"]
-        max_voxels = get_dataset_n_voxels(plans, n_cases) * 0.05  # Max is 5% of total
+        # To be consistent with nnunet, a batch cannot contain more than 5% of the voxels in the dataset.
+        max_voxels = np.prod(resampled_median_shape, dtype=np.float64) * self.dataset_json["numTraining"] * 0.05
```
emersodb, line 430: 🎉

emersodb (approved):
A couple of quite minor comments. I'm going to trust that you've tested your modifications and calculations in here thoroughly, as it's difficult for me to verify their correctness 🙂. At some point we should probably write some formal tests of some of these manipulations, but that can be done in a later PR.

emersodb:
> Most annoyingly an error where all logs were printed to the console twice. This was present for all clients, not just the nnunet client.

With respect to what you wrote above about the logging happening twice. Does the change you made to the Opacus client fix that issue? I might be misunderstanding whether or not that double logging is actually fixed in this PR.

scarere:
> > Most annoyingly an error where all logs were printed to the console twice. This was present for all clients, not just the nnunet client.
> 
> With respect to what you wrote above about the logging happening twice. Does the change you made to the Opacus client fix that issue? I might be misunderstanding whether or not that double logging is actually fixed in this PR.

Well actually I only implemented the fix to the opacus checkpointer as that was the only import affecting the logs in the basic and nnunet examples. But now that you mention it, the double logging is likely to happen if opacus is imported somewhere else without the root logger being cleared of handlers. I can try and add this to all files that import opacus

emersodb:
> > > Most annoyingly an error where all logs were printed to the console twice. This was present for all clients, not just the nnunet client.
> > 
> > 
> > With respect to what you wrote above about the logging happening twice. Does the change you made to the Opacus client fix that issue? I might be misunderstanding whether or not that double logging is actually fixed in this PR.
> 
> Well actually I only implemented the fix to the opacus checkpointer as that was the only import affecting the logs in the basic and nnunet examples. But now that you mention it, the double logging is likely to happen if opacus is imported somewhere else without the root logger being cleared of handlers. I can try and add this to all files that import opacus

No no...don't worry. We can do that another time. That's for surfacing that. Really silly.

fl4health/clients/nnunet_client.py
scarere: What exactly did you have in mind? I'm having trouble trying to think of a way to do it gracefully without it just being way more lines of code.

emersodb: Feel free to ignore, but I was roughly thinking something like ```def modify_plans_file_with_median_size_and_spacing(self, fp: dict[str, Any], plans: dict[str, Any]) -> None:
     """
     Doc stuff

     Args:
         fp (dict[str, Any]): _description_
         plans (dict[str, Any]): _description_
     """
     # Compute the median image size and spacing of the local client dataset
     median_shape = np.median(fp["shapes_after_crop"], axis=0)[plans["transpose_forward"]]
     median_spacing = np.median(fp["spacings"], axis=0)[plans["transpose_forward"]]
     plans["original_median_shape_after_transp"] = [int(round(i)) for i in median_shape]
     plans["original_median_spacing_after_transp"] = [float(i) for i in median_spacing]

def get_median_shape_after_resampling_and_max_voxels(self, fp: dict[str, Any], plans: dict[str, Any]) -> tuple[list[int], float]:
    """
    Doc stuff

    Args:
        fp (dict[str, Any]): _description_
        plans (dict[str, Any]): _description_

    Returns:
        tuple[list[int], float]: _description_
    """
    # Get the median shape after resampling to the desired/input voxel spacing
    fullres_spacing = plans["configurations"]["3d_fullres"]["spacing"]
    resampled_shapes = [
        compute_new_shape(i, j, fullres_spacing) for i, j in zip(fp["shapes_after_crop"], fp["spacings"])
    ]
    resampled_median_shape = np.median(resampled_shapes, axis=0)[plans["transpose_forward"]].tolist()

    # To be consistent with nnunet, a batch cannot contain more than 5% of the voxels in the dataset.
    max_voxels = np.prod(resampled_median_shape, dtype=np.float64) * self.dataset_json["numTraining"] * 0.05

    return resampled_median_shape, max_voxels


def create_plans(self, config: Config) -> dict[str, Any]:
    """
    Modifies the provided plans file to work with the local client dataset and then saves it to disk. Requires the
    local dataset_fingerprint.json to exist, the local dataset_name, plans_name, data_identifier and dataset_json.

    The following fields are modified:
        - plans_name
        - dataset_name
        - original_median_shape_after_transp
        - original_median_spacing_after_transp
        - configurations.{config}.data_identifier
        - configurations.{config}.batch_size
        - configurations.{config}.median_image_size_in_voxels
        - foreground_intensity_properties_per_channel

    Args:
        config (Config): The config provided by the server. Expects the 'nnunet_plans' key with a pickled
            dictionary as the value

    Returns:
        dict[str, Any]: The modified nnunet plans for the client

    TODO:
        - Make this an external function or part of another class and explicitly accept the required arguments
          rather than using class attributes.
    """
    # Get the source nnunet plans specified by the server
    plans = pickle.loads(narrow_dict_type(config, "nnunet_plans", bytes))

    # Change plans name.
    if self.plans_name is None:
        self.plans_name = "FL-" + plans["plans_name"] + f"-{self.dataset_id}local"

    plans["source_plans_name"] = plans["plans_name"]
    plans["plans_name"] = self.plans_name

    # Change dataset name
    plans["dataset_name"] = self.dataset_name

    # Load the dataset fingerprint for the local client dataset
    fp_path = Path(nnUNet_preprocessed) / self.dataset_name / "dataset_fingerprint.json"
    assert fp_path.exists(), "Could not find the dataset fingerprint file"
    fp = load_json(fp_path)

    # Change the foreground intensity properties per channel
    plans["foreground_intensity_properties_per_channel"] = fp["foreground_intensity_properties_per_channel"]

    self.modify_plans_file_with_median_size_and_spacing(fp, plans)

    resampled_median_shape, max_voxels = self.get_median_shape_after_resampling_and_max_voxels(fp, plans)

    # Change data identifier
    if self.data_identifier is None:
        self.data_identifier = self.plans_name

    # Iterate through nnunet configs in plans file
    for c in plans["configurations"].keys():
        # Change the data identifier
        plans["configurations"][c]["data_identifier"] = self.data_identifier + "_" + c

        # Ensure batch size is at least 2 and at most 5 percent of dataset
        # Otherwise we keep it the same as it affects the target VRAM consumption
        if "batch_size" in plans["configurations"][c].keys():
            old_bs = plans["configurations"][c]["batch_size"]
            bs_5percent = round(max_voxels / np.prod(plans["configurations"][c]["patch_size"], dtype=np.float64))
            new_bs = max(min(old_bs, bs_5percent), 2)
            plans["configurations"][c]["batch_size"] = new_bs
        else:
            log(WARNING, f"Did not find a 'batch_size' key in the nnunet plans dict for nnunet config: {c}")

        # Update median shape of resampled input images
        if str(c).startswith("2d"):
            plans["configurations"][c]["median_image_size_in_voxels"] = resampled_median_shape[1:]
        else:
            plans["configurations"][c]["median_image_size_in_voxels"] = resampled_median_shape

    # Save local plans file
    os.makedirs(join(nnUNet_preprocessed, self.dataset_name), exist_ok=True)
    plans_save_path = join(nnUNet_preprocessed, self.dataset_name, self.plans_name + ".json")
    save_json(plans, plans_save_path, sort_keys=False)
    return plans
```

emersodb: However, I'm okay with you leaving this to later. It isn't some crazy long function or anything.

scarere: I see what you mean. I agree it would be nice to split up the function, I was trying to think of a way to do it without implicitly modifying the plans dictionary. Maybe if I make a copy and return a new plans dictionary? Just to play devils advocate a bit though, I'd eventually like to reduce the scope of the nnunet client as I believe the number of methods has grown too large. I added a ticket on clickup, perhaps this change can wait until that PR?

emersodb: Yup...let's do it there and leave these as they are.

scarere: Very much looking forward to when I have time to refactor. No just the nnunet client but hopefully the basic client as well. I love deleting code!

PR #334, opened 2025-02-04
[CI/CD] More Codecov configurations

nerdai:
# PR Type
Other

# Short Description

Codecov annotations popping up all over the place in our PRs, which are a bit annoying. This PR turns that off by specifying the correct toggle in `codecov.yml`. (Thanks to @emersodb for initial investigations on how to do this.)

In this PR, I temporarily added this `DummyTestClass` without any tests coverage to see if annotations were turned off in the PRs. Happy to report that is the case.
image

Showing here as well that while the annotations are off, the checks are still in place:
image

I'll remove this dummy test class after reviewers have given green light on this PR>

# Tests Added

N/A

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,3 +4,16 @@ coverage:
       default:
         target: 80%
         threshold: 5%
+ignore:
+  - ".gitignore"
+  - ".flake8"
+  - "poetry.lock"
+  - "**/*.yaml"
+  - "**/*.yml"
+  - "**/*.toml"
+  - "**/*.md"
+  - "**/*.sh"
+  - "**/*.ini"
+  - "research"
+github_checks:
+    annotations: false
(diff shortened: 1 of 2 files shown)


emersodb (approved):
Nice. Just make sure to remove your dummy code 🙂

nerdai:
Haha yea, removing dummy code now! thanks

lotif (approved):
LGTM but it seems like codecov is failing with the coverage below the threshold. If it's going to check the whole project every time I think we should set to something closer to the project current coverage +- the threshold.

nerdai:
> LGTM but it seems like codecov is failing with the coverage below the target. If it's going to check the whole project every time I think we should set to something closer to the project current coverage +- the threshold.

Ya I agree:

The project coverage check failed in latest merge commit in `main` as well.

image

Going to set target to 65% for now. With the current threshold we should be passing project coverage.

PR #335, opened 2025-02-05
Resolving Some Warnings in Docs Compilation and Improving Docs Rendering Across the Library

emersodb:
# PR Type
Documentation

# Short Description

Clickup Ticket(s): N/A

We recently added Sphinx compiled documentation and hosting. It looks decent, but we hadn't ever rendered our docstrings in a real way before. So we had a number of "issues" in the way they looked etc. I took a first pass at cleaning up as much of these as I could find. **It wasn't very fun** 😂.

In doing so, I touched a **ton** of files very superficially. I'd say that, during review, we need not look over every change I made, but rather see some of them to understand where issues popped up in rendering and how to address them.

# Tests Added

N/A. I made few if any actually functional changes.


nerdai (approved):
Such a scary big PR. Awesome work! Docs are in a much better place as a result of this.

lotif (approved):
👏🏼

PR #336, opened 2025-02-06
Updating the poetry lock and fixing up test.

emersodb:
# PR Type
Other

# Short Description

Clickup Ticket(s): N/A

Updating poetry lock to be compatible with some poetry upgrades. Shouldn't change anything about our workflow.

In the course of doing this, I realized that we have a unit test that doesn't seem to pass locally for our macs (confirmed to fail on my new machine, Andrei's machine, and Marcelo's machine) but does pass on my old intel based mac and the github runners. I confirmed through debugging that the unit test divergence is due to an accumulation of very small numerical fluctuations that eventually make a large difference. So, for now, I'm going to mark this single failing test as run on github only and it will be skipped locally.

Finally, Marcelo pointed out that it's likely better to use python -m venv to create environments, since not everyone will have virtualenv. So I added that change to the CONTRIBUTING.md

# Tests Added

N/A


tests/utils/synthetic_generator_test.py
```diff
@@ -1,9 +1,14 @@
+import os
+
 import pytest
 import torch
 import torch.nn.functional as F
 
 from fl4health.utils.data_generation import SyntheticIidFedProxDataset, SyntheticNonIidFedProxDataset
 
+# skip some tests that currently fail if running locallly
+IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
```
nerdai, line 10: I'm okay to do this!

Just in case though if you wanted to split up this test based on apple/intel, then I just created a PR into this branch that might do the job:

[PR](https://github.com/VectorInstitute/FL4Health/pull/337/files)

emersodb: Cool...will merge this and then review yours

PR #337, opened 2025-02-06
Split tests for apple/intel

nerdai:
# PR Type

Other

# Short Description

Currently, we skip the unit test `tests/utils/synthetic_generator_test.py::test_generate_client_tensors` if not running in Github because of random fluctuations that differ when running on mac vs intel chips. This PR adds a check on what chip is being used on the test runner (local or remote) so that we can have checks in both cases (which is desirable).

When running locally on my machine, `apple` parameters of this test get ran, but the `intel` one is skipped:

```sh
tests/utils/synthetic_generator_test.py::test_covariance_matrix_construction PASSED                                                                                                                                    [ 16%]
tests/utils/synthetic_generator_test.py::test_get_input_output_tensors PASSED                                                                                                                                          [ 33%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors[intel] SKIPPED (Test expected values are set for intel chips)                                                                                    [ 50%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors[apple] PASSED                                                                                                                                    [ 66%]
tests/utils/synthetic_generator_test.py::test_dataset_generation PASSED                                                                                                                                                [ 83%]
tests/utils/synthetic_generator_test.py::test_iid_dataset_generation PASSED                                                                                                                                            [100%]

================================================================================================ 5 passed, 1 skipped in 1.66s ================================================================================================
```

# Tests Added

This PR is specifically for re-jigging/enhancing a specific test.


nerdai:
When running tests locally, on my machine the apple test passes and the intel one is skipped:

```sh
tests/utils/synthetic_generator_test.py::test_covariance_matrix_construction PASSED                                                                                                                                    [ 98%]
tests/utils/synthetic_generator_test.py::test_get_input_output_tensors PASSED                                                                                                                                          [ 98%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors_intel SKIPPED (Test expected values are set for intel chips)                                                                                     [ 99%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors_apple PASSED                                                                                                                                     [ 99%]
tests/utils/synthetic_generator_test.py::test_dataset_generation PASSED                                                                                                                                                [ 99%]
tests/utils/synthetic_generator_test.py::test_iid_dataset_generation PASSED          
```

tests/utils/synthetic_generator_test.py
```diff
@@ -9,6 +10,11 @@
 # skip some tests that currently fail if running locallly
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
```
emersodb: If we go with this solution, we can drop this variable I think?

emersodb (approved):
One thing to double check is that this test does run on the github remote. That is, the INTEL manufacturer gets popped.

I'm good with this...definitely a bit annoying but it is what it is. Maybe just have @lotif take a quick look to see if he sees anything out of place?

tests/utils/synthetic_generator_test.py
nerdai: Yes, agreed. This was just left there to show you quickly this alternative. But cleaning this up to standalone as a PR that can be merged into main.

nerdai: to be clearer -- this is done!

tests/utils/synthetic_generator_test.py
```diff
@@ -59,8 +60,37 @@ def test_get_input_output_tensors() -> None:
...
+            [1, 5, 680, 103, 4, 3560, 559, 88, 0, 0],
+            [79, 228, 0, 1, 26, 6, 3344, 720, 1, 595],
+            marks=pytest.mark.skipif(not INTEL, reason="Test expected values are set for intel chips"),
+        ),
+        pytest.param(
+            [2, 10, 4932, 2, 24, 16, 1, 13, 0, 0],
+            [21, 547, 575, 192, 0, 1951, 3, 346, 1359, 6],
+            [3644, 2, 194, 5, 0, 3, 0, 1115, 32, 5],
+            [57, 3871, 0, 0, 62, 0, 5, 81, 923, 1],
+            marks=pytest.mark.skipif(not APPLE_SILICON, reason="Test expected values are set for apple m chips"),
```
lotif, line 83: What happens if it's not intel or apple? Maybe the conditions should be `APPLE_SILICON` and `not APPLE_SILICON`?

nerdai: Very good point. Made the changes according to your suggestion.

nerdai: To verify, the tests ran successfully on our Github runners:

```sh
tests/utils/synthetic_generator_test.py::test_covariance_matrix_construction PASSED [ 98%]
tests/utils/synthetic_generator_test.py::test_get_input_output_tensors PASSED [ 98%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors[other-chip] PASSED [ 99%]
tests/utils/synthetic_generator_test.py::test_generate_client_tensors[apple-chip] SKIPPED [ 99%]
tests/utils/synthetic_generator_test.py::test_dataset_generation PASSED  [ 99%]
tests/utils/synthetic_generator_test.py::test_iid_dataset_generation PASSED [100%]
```

nerdai:
Looks like our project coverage reduced to 58% -- and this is in `main` branch. 🤔

emersodb:
> Looks like our project coverage reduced to 58% -- and this is in `main` branch. 🤔
> 
> UPDATE:
> 
> I adjusted target from 65% down to 60%. But let's try to not push this down from here...

I'm a little confused how this ended up reduced, but also how we compute coverage to be honest 🙂. I also have this failure on my PR but then it jumped up after the smoke tests concluded to 
image
So I'm not sure if the one you're seeing above is the coverage prior to running the smoke tests?

lotif:
@nerdai I think the coverage reduced because one of the test files failed to run. Once they all pass, I don't think there is a need to reduce the target.

nerdai:
> @nerdai I think the coverage reduced because one of the test files failed to run. Once they all pass, I don't think there is a need to reduce the target.

Ah okay, that's good then. This coverage stuff is a bit confusing at times ngl. Going to pump it back up then.

lotif:
@nerdai The way codecov works, they post the comment with the coverage as soon as they receive the first report, and they keep editing the comment as they receive more reports. So it will always show a low coverage at first and then it will show the correct coverage once all tests are finished and have sent their reports.

nerdai:
@lotif I'm actually confused still cause how come our `main` branch coverage is below 60%?

[](https://app.codecov.io/gh/VectorInstitute/FL4Health)

lotif:
@nerdai that's a weird one, just changing the lock file led to a drop in coverage? I'm confused as well.

nerdai:
lol, well this is awesome! we're passing now... not 100% sure why but perhaps you're right @lotif!

Merging this bad boy before codecov changes its mind (half joking)

PR #338, opened 2025-02-10
Turn on cudnn.benchmark to accelerate convolutions

scarere:
# Feature

Models trained with the nnunet client were significantly slower than identical models trained with the nnunet CLI. After some in depth profiling I identified that the torch.cudnn.benchmark setting, which defaults to False, was being turned on by the CLI to increase model training efficiency by over a factor of 2 at the cost of some overhead time on the first step.

In this PR I added cudnn.benchmark to the nnunet client. I packaged it along with the JIT compiling as they both serve a similar purpose of increasing the efficiency of training steps at the cost of some overhead time on the first step. It seems unlikely that any user would want to use one but not the other, and therefore they are controlled by the same boolean init argument.

I also changed some of the logic to ensure that the nnUNet_compile environment variable is set to True when the compile argument is True. Previously we were relying on the environment variable not being set (which may not be the case) and nnunet defaulting to using JIT compilation in this scenario, now we explicitly ensure JIT compilation will be used.

In my profiling of training runtimes I identified roughly 10-15s of overhead as a result of using federated learning. Roughly 65% of this time can be attributed to overhead incurred by communication delays between the client and server with the remaining 35% potentially being attributed to overhead introduced by FL4Health.

Lastly I note that the nnunet client has grown too large in scope. One suggested refactor would be to create a subclass of basic client that optionally employs mixed precision training depending on the device used (and perhaps an additional boolean control variable that defaults to False). This is a feature that is likely to be useful in other applications and should be abstracted away from the nnunet client.

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -16,7 +16,7 @@ import torch
 from flwr.common.logger import FLOWER_LOGGER, console_handler, log
 from flwr.common.typing import Config, Scalar
 from torch import nn
-from torch.cuda.amp import GradScaler
+from torch.amp import GradScaler
 from torch.nn.modules.loss import _Loss
 from torch.optim import Optimizer
 from torch.optim.lr_scheduler import _LRScheduler
@@ -109,9 +109,11 @@ class NnunetClient(BasicClient):
                 as default None, the plans identifier will be set as "FL-plansname-000local" where 000 is the
                 ``dataset_id`` and plansname is the "plans_name" value of the source plans file. The original plans
                 will be saved under the ``source_plans_name`` key in the modified plans file.
-            compile (bool, optional): If True, the client will jit compile the pytorch model. This requires some
-                overhead time at the beginning of training to compile the model, but results in faster training times.
-                Defaults to True
+            compile (bool, optional): If set to True, the client will Just-In-Time (JIT) compile the nnUNet model and
+                perform optimizations at the start of training. This process significantly reduces the runtime for
+                nnUNet models, especially for larger models or long-running jobs. However, it introduces some overhead
+                time and computation during the initial step. It is recommended to keep this option enabled. The
+                default value is True.
             always_preprocess (bool, optional): If True, will preprocess the local client dataset even if the
                 preprocessed data already seems to exist. The existence of the preprocessed data is determined by
                 matching the provided ``data_identifier`` with that of the preprocessed data already  on the client.
@@ -141,6 +143,7 @@ class NnunetClient(BasicClient):
                 Defaults to empty dictionary.
         """
         metrics = metrics if metrics else []
+
         # Parent method sets up several class attributes
         super().__init__(
             data_path=Path("dummy/path"),  # data_path not used by NnunetClient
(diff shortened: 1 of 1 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -230,10 +233,6 @@ def train_step(self, input: TorchInputType, target: TorchTargetType) -> tuple[Tr
         target = self.transform_target(target)
         losses = self.compute_training_loss(preds, features, target)
 
-        # Custom backward pass logic with gradient scaling adapted from nnUNetTrainer:
```
emersodb, line 233: Perhaps we still want to keep this pointer comment?

fl4health/clients/nnunet_client.py
```diff
@@ -608,8 +609,6 @@ def predict(self, input: TorchInputType) -> tuple[TorchPredType, dict[str, torch
         """
         if isinstance(input, torch.Tensor):
             # If device type is cuda, nnUNet defaults to mixed precision forward pass
-            # https://github.com/MIC-DKFZ/nnUNet/blob/43349fa5f0680e8109a78dca7215c19e258c9dd7 \
```
emersodb, line 611: Or here? Perhaps we don't need both.

fl4health/clients/nnunet_client.py
```diff
@@ -665,8 +665,6 @@ def compute_loss_and_additional_losses(
             )
 
         # If device type is cuda, nnUNet defaults to compute loss in mixed precision
-        # https://github.com/MIC-DKFZ/nnUNet/blob/43349fa5f0680e8109a78dca7215c19e258c9dd7 \
-        # nnunetv2/training/nnUNetTrainer/nnUNetTrainer.py#L993
         if self.device.type == "cuda":
             with torch.autocast(self.device.type, enabled=True):
                 loss = self.criterion(loss_preds, loss_targets), None
```
emersodb, line 670: Did we confirm that there isn't any benefit to grouping operations under one context here? If I remember correctly, we do this autocast in two places. Just want to make sure there's actually no difference between splitting it versus having it all under one.

emersodb (approved):
Overall, looks good. Just a few small questions.

fl4health/clients/nnunet_client.py
scarere: I compared and didn't see any difference. But that was compared to the following

emersodb: Sounds good. Just wanted to double check that we didn't see any difference after doing the cuda benchmarking. I was thinking that perhaps the effect of not benchmarking was just so large that it would drown out any additional benefit of the grouping above.

scarere: I mean if there was any difference, it was small enough that I did not notice a runtime difference in the profiler. I don't want to say I've ruled it out entirely, but it's at least something we don't need to worry about for now

fl4health/clients/nnunet_client.py
emersodb: That's fair. I guess I was more thinking we'd retain the comment about the custom backward logic and where it occurs in nnunet's code just so we know where to find their implementation. I agree we don't need the mix precision part.

scarere: I'll add a comment

PR #341, opened 2025-02-12
Better Batch Behavior When Training By Steps

emersodb:
# PR Type
Small Fix

# Short Description

Clickup Ticket(s): NA

There are a few things folded into this PR. The first two are the most important. The others are just housekeeping that I'm folding in.

1.  Changing train by steps iterator setup to provided better behavior when shuffle is off for the trainloader. If the trainloader did not have `shuffle = True` (which is rare but not impossible) and the user trained by steps that didn't traverse the entire training set, then the next time they strained by steps (perhaps in the next server round), the iterator started at the beginning again and traversed the same batches in the same order. So you wouldn't see the full dataset. The changes to basic client ensure that the iterator remembers where it left off.
2. Change the way we limit the number of validation steps to run during evaluation. Previously, we were specifying a value in the configuration and then just breaking out of an iteration through the dataloader at the "correct" time. The issue there is that if the validation dataloader doesn't shuffle the data, then we run through the same validation batches every time. Now if `num_validation_steps` is specified in the configuration, we perform steps through an iterator on the dataloader. If the number of validation steps requested is larger than the loader itself, we'll simply loop batch around and keep going. If we don't make it all the way through, we'll pick up where we left off and loop around if there aren't enough steps left to complete the set of steps requested.
3. The FedProx readme referred to an old weights and biases config setup. So I update it. I also changed the example to not use W and B by default, since the example will fail if you don't have a valid W and B entity, which we probably don't want.
4. Moving the `run_fl_local` to do the basic example instead of FedProx, simply because it seems like a more basic first target.
5. Moving a useful function out of `example/functions/` to the core library since it's generally useful.
6. Just doing a bit of documentation updates to the nnunet client and wrapping the lines a little less often.

**NOTE**: This PR also includes some fixes for nnunet that Shawn has implemented and tested. I don't think we need to review those more than he already has.

# Tests Added

Some small unit tests to test a bit of encapsulation and other methods.


examples/fedprox_example/config.yaml
```diff
@@ -15,10 +15,11 @@ n_clients: 3 # The number of clients in the FL experiment
...
 batch_size: 128 # The batch size for client training
 
-reporting_config:
-  project: FL4Health # Name of the project under which everything should be logged
-  name: "FedProx Server" # Name of the run on the server-side, each client will also have it's own run name
-  group: "FedProx Experiment" # Group under which each of the FL run logging will be stored
-  entity: "your_entity_here" # WandB user name
-  notes: "Testing WB reporting"
-  tags: ["Test", "FedProx"]
+# The Reporting Configuration can be uncommented to log training to Weights and Biases
```
emersodb, line 18: Commenting this but leaving it as an example of the config structure. This example would fail if the person didn't have a proper wanb entity, which maybe we don't want by default for an example.

examples/utils/run_fl_local.sh
```diff
@@ -1,17 +1,17 @@
 #!/bin/bash
```
emersodb, line 2: Moving this script to run the basic example by default since that's a gentle entry into the library.

fl4health/clients/basic_client.py
```diff
@@ -128,6 +128,9 @@ def __init__(
         # By specifying this in the config we cannot guarantee the validation set is the same
         # across rounds for clients.
         self.max_num_validation_steps: int | None = None
+        # Train iterator is used in executing training by steps. It is only instantiated if needed.
+        # NOTE: It is of type _BaseDataLoaderIter, which is not importable...so we're forced to use Any
```
emersodb, line 132: It anyone knows how to get around this please let me know. I spent 30 minutes last night trying to do a better job typing and couldn't find anything...

fl4health/utils/config.py
```diff
@@ -95,3 +95,24 @@ def narrow_dict_type_and_set_attribute(
     val = narrow_dict_type(dictionary, dictionary_key, narrow_type_to)
     val = func(val) if func is not None else val
     setattr(self, attribute_name, val)
+
+
+def make_dict_with_epochs_or_steps(local_epochs: int | None = None, local_steps: int | None = None) -> dict[str, int]:
```
emersodb, line 100: This function is generally useful. So just moving it to the core of the library.

emersodb (commented):
Smoke failures for APFL and FedDG-GA are expected since they train by steps and this PR alters the trajectory of training examples they see after the first server round. Will fix them out soon.

fl4health/clients/basic_client.py
```diff
@@ -690,8 +693,10 @@ def train_by_steps(
         """
         self.model.train()
 
-        # Pass loader to iterator so we can step through train loader
-        train_iterator = iter(self.train_loader)
+        # If the train_iterator hasn't been created before, we do so now.
+        if self.train_iterator is None:
```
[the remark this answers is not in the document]
scarere: ie.
```python
try:
    input, target = next(self.train_iterator)
except StopIteration:
    input, target = next(self.train_iterator)
```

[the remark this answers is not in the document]
scarere: Would that work?

fl4health/clients/basic_client.py
```diff
@@ -747,32 +775,31 @@ def _validate_or_test(
...
+        assert logging_mode in [LoggingMode.VALIDATION, LoggingMode.TEST], "logging_mode must be VALIDATION or TEST"
+
         self.model.eval()
         metric_manager.clear()
         loss_meter.clear()
         with torch.no_grad():
             for i, (input, target) in enumerate(maybe_progress_bar(loader, self.progress_bar)):
                 # Limit validation to self.max_num_validation_steps if it is defined
-                if logging_mode == LoggingMode.VALIDATION and self.max_num_validation_steps == i:
+                if self._should_stop_iteration(logging_mode, i):
```
scarere, line 802: Also if we do the break check at the beginning of the loop, then technically we will run throug `self.max_num_val_steps + 1` steps through the val loader and we are just discarding the last step. Would it be better to check at the end of the loop and ensure we only iterate through the loader the intended number of times? This also doesn't protect against the edge case of the first iteration raising a stop condition. My PR is overlapping in scope with this one. For this section

scarere: If we compute the n_val_steps before hand we avoid issues with iterating too many times and edge cases where we don't iterate at all (unless someone sets max_num_val_steps to 0 or the length of the loader is 0, maybe best to have a quick assertion on that)

fl4health/clients/basic_client.py
```diff
@@ -862,15 +889,11 @@ def setup_client(self, config: Config) -> None:
         if "max_num_validation_steps" in config:
             log(
                 INFO,
-                """
-                    max_num_validation_steps specified in config. Only a random subset of batches will \
-                    be sampled from the validation set if max_num_validation_steps is greater \
-                    than the number of batches in the validation dataloader.
-                    """,
+                "max_num_validation_steps specified in config. Only a subset of batches will be processed from the "
```
scarere, line 892: Seeing this now I realize that we should be treating max_num_val_steps as an upper bound. So maybe a revised version of my previous suggestion would be
```python
num_val_steps = min(self.max_num_val_steps, len(loader)) if self.max_num_val_steps is not None and logging_mode == LoggingMode.VALIDATION else len(loader)
```

scarere (commented):
I only took a look at the changes to basic client. It seems like the changes to nnunet_client were just comment formatting. Overall it seems like it make sense to pack the val dataloader stuff into this PR which is why I made comments regarding that. The branch I was working out of was [fix_val_steps](https://github.com/VectorInstitute/FL4Health/tree/fix_val_steps). It also had some other changes packed in that modified the nnunet dataloaders so maybe we merge these branches at some point.

fl4health/clients/basic_client.py
emersodb: I'll try it out and see 🙂

emersodb: Mypy seemed okay with that, so we'll use that for now

fl4health/clients/basic_client.py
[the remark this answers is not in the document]
emersodb: mmm...I think we'd still need to reset the iterator and we might lose some of the underlying mechanics of the dataloader. For example, the reshuffle of data upon restarting the iterations.

fl4health/clients/basic_client.py
emersodb: Yeah I think this is a good point. I'll refactor this a bit to essentially step through an iterator and wrap around if you don't have enough batches to complete the requested steps.

fl4health/clients/basic_client.py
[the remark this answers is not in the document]
scarere: I tried it and it doesn't work anyways. The iterator can't reset itself without the original dataloader

fl4health/clients/basic_client.py
```diff
@@ -830,14 +859,22 @@ def validate(self, include_losses_in_metrics: bool = False) -> tuple[float, dict
...
             (and test if present).
         """
-        val_loss, val_metrics = self._validate_or_test(
-            self.val_loader,
-            self.val_loss_meter,
-            self.val_metric_manager,
-            include_losses_in_metrics=include_losses_in_metrics,
-        )
+        if self.num_validation_steps is None:
+            val_loss, val_metrics = self._fully_validate_or_test(
```
scarere, line 863: nit: Just curious why you seperated val_by_steps and full_val into two seperate methods? Should we not avoid adding more methods for simplicity?

scarere (approved):
Overall looks good to me. Just some questions about the tradeoff between separating functionality vs reducing the size and number of methods in BasicClient and the codebase as a whole.

fl4health/clients/basic_client.py
emersodb: Yes, I didn't really want to do it this way...however, I actually did this due to your observation that `len(dataloader)` isn't always 100% accurate in terms of how many batches a dataloader actually has in an "epoch." By separating these methods, I push the calculation of when a "full" pass of the validation/test dataloaders is complete onto pytorch instead of us doing it ourselves, which I thought was a little safer.

scarere: Yep that makes a lot of sense and is probably better than opening ourselves up to errors in weird edge cases. Maybe as a future PR we can get around this by creating a wrapper class that takes a dataloader and a fixed number of steps as init arguments.

emersodb: Yeah I think that's a great idea for a refactor in future work

PR #344, opened 2025-02-19
Unrestrict ecos version

scarere:
# Fix
The currently pinned version of `ecos==2.0.7.post1` is not available on windows and leads to the pyproject.toml being unresolvable on windows machines. The package was originally pinned for the `qpth` package which already has several unit tests. Changing the requirement to `ecos = ^2.0.1.post1` currently resolves to `ecos==2.0.14` which is available on windows machines. All tests continue to pass so this should not interfere with `qpth`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1680,33 +1680,30 @@ torch = ">=1.6.0a"
 
 [[package]]
 name = "ecos"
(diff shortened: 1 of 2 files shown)


emersodb (approved):
🎉

PR #345, opened 2025-02-24
[docs] Add missing dataset_path arg in quickstart

nerdai:
# PR Type
Other

# Short Description

Was running through our quickstart example and noticed that the command for starting client is missing `dataset_path` option. At least for me running this command without `--dataset_path` did not work.


emersodb (approved):
👍

PR #349, opened 2025-02-25
Testing Metric Manager Clear

emersodb:
# PR Type

Other

# Short Description

Adding a quick test of the metric manager clear functionality for coverage.

# Tests Added

See description.

### DIFF
--- a/tests/utils/metrics_mangager_test.py
+++ b/tests/utils/metrics_mangager_test.py
@@ -1,0 +1,87 @@
+import pytest
+import torch
+
+from fl4health.utils.metrics import F1, Accuracy, MetricManager
+
+
+def test_metric_manager() -> None:
+    logits1 = torch.Tensor(
+        [
+            [0.8, 0.05, 0.15],
+            [0.88, 0.06, 0.06],
+            [0.1, 0.3, 0.6],
+            [0.4, 0.1, 0.5],
+            [0.1, 0.6, 0.3],
+        ]
+    )
+    target1 = torch.Tensor([0, 0, 2, 0, 2])
+
+    logits2 = torch.Tensor(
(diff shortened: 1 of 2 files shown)


tests/utils/metrics_mangager_test.py
```diff
@@ -0,0 +1,87 @@
...
+    for logits, target in zip(logits_list, target_list):
+        preds = {"prediction": logits}
+        mm.update(preds, target)
+    metrics = mm.compute()
+
+    assert metrics["test - prediction - F1 score"] == pytest.approx(0.80285714285, abs=0.00001)
+    assert metrics["test - prediction - accuracy"] == 0.8
+
+
+def test_metric_manager_clear() -> None:
```
scarere, line 44: Out of complete curiosity, can unit tests have return values or be reused in someway? I just noticed that the second unit test here is an extension of the first unit test, and I wondered if the first one could return the the metric manager so that the second one can call the first unit test to be more concise? Furthermore if it is possible does that break best practices?

scarere: I ask because I'm also working on some unit tests right now that are structured similar to these

scarere (approved):
Looks good to me. I'll just trust you that the logits and targets evaluate to the correct accuracy and F1 that you are testing for.

tests/utils/metrics_mangager_test.py
emersodb: You're actually correct to call me out for not reusing artifacts here. To the best of my knowledge, you tests don't typically return anything. However, they can share global variables or have "fixtures" which are only created once and reused by the tests. I should have done that here. Will clean it up to at least use global variables.

PR #352, opened 2025-03-04
Ema checkpointers

scarere:
# Feature

# Short Description

The scope of this PR expanded beyond just EMA checkpointers and has become quite substantial. The gist is that I made all the metrics way more memory efficient, implemented a new EMAMetric wrapper class as well as new Dice, HardDice and Recall metrics.

# Changelog
- Added EMAMetric wrapper class so that the expected moving average of any scalar metric can be computed.
- Added BestMetric checkpointer class so that checpointing can be done on metrics, including EMA metrics.
- Completely overhauled almost all of the existing metrics in the library:
    - Existing metrics were based on the SImpleMetric base class which accumulated preds and targets in memory, resulting in OOM for anything but small/toy examples.
    - Implemented a new ClassificationMetric base class for metrics that can be computed from 'counts' (tp, fp, fn and tn).
    - The new base class computes the counts and then reduces them over user specified axes before storing them in memory resulting in drastically lower memory usage
    - The ClassificationMetric base class has several generalizable features built in to allow computing metrics along/over different axes of the tensors, binarizing predictions, removing the background class and several options for further memory optimization
    - Reimplemented all the existing metrics in the library (except ROC because its more complex and hence serves as a good example for SimpleMetric) using the ClassificationMetric base class to make them more memory efficient
- Most existing metrics had non-generalizable assumptions about the input tensors. For example, assuming the channel dimension, or expecting preds to be one-hot-encoded but not the targets. Created a utility function that can in nearly all instances resolve any shape mistmatches between predictions and targets by inferring the channel dimensions and which tensor needs to be one hot encoded
- Replaced all the imports and tests with the new Memory efficient metric implementations and removed the old ones. Ensured the new versions could essentially be used as drop in replacements 
    - The one unique exception is accuracy metric when preds and targets are both binary and not one-hot-encoded, then you need to change one argument i think.
    - I set the defaults of the metrics to replicate previous behaviour in most instances, but it might be worth evaluating in the future whether we should make the defaults more intuitive.
- Implemented new Dice, HardDice and Recall metrics

# Tests Added
- Added several tests for the autoaligment function used by the ClassificationMetric base class
- Added like 10 new tests for the Dice and HardDice metrics. Many of these test the underlying capabilities of the ClassificationMetric base class.
- The Recall metric is implicitly tested by the existing test for balanced accuracy since the new BalancedAccuracy metric is a subclass of the Recall metric.
- Replaced all existing metric tests with the new memory efficient versions of the metrics and ensured they all passed.
- All the smoke tests should use the new versions of the metrics so they'll be covered there as well
- nnunet smoke test uses the EMA metric and the new BestMetricCheckpointer so both those should be covered as well.

# Future PR

Created the following ticket related to this PR: https://app.clickup.com/t/868ctjam8

- The metrics module in utilities has grown large enough that it would probably be better to organize is as a seperate subpackage as opposed to being split into multiple modules in the utils subpackage
- Something like the following would probably be good

```
.
|
|--fl4health
|    |--metrics
|         |--__init__.py
|         |--metrics.py
|         |--metric_managers.py
|         |--metric_aggregation.py
|         |--utils.py
```
Furthermore we can probably merge that PR with this one: https://app.clickup.com/t/868ctjd74 which basically is just importing classes from the 'main' file of a subpackage in the `__init__.py` so that we can avoid duplicated or unecessarily long import paths. Eg. for the above proposed metrics subpackage structure, metric classes from `metrics.py` should be imported into the main package to avoid imports with repeated names such as `from fl4health.metrics.metrics import Accuracy`

The two PR's kind of cover similar ground in terms of modifying import paths.


scarere:
I think the reason that the code coverage patches are failing is that I removed the need for the TransformsMetric in the nnunet smoke tests, so that entire class is basically uncovered now. Hopefully adding a unit test will result in enough coverage to get the tests to pass, should be an easy fix

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-class BalancedAccuracy(SimpleMetric):
-    def __init__(self, name: str = "balanced_accuracy"):
+class HardDice(Dice):
+    def __init__(
+        self,
+        name: str = "HardDice",
+        along_axes: Sequence[int] = (0,),
+        ignore_background: int | None = None,
+        zero_division: float | None = None,
+        binarize: float | int | None = None,
```
emersodb, line 823: Since we're expecting this to not be a soft dice metric, should we remove the possibility that binarize is None?

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        ignore_background: bool = False,
+        dtype: torch.dtype = torch.float32,
+    ) -> None:
+        """Balanced accuracy metric for classification tasks. Used for the evaluation of imbalanced datasets.
+
+        Balanced accuracy is defined as the average recall obtained on each class.
+
+        Args:
+            name (str): The name of the metric.
+            channel_dim (int, optional): Index specifying the axis representing the channel dimension. Defaults to 1.
```
emersodb, line 877: It doesn't appear optional here. I'm also confused why this particular metric has a channel dim as expressly provided but others do not?

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+                reducing the memory footprint of metrics that do not use all of the counts in their computation
         """
-        A thin wrapper class to allow transforms to be applied to preds and targets prior to calculating metrics.
-        Transforms are applied in the order given
+        self.name = name
+        self.along_axes = tuple([a for a in along_axes])
+        self.dtype = dtype
+        self.binarize = binarize
+        self.ignore_background = ignore_background
+        self.channel_dim = ignore_background  # If channel dim is None then it will be inferred
```
emersodb, line 456: Perhaps I'm missing something, but this feels a little dangerous setting it to something that isn't related.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+            name=name,
+            along_axes=[channel_dim],
+            dtype=dtype,
+            binarize=binarize_arg,
+            ignore_background=channel_dim if ignore_background else None,
+        )
+
+        # We override the channe dim attribute since this subclass forces it to be known.
+        # Can prevent rare instances where channel dim is unable to be automatically inferred.
+        self.channel_dim = channel_dim
```
emersodb, line 906: I'm a little confused why this option is only available for balanced accuracy or why we're only concerned with failure to infer the channel for this metric? Should we be providing the option to specify a channel dimension to other metrics and just inferring if it isn't provided? That sort of feels like the right thing to do if you're concerned about channel inference edge cases.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        if isinstance(binarize, float):
+            binarize_arg = binarize
+        elif binarize:
+            binarize_arg = int(channel_dim)
+        else:
+            binarize_arg = None
+
+        super().__init__(
+            name=name,
+            along_axes=[channel_dim],
```
emersodb, line 898: I'm probably missing something, but don't we want `along_axes` to correspond to the batch dimension not the channel dimension?

fl4health/utils/metrics.py
```diff
@@ -326,35 +924,74 @@ def __call__(self, logits: torch.Tensor, target: torch.Tensor) -> Scalar:
...
     def __init__(
         self,
-        name: str = "F1 score",
-        average: str | None = "weighted",
-    ):
-        """
-        Computes the F1 score using the ``sklearn f1_score`` function. As such, the values of average correspond to
-        those of that function.
+        name: str = "F1",
+        along_axes: Sequence[int] = (),
```
emersodb, line 931: Should this default to zero, similar to other accuracy? In the tests, there seems to be something different in the way accuracy and F1 use `along_axes`, but its not immediately obvious to me what. In the tests, we're specifying `along_axes` as 0 for accuracy and 1 for F1...that feels weird.

fl4health/utils/metrics.py
```diff
@@ -115,12 +236,118 @@ def clear(self) -> None:
...
+        for transform in self.pred_transforms:
+            pred = transform(pred)
+
+        for transform in self.target_transforms:
+            target = transform(target)
+
+        self.metric.update(pred, target)
+
+    def compute(self, name: str | None) -> Metrics:
+        # Change the name of the original metric class temporarily s
```
emersodb, line 276: Incomplete thought?

emersodb: Definitely needs a bit more documentation. Are we overriding the name of the metric here with the name of the wrapping class? I'm not sure why we need to.

fl4health/utils/metrics.py
```diff
@@ -115,12 +236,118 @@ def clear(self) -> None:
...
+        self.smoothing_factor = smoothing_factor
+        self.previous_score: Metrics | None = None
+        self.name = f"EMA_{self.metric.name}" if name is None else name
+
+    def update(self, input: torch.Tensor, target: torch.Tensor) -> None:
+        return self.metric.update(input, target)
+
+    def compute(self, name: str | None = None) -> Metrics:
+        # Compute current metric score.
+        # Temporarily change name of the underlying metric so that we get the EMAMetric name in keys of metrics_dict
```
emersodb, line 308: Perhaps we should be doing something a little different than overriding? Maybe appending. If feels a little funny.

emersodb: Can we just use the name that we're allowed to inject as part of `.compute`?

fl4health/utils/metrics.py
```diff
@@ -115,12 +236,118 @@ def clear(self) -> None:
...
+            return metrics_dict
+
+        # Otherwise compute EMA score for each 'metric' in Metrics dict
+        for key, current_score in metrics_dict.items():
+            previous_score = self.previous_score[key]
+            if (
+                not isinstance(current_score, str)
+                and not isinstance(current_score, bytes)
+                and not isinstance(previous_score, str)
+                and not isinstance(previous_score, bytes)
```
emersodb, line 326: I know why you have to do this, but yuck 😂

fl4health/utils/metrics.py
```diff
@@ -115,12 +236,118 @@ def clear(self) -> None:
...
     def __init__(self, name: str) -> None:
         """
         Abstract metric class with base functionality to update, compute and clear metrics. User needs to define
         ``__call__`` method which returns metric given inputs and target.
 
+        WARNING: This class accumulates all the predictions and targets in memory throughout each FL server round. This
+        may lead to out of memory (OOM) issues. Subclassing SimpleMetric is recommended only for quickly prototyping
+        new metrics that have existing implementations in other packages. See the ROC_AUC class for an example.
+        In other scenarious it is generally recommended to subclass the base Metric class instead and implement a
+        custom update method that reduces the memory footprint of the metric. See the Accuracy clas for an example.
```
emersodb, line 349: clas -> class

tests/smoke_tests/nnunet_config_2d.yaml
```diff
@@ -1,8 +1,9 @@
 # Parameters that describe the server
-n_server_rounds: 1
+n_server_rounds: 2
 
 # Parameters that describe the clients
 n_clients: 1
 local_epochs: 1
 
 nnunet_config: 2d
+nnunet_plans: "tests/smoke_tests/msd004_nnunet_plans.json"
```
emersodb, line 9: Is this being added to have one of the smokes test using a server provided plan? I think that's great. Just want to make sure I understand why we'd want the change.

tests/utils/metric_test.py
```diff
@@ -1,31 +1,331 @@
...
+    t = torch.tensor([[0, 0, 1, 0, 1, 1, 1, 1]])
+    hd.update(p, t)
+    result = hd.compute()
+    assert result["DICE"] == approx(0.5)
+
+    p = torch.tensor([[0, 0, 0, 1, 1, 1, 1, 1]])
+    t = torch.tensor([[0, 0, 1, 0, 1, 1, 1, 1]])
+    hd.update(p, t)
+    result = hd.compute()
+    assert result["DICE"] == approx(0.65)  # Avg of 0.8 and 0.5
```
emersodb, line 179: Are we going to run into issues if the batch sizes in accumulate are different sizes? That is, when we accumulate three batches, two of size 32 and the last one is incomplete with size 17 or something? This can happen at the end of an epoch if the dataloaders don't drop incomplete batches.

emersodb (commented):
Overall, I think there is a ton of really awesome refactors in here and I think the metrics will be way better when we get this sorted out. I think my biggest is making sure we're consistent in a few areas. 

For example, it seems like BalancedAccuracy might be an outlier in terms of behavior for some reason.

I'm also a little confused in some settings for the specification of the `along_axes`. For accuracy and Dice, the batch dimension seems important but the F1 metrics seems to need the `along_axis` to be the channel dimension and I'm not sure I get why.

Either way, some really nice work. Perhaps just some kinks to work through.

tests/utils/snapshotter_test.py
```diff
@@ -138,7 +138,7 @@ def test_reports_manager_snapshotter() -> None:
 
 
 def test_metric_manager_snapshotter() -> None:
-    metrics = [Accuracy("accuracy")]
+    metrics = [ROC_AUC()]
```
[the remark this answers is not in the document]
scarere: The snapshotter tests the metric manager by checking that the targets and predictions are being accumulated. Since I replaced the accuracy metric with one that does not accumulate preds and targets, I switched the metric to ROC_AUC which I did not change in this PR. The ROC_AUC metric is metric that I left the same as an example of subclassing SimpleMetric, mainly because it would have been the most complicated to implement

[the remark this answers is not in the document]
emersodb: Makes sense 🙂

tests/smoke_tests/nnunet_config_2d.yaml
```diff
@@ -1,8 +1,9 @@
 # Parameters that describe the server
-n_server_rounds: 1
+n_server_rounds: 2
```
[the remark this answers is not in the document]
scarere: Not sure but I wasn't getting timeout errors with the smoke tests on this PR so I'm assuming its fine. On my local machine they run in like 1 minute but I have the 4090 so... For the 3d example we're running with local_steps=5 so we should be good for that one.

examples/nnunet_example/client.py
```diff
@@ -1,30 +1,21 @@
 import argparse
```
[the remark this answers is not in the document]
scarere: But those only popped up after I started looking for it due to the smoke test failures

fl4health/checkpointing/checkpointer.py
```diff
@@ -79,12 +80,15 @@ def __init__(
...
                 metrics and produces a score based on these.
+            checkpoint_score_name (str | None, optional): Name of the score produced by the scoring function. This is
+                used for logging purposes. If not provided, the name of the function will be used. Defaults to None.
             maximize (bool, optional): Specifies whether we're trying to minimize or maximize the score produced
                 by the scoring function. Defaults to False.
         """
         super().__init__(checkpoint_dir, checkpoint_name)
         self.best_score: float | None = None
         self.checkpoint_score_function = checkpoint_score_function
+        self.checkpoint_score_name = checkpoint_score_name or checkpoint_score_function.__name__
```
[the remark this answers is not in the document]
scarere: How about "Unknown"?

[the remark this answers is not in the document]
scarere: I feel like that is more informative than None. Plus I can add a warning. Or we can leave it as the function name and just add the warning.

fl4health/checkpointing/checkpointer.py
```diff
@@ -131,20 +135,20 @@ def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: dict[str, Sca
...
             log(
                 INFO,
-                f"Checkpointing the model: Current score ({comparison_score}) "
+                f"Checkpointing the model: Current {self.checkpoint_score_name} score ({comparison_score}) "
                 f"{self.comparison_str} Best score ({self.best_score})",
             )
             self.best_score = comparison_score
             try:
-                log(INFO, f"Saving checkpoint as {str(self.checkpoint_path)}")
+                log(WARNING, f"Saving checkpoint as {str(self.checkpoint_path)}")
```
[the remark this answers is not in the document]
scarere: My bad must have changed that for debugging purposes

fl4health/checkpointing/checkpointer.py
```diff
@@ -252,6 +256,41 @@ def maybe_checkpoint(self, model: nn.Module, loss: float, metrics: dict[str, Sca
...
+        """Checkpointer that checkpoints based on the value of a user defined metric.
+
+        Args:
+            checkpoint_dir (str): Directory to which the model is saved. This directory should already exist. The
+                checkpointer will not create it if it does not.
+            checkpoint_name (str): Name of the checkpoint to be saved.
+            metric (str): The name of the metric to base checkpointing on. After prepending the prefix, should be a
+                key in the metrics dictionary passed in self.maybe_checkpoint. In BasicClient this is the 'name'
+                attribute of the corresponding fl4health.utils.metrics.Metric that was provided to the clients.
+            prefix (str, optional): A prefix to add to the metric name to create the key used to find the metric.
```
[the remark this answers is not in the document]
scarere: Yeah the main idea was to just be able to use the default prefix. Honestly the one downside to this approach is that the prefix is sort of determined deep in the client code, so its not really visible to the user until the run an experiment and find what they're looking for.

fl4health/checkpointing/checkpointer.py
```diff
@@ -165,7 +169,7 @@ def __init__(self, checkpoint_dir: str, checkpoint_name: str) -> None:
         def null_score_function(loss: float, _: dict[str, Scalar]) -> float:
             return 0.0
 
-        super().__init__(checkpoint_dir, checkpoint_name, null_score_function, False)
+        super().__init__(checkpoint_dir, checkpoint_name, null_score_function, "Latest", False)
```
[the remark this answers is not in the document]
scarere: sure

fl4health/utils/nnunet_utils.py
```diff
@@ -173,7 +173,9 @@ def convert_deep_supervision_dict_to_list(tensor_dict: dict[str, torch.Tensor])
 
 def get_segs_from_probs(preds: torch.Tensor, has_regions: bool = False, threshold: float = 0.5) -> torch.Tensor:
     """
-    Converts the nnunet model output probabilities to predicted segmentations
```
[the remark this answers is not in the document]
scarere: I don't think its used anywhere since I basically baked it into the classification metric class, but I think its useful to keep around for now, especially since its relegated to the nnunet_utils which isn't overcrowded at the moment

research/cifar10/adaptive_pfl/ditto/client.py
```diff
@@ -160,9 +160,9 @@ def get_model(self, config: Config) -> nn.Module:
     client = CifarDittoClient(
         data_path=data_path,
         metrics=[
-            Accuracy("accuracy"),
-            F1("f1_score_macro", average="macro"),
-            F1("f1_score_weight", average="weighted"),
+            Accuracy("accuracy", along_axes=[0], binarize=1, exact_match=True),
+            F1("f1_score_macro", along_axes=[1], binarize=1, weighted=False),
```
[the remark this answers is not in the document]
scarere: Almost no classification metrics are actually computed *along* the batch dimension, they are generally computed *over* the batch dimension. The default for sklearn is to compute exact match accuracy. In order to do this they require that the inputs has exactly 1 dimension and is not one-hot-encoded. By setting along_axes to 0 and exact_match to True, we get the same behaviour, but it is generalizable to both one-hot-encoded and not one-hot-encoded tensors as well as tensors with any number of dimension.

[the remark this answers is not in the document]
scarere: I tried to make it clear how the along_axes argument works, but maybe didn't do enough to relate it to how sklearn computes metrics.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+                accuracy scores must be stored in memory until cleared, this may cause memory build up in some
+                instances. Default is to compute along the first dimension which is assumed to be the batch/n_samples
+                dimension.
+            binarize (float | int | None, optional): A float for thresholding values or an integer specifying the
+                index of the channel/class dimension. If a float is given, predictions below the threshold are mapped
+                to 0 and above are mapped to 1. If an integer is given, predictions are binarized based on the class
+                with the highest prediction. If None leaves preds unchanged. Default is 1 since this is usually the
+                channel dimension.
+            exact_match (bool): If True computes the 'Subset Accuracy'/'Exact Match Ratio'. Individual accuracies that
+                are not prefect/exact (== 1) are set to 0 before computing the final accuracy score. This is useful for
```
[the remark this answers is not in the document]
scarere: I'll update the argument description to maybe make it clear on when you would want to turn it off

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
 TEST_LOSS_KEY = f"{MetricPrefix.TEST_PREFIX.value} checkpoint"
 
 
+def infer_channel_dim(tensor1: torch.Tensor, tensor2: torch.Tensor) -> int:
+    """Infers the channel dimension given two related tensors of different shapes.
+
+    Generally useful for inferring the channel dimension when one tensor is one-hot-encoded and the other is not. The
+    channel dimension is inferred by looking for dimensions that either are not the same size, or are not present int
+    tensor 2. If a dimension adjacent to the
```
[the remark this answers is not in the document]
scarere: Good catch. Removed the incomplete sentence and described an example for the assertion error

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
...
+    """
+    Attempts to correct shape mismatches between the given tensors by inferring which one to one-hot-encode.
+
+    If both input tensors are not one-hot-encoded, then they are returned unchanged. If one is one-hot-encoded but not
+    the other, then both are returned as one-hot-encoded tensors.
+
+    Args:
+        preds (torch.Tensor): The tensor with model predictions.
+        targets (torch.Tensor): The tensor with model targets.
+        channel_dim (int | None): Index of the channel dimension. If left as None then this method attempts to infer
```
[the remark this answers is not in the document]
scarere: You're right I've been using them interchangeably, i'll defer to using class

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
...
+            "A dimension adjacent to the channel dimension appears to have the same size."
+        )
+
+    # If tensors have same ndim but diff shape, then this only works if channel dim was empty for one of them
+    if tensor1.ndim == tensor2.ndim:
+        assert (tensor1.shape[ch] == 1) or (tensor2.shape[ch]) == 1, (
+            f"Could not infer the channel dimension of tensors with shapes: ({tensor1.shape}), ({tensor2.shape}). "
+            "The inferred candidate dimension has different sizes on each tensor, was expecting one to be empty."
+        )
+    return ch
```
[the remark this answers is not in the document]
scarere: Yeah basically the two cases I want to cover are `t1.shape==(..., N, C, x, y, z, ...) and t2.shape==(..., N, 1, x, y, z, ...)` and `t1.shape==(..., N, C, x, y, z, ...) and t2.shape==(..., N, x, y, z, ...)`. It's super intuitive to infer the channel dimension in these scenarious for a human, but involves covering a few edge cases to get an algorithm to infer it. is there anything specific about the function you find unclear or hard to debug? Maybe I could think of a way to improve it.

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
...
+    # Determine channel dimension. This method also has a bunch of necessary assertions.
+    ch = infer_channel_dim(tensor1, tensor2) if channel_dim is None else channel_dim
+
+    assert (
+        tensor1.ndim - tensor2.ndim
+    ) <= 1, f"Can not align pred and target tensors with shapes {preds.shape}, {targets.shape}"
+
+    # Add channel dimension if there isn't one
+    if tensor1.ndim != tensor2.ndim:
+        tensor2 = tensor2.view((*tensor2.shape[:ch], 1, *tensor2.shape[ch:]))
```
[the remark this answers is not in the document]
scarere: I was using nnunet as a reference and thats how they did it. I'll replace it with the unsqueeze function which seems more appropriate

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
...
+
+    # Swap tensors on off chance that the first one had an empty dim and the second didn't
+    if tensor1.shape[ch] < tensor2.shape[ch]:
+        tensor1, tensor2 = tensor2, tensor1
+        swapped = not swapped
+
+    # One-hot-encode tensor2. We know at this point that it has an empty dim along channel axis
+    if torch.is_floating_point(tensor2) and torch.frac(tensor2).sum() != 0:
+        # If tensor is continious it must be binary classification and elements must be probabilities in range [0, 1].
+        t2_ohe = torch.cat([1 - tensor2, tensor2], dim=ch)
```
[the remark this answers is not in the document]
scarere: I've been defining 'one-hot-encoding' as having a class dimension with seperate channels in that dimension for each class. So in my mind this includes softmax tensors which have continious predictions for each class that sum to 1. Is one-hot-encoded generally assumed to be 'hard' predictions only?

fl4health/utils/metrics.py
```diff
@@ -21,6 +23,125 @@ class MetricPrefix(Enum):
...
+    else:
+        # If tensor2 is not continious then it must contain class labels. One hot encode the tensor.
+        t2_ohe = torch.zeros(tensor1.shape, device=tensor1.device)
+        t2_ohe.scatter_(ch, tensor2.to(torch.int64), 1)
+
+    # Return modified tensors in their original positions.
+    return (t2_ohe, tensor1) if swapped else (tensor1, t2_ohe)
```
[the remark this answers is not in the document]
scarere: I think I tested all the cases with: `test_multiclass_align`, `test_binary_align` and `test_align_exceptions`. In the latter I make sure that exceptions are raised for the more unique corner cases.

tests/utils/metric_test.py
```diff
@@ -1,31 +1,331 @@
...
-    accuracy_metric = Accuracy()
+    accuracy_metric = Accuracy(along_axes=[0], exact_match=True)
 
-    pred1 = torch.eye(5)
-    target1 = torch.arange(0, 5)
-    accuracy1 = accuracy_metric(pred1, target1)
-    assert accuracy1 == 1.0
+    pred1 = one_hot(torch.arange(0, 6) % 5)
+    target1 = torch.tensor([0, 1, 2, 3, 4, 0])
+    accuracy_metric.update(pred1, target1)
```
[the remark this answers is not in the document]
scarere: I was just recreating the old test which didn't test for that either. I assumed that we weren't testing for it before since ithe accumulation behaviour was inherited by SimpleMetric which was tested in other test cases. Similarly, here the accumulation is inherited from the ClassificationMetric which I believe is tested with the dice metrics. However its minimal effort to add this to the unit test so I have no problem adding it in

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        targets passed to update method have different shapes, this class will attempt to infer the channel dimension
+        and align the shapes by one-hot-encoding one of the tensors.
+
+        Args:
+            name (str): The name of the metric.
+            along_axes (Sequence[int]): Sequence of indices specifying axes *along* which to accumulate tp, fp, fn and
+                tn. The counts will be summed *over* the axes not specified. The 0th axis is assumed to be the batch or
+                sample dimension. If provided an empty sequence, then the counts are scalar values computed *over* all
+                axes.
+            dtype (torch.dtype): The dtype to store the counts as. If preds or targets can be continious, specify a
```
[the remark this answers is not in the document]
scarere: My logic was that if the default is a float, then metrics will work in both the 'soft' and 'hard' setting. Setting the dtype to an integer then serves as more of an preventative optimisation to reduce the risk of overflow when you can guarantee that your predictions will be integers as well. If we were to make the default an integer, then I could see a scenario where someone passes in 'soft' preds expecting it to work, and funnily enough this wouldn't raise an error, but the output would technically be wrong.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-                apply to the model predictions before computing the metrics. Each callable must accept and return a
-                ``torch.Tensor``. Use partial to set other arguments.
-            target_transforms (Sequence[TorchTransformFunction] | None, optional): A list of transform functions to
-                apply to the targets before computing the metrics. Each callable must accept and return a
-                ``torch.Tensor``. Use partial to set other arguments.
+            input (torch.Tensor): The tensor to binarize.
+            binarize (float | int, optional): A float for thresholding values or an integer specifying the
+                index of the channel/class dimension. If a float is given, elements below the threshold are mapped
+                to 0 and above are mapped to 1. If an integer is given, elements are binarized based on the class
+                with the highest prediction. If binarize is None then the input is returned unchanged
```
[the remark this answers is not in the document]
scarere: Very good catch

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        sum_axes = tuple([i for i in range(preds.ndim) if i not in self.along_axes])
+
+        # Compute counts. If were ignoring a count, set it as an empty tensor to avoid downstream errors.
+        tp = (preds * targets) if not self.discard_tp else torch.tensor([])
+        fp = (preds * (1 - targets)) if not self.discard_fp else torch.tensor([])
+        fn = ((1 - preds) * targets) if not self.discard_fn else torch.tensor([])
+        tn = ((1 - preds) * (1 - targets)) if not self.discard_tn else torch.tensor([])
+
+        # Sum along specified axes only if there are axes to sum over.
+        tp = tp.sum(sum_axes, dtype=self.dtype) if len(sum_axes) > 0 and not self.discard_tp else tp
```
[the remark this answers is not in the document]
scarere: One example is 1D exact_match accuracy. We would compute accuracy along dimension 0. So essentially each individual accuracy for each sample is 0 or 1. Then we return the average of all the accuracy values for each sample which just happens to also be the accuracy.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+        # Get tp, fp, fn and tn counts for current update. If a count is discarded, then an empty tensor is returned.
+        tp, fp, fn, tn = self.count_tp_fp_fn_tn(preds, targets)
+
+        # If this is first update since init or clear, initialize counts attributes and exit function
+        if not self.counts_initialized:
+            self.tp, self.fp, self.fn, self.tn = tp, fp, fn, tn
+            self.counts_initialized = True
+            return
+
+        # If the batch/sample dimension is in self.along_axes, we must concatenate the values; otherwise, we sum them
```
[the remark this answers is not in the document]
scarere: I think I explained this in a previous comment, but essentially the shape of our tp/fp/fn/tn counts will be whatever leftover after aggregating over axes not specified in `along_axes`. If we are computing along the 0th batch/sample dimension, then we want an individual value for each sample and must therefore concatenate the new counts. If we are computing 'over' the batch/sample dimension (as in it is not specified in along_axes) then we want to aggregate the counts across all samples and therefore we sum the counts. We don't have to take this into consideration for other axes since the batch/sample dimension is the only one with a dynamic size that depends on the number of updates.

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-            y_pred = (logits > self.logits_threshold).int()
+    def compute_from_counts(self, tp: torch.Tensor, fp: torch.Tensor, fn: torch.Tensor, tn: torch.Tensor) -> Metrics:
+        # Compute individual accuracy scores. Div by 0 shouldn't be possible here.
+        accuracy = (tp + tn) / (tp + fp + fn + tn)
+        # If exact match is true, set all scores below 1 to 0.
+        accuracy = (accuracy == 1).to(torch.float) if self.exact_match else accuracy
+        return {self.name: torch.mean(accuracy).item()}
+
+
+class Recall(ClassificationMetric):
```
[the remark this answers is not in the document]
scarere: Not explicitly, but it is implicitly tested since BalancedAccuracy is basically a special case of Recall.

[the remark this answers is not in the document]
scarere: Definitely an argument to be made to add some explicit tests though

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+    def __init__(
+        self,
+        name: str = "Recall",
+        along_axes: Sequence[int] = (),
```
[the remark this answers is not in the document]
scarere: In truth I feel like the default should be an empty tuple for `along_axes` which would lead to a single value being computed over all dimensions. Thats why I have that as the default for the Recall metric. However for the other metrics where I was replacing an existing one, I went against this idea in order to make the default behaviour as similar to what we had before as an attempt to reduce the descriptiveness of this PR. It might be tedious to go through the entire library and modify all the instances where metrics are used to align with what I felt the default behaviour should be, so I was trying to defer that to a future PR. It doesn't really make sense to compute recall individually for each sample unless you have images or higher dimensional data. For accuracy as I mentioned before, were not really computing accuracy for each sample, we're merely using that as a trick for implementing exact_match accuracy

fl4health/utils/metrics.py
```diff
@@ -181,131 +407,503 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
-        dice = intersection / (union + self.epsilon)
-        # If both inputs are empty the dice coefficient should be equal 1
-        dice[union == 0] = 1
-        return torch.mean(dice).item()
+            tp[denominator == 0] = self.zero_division
+            denominator[denominator == 0] = 1
 
+        # Compute recall scores and return mean
+        recall = tp / denominator
+        return {self.name: torch.mean(recall).item()}
```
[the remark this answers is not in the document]
scarere: yeah I agree that this is maybe a con to doing things this way. We should brainstorm ways to ensure that the result is the same regardless of whether or not the inputs are one-hot-encoded without sacrificing the generalizability

fl4health/utils/metrics.py
scarere: My logic was that if your outputs were already binarized in some way (say by transforms metric, or something similar) then you might want to set this to None.

fl4health/utils/metrics.py
scarere: You're right it shouldn't be optional. The reason Balanced accuracy has the class dim expressely provided is that it can not be computed without expressely knowing the class dim. Balanced accuracy is the average recall for each class, so you can't compute it without knowing the class dimension. I guess the one exception would be if your data was not one-hot-encoded, then you could technically figure it out. This class might have some room to be further generalized.

scarere: So we assume that all preds and targets are in range [0,1] for the classification metric class. For the other metrics we allow computing them on binary preds and targets that are not one-hot-encoded, otherwise an error should be thrown. For recall I guess we enforce that they must be one-hot-encoded as there is no way to know whether the inputs are binary classification not OHE or multiclass OHE

fl4health/utils/metrics.py
scarere: Yeah so the idea is that we may not always need to infer the channel, namely for binary classification there might not be a channel dimension. Though as you brought up in a previous comment, this might results in metric values being different depending on whether data is OHE. Balanced accuracy I was enforcing it because we need to compute the recall values along the channel dimensions

fl4health/utils/metrics.py
scarere: Balanced accuracy is defined as the average recall for each class. Hence by computing recall along the class dim, we will compute a single recall score for each class and then take the average

fl4health/utils/metrics.py
scarere: Might be better for us to meet to discuss the along_axes argument

fl4health/utils/metrics.py
scarere: The reason we need to override the name of the metric is so that the right key is returned. Eg. We want EMA_Dice instead of Dice to be the metric name. The reason we can't just replace the key with f"{name} - {self.name}" is that there might be multiple metrics in the metrics dict in which case the user would have to override the compute method of the wrapper class to define a new key for each metric. The more robust way of doing this is replacing the name of the self.metric object so that the EMA prefix is added wherever self.name is used.

fl4health/utils/metrics.py
scarere: That might be a better solution is to add EMA to the name argument as opposed to overriding the attribute. The danger here is we can't guarantee how the user will connect the name argument to self.name in the key. Since the name of the metric is now EMA_XXX we need to ensure that that is used in place of the actual metric name wherever it might be used.

fl4health/utils/metrics.py
scarere: I can make this more concise

tests/smoke_tests/nnunet_config_2d.yaml
scarere: Yes that's exactly why I added it. John introduced some breaking changes before the break that passed the smoke tests because they only arose when the plans file was not specified. I wanted to cover this in the tests to prevent future breaking changes

tests/utils/metric_test.py
scarere: No I don't think so. When 0 is in along_axes, batches are concatenated, so any shape should work. When 0 is not in along_axes, the batch dimension is summed over, and therefore the shape doesn't matter

fl4health/utils/metrics.py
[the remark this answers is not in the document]
emersodb: I think this dissonance only exists for binary settings. It looks like in scikit learn, they have a "binary" option that you can specify for how you want to "aggregate" multi-class scores. We should probably implement something like that, where we throw an error if there are more than two label classes. They also have an option to specify the pos_label that is used in this regard. https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html#sklearn.metrics.recall_score

fl4health/utils/metrics.py
emersodb: I see...essentially, we're taking the specification of the along axis out of the users hands here, because any other specification doesn't make sense?

fl4health/utils/metrics.py
emersodb: Yes. Now that I've taken a deeper dive on the explicit way things are calculated I have a better sense for this.

fl4health/utils/metrics.py
emersodb: hmmm...I see. Very tricky. I get the need though. Perhaps we can just finish the comment thought then.

fl4health/utils/metrics.py
emersodb: Yeah I see...let's leave it for now.

fl4health/utils/metrics.py
emersodb: You did a good job.

emersodb (commented):
Alright...took a fairly thorough second pass through and left a few additional small comments. 

I think we're getting closer to the right spot, but I'm still concerned about the complexity of the options that we're providing to the user. It feels like it would be pretty easy for someone to accidentally compute the wrong value, but for it to also _appear_ reasonable if they didn't look to hard. A prime example of that is multi-class accuracy with along_axes=(0, ) or (1,). They are both going to look fairly plausible, but they are also very different.

We also still have an issue where the same recall can be quite different depending on whether you present one-hot or not preds and labels. Scikit learn has some ways of dealing with this that we probably should adopt.

nerdai (commented):
I just realized we don't have a CHANGELOG.md. I guess we're more relient on the release notes.

nerdai (commented):
code coverage looks to be failing. I see that we have added a good amount of tests, but our coverage is still too low. Can we add some tests for the missing lines?

nerdai (commented):
This is a pretty beefy PR .  Took a look at `EMAMetric` for now to start and left some comments. I'll keep the comments coming for `ClassificationMetric` as well.

fl4health/utils/metrics.py
emersodb: I think there are some pros and cons to this that we want to think about in the sense that having this as a mode might add additional functionality that maybe metric shouldn't "know about" by default. For each kind of mode (geometric mean etc.) we'd possibly have hyper-parameters that need to be set and more.

fl4health/utils/metrics.py
```diff
@@ -181,131 +415,508 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
+            binarize=binarize_arg,
+            ignore_background=label_dim if ignore_background else None,
+        )
+
+        # We override the label dim attribute since this subclass forces it to be known.
+        # Can prevent rare instances where label dim is unable to be automatically inferred.
+        self.label_dim = label_dim
 
 
 class ROC_AUC(SimpleMetric):
```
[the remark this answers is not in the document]
emersodb: I think the goal is to have a base class to allow for onboarding of other metrics that aren't easily computed in an accumulative fashion. `ROC_AUC` is just one example, but it feels like there are others. So I think it makes sense to have the parent?

For the second question, I actually don't know. I think the idea is to force definition of the function. So it seems like the right thing to do?

fl4health/utils/metrics.py
scarere: Yes exactly. Balanced accuracy must be computed along the class dimension. So i felt implementing it this way made the arguments simpler and reduced the chance of a user using the metric in the wrong way

fl4health/utils/metrics.py
scarere: Yeah I think the pro of adding it to the base class is the the interface for the user is simpler as they don't need to use the wrapper class. As david mentioned however, the con is that we're adding functionality to the base class that for most users isn't necessary. For me personally, I'd say that EMA feels conceptually like a subclass of Metric since it adds on to it and is derivative of it,

fl4health/utils/metrics.py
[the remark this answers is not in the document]
scarere: I second the answer to the first question. Implementing a metric using the classification metric base class is more involved and requires getting into the weeds a bit. The reason to do so is that you end up with an overall better, more generalisation and memory efficient implementation. However for quick prototyping of new metrics simple metric offers an easier interface/way to add new metrics

fl4health/utils/metrics.py
```diff
@@ -19,6 +19,13 @@ class MetricPrefix(Enum):
     VAL_PREFIX = "val -"
 
 
+class MetricOutcome(Enum):
```
scarere, line 22: nit: I don't know if the name of this enum class matches its purpose. The term "outcome" is confusing to me here. I think ideally this should clarify that its specifically for classification, and that it a type of "count". (Not sure if count is the right word). Also maybe it makes sense to reference the term "confusion" matrix somehow. ConfusionCount? Or ClassificationCount?

PR #356, opened 2025-03-10
Fixing an issue with the test_client_level_dp_breast_cancer smoke

emersodb:
# PR Type
Fix

# Short Description

The `test_client_level_dp_breast_cancer` smoke test has been flaky for a long time. I was finally able to carve out some time to investigate. It turns out that the flakiness was, indeed, a subtle bug associated with the way we implemented the strategy when we get unlucky with Poisson sampling.

The gist is as follows:
1. The DP mechanism relies on Poisson sampling clients to participate in each round of FL. For the smoke test, we had a Poisson probability of 2/3 with two clients participating. 
2. If, on the **first round**, we got unlucky and sampled no clients for the `fit()` call, we have an optimization in our base server that will skip doing anything, including aggregate fit.
3. However, if we were doing `weighted_aggregation` and `per_client_example_cap` had not been provided, we're supposed to compute that value (along with `total_client_weight`). If we get unlucky (as in (2)) then we never do this and it was never computed thereafter (due to the `server_round == 1` condition). When we go to `aggregate_fit` in a second round, `self.per_client_example_cap` is still None and we trip the assert on Line 337 of `client_dp_fedavgm.py`. This takes down the server, orphans the clients, and triggers the timeout that we were seeing. 🤦 

The fix is simple, if we're doing `weighted_aggregation` then we just compute `per_client_example_cap` if it is None (regardless of what round) and then don't do it again thereafter.

# Tests Added

None. The smokes will (hopefully) no longer fail on this test intermittently. I tested this fix in two ways

1. I forced the conditions above to happen and verified that it doesn't happen with the fix below and that everything functions as expected.
2. I range the smoke test in a loop 100 times and it didn't fail at any point.

### DIFF
--- a/fl4health/strategies/client_dp_fedavgm.py
+++ b/fl4health/strategies/client_dp_fedavgm.py
@@ -308,8 +308,9 @@ class ClientLevelDPFedAvgM(BasicFedAvg):
         if not self.accept_failures and failures:
             return None, {}
 
-        # If first round compute total expected client weight
-        if self.weighted_aggregation and server_round == 1:
+        # If we're doing weighted aggregation, we need to compute a per_client_example_cap. If it has been provided
+        # manually, we skip this step. Otherwise, we perform the necessary calculations.
+        if self.weighted_aggregation and self.per_client_example_cap is None:
             assert self.sample_counts is not None
 
             total_samples = sum(self.sample_counts)
--- a/tests/smoke_tests/test_smoke_tests.py
+++ b/tests/smoke_tests/test_smoke_tests.py
@@ -5,7 +5,6 @@ from pathlib import Path
 import pytest
 
 from .run_smoke_test import (
-    SmokeTestTimeoutError,
     load_metrics_from_file,
     run_fault_tolerance_smoke_test,
     run_smoke_test,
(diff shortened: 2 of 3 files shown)


nerdai:
One way to test this is to trigger smoke tests manually using CLI.

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -5,16 +5,15 @@
...
     load_metrics_from_file,
     run_fault_tolerance_smoke_test,
     run_smoke_test,
 )
 
-# skip some tests that currently fail if running locallly
+# skip some tests that currently fail if running locally
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
 
-# Marks all test coroutineutines in this module
```
nerdai, line 17: lol what is this jargon. I'll take full responsibility for this nonsense.

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -185,35 +184,16 @@ async def test_client_level_dp_cifar(tolerance: float) -> None:
 
 @pytest.mark.smoketest
 async def test_client_level_dp_breast_cancer(tolerance: float) -> None:
-
-    num_attempts = 3
```
nerdai, line 189: Yay. Getting rid of this nasty stuff!

fl4health/strategies/client_dp_fedavgm.py
```diff
@@ -308,8 +308,9 @@ def aggregate_fit(
         if not self.accept_failures and failures:
             return None, {}
 
-        # If first round compute total expected client weight
-        if self.weighted_aggregation and server_round == 1:
+        # If we're doing weighted aggregation, we need to compute a per_client_example_cap. If it has been provided
+        # manually, we skip this step. Otherwise, we perform the necessary calculations.
+        if self.weighted_aggregation and self.per_client_example_cap is None:
```
nerdai, line 313: Sorry what does this per client example cap represent and how come it replaces server round condition?

nerdai (commented):
LGTM. Just left you a clarifying question and celebratory comments.

fl4health/strategies/client_dp_fedavgm.py
emersodb: This is a necessary variable for computing weighted aggregation in this setting. The if statement was basically saying just do the computation on the first server round. However, due to the conditions listed in the PR description, if we got unlucky, we wouldn't enter this function on the first round. So the quantity would not be computed and would blow up the server on the second round. 

Since what we care about is whether this quantity is None (not the server round specifically) in order to trigger computing it (rather than it being provided by the user), we just test for that instead.

nerdai: Ahhh got it. Thanks for this thorough explanation! I must've missed this completely in my reading of your PR message. My apologies.

nerdai (approved):

PR #359, opened 2025-03-18
Ft/ag bert example

fatemetkl:
# PR Type
Example

# Short Description
A simple example of funetuning BERT on AG News classification task.

### DIFF
--- a/examples/bert_finetuning_example/README.md
+++ b/examples/bert_finetuning_example/README.md
@@ -1,0 +1,55 @@
+# BERT fine-tuning example
+In this example, a pre-trained BERT encoder is fine-tuned on the AG News classification task using the FedAvg strategy. The AG News dataset is a collection of more than 1 million news articles classified into four categories: “World”, “Sports”, “Business”, and “Sci/Tech”. Each client performs Dirichlet sampling on the AG News training set, resulting in heterogeneous data distributions across clients. Server-side checkpointing is used to save the latest as well as the best server model.
+This example is a simplified version of the research code in `research/ag_news`, where partial parameter exchange techniques are explored.
+
+## Dataset pre-processing
+`client_data.py` handles all data preprocessing steps and is responsible for constructing the dataloaders for training and evaluation.
+Since we are using a pre-trained BERT model, its corresponding tokenizer and vocabulary (all pre-trained) are used for processing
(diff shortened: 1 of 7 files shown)


examples/bert_finetuning_example/README.md
```diff
@@ -0,0 +1,54 @@
...
+* `checkpoint_path`: path to save the server's checkpoints
+* `sample_percentage`: the proportion of the training data that will be sampled for each client. Should be between 0 and 1 inclusive.
+* `beta`: a positive real number that controls the heterogeneity of the data distributions among clients. The smaller beta is, the more heterogeneous the distributions are.
+
+## Starting Clients
+
+Once the server has started and logged "FL starting," the next step, in separate terminals, is to start the two
+clients. This is done by simply running (remembering to activate your environment)
+```
+python -m examples.bert_finetuning_example.client --dataset_path /path/to/data
```
emersodb, line 36: It looks like one of the required arguments is `client_number`. So I think we want to put that in this command as well and very briefly discuss it here?

examples/bert_finetuning_example/server.py
```diff
@@ -0,0 +1,123 @@
...
+
+    server = FlServer(
+        client_manager=SimpleClientManager(),
+        fl_config=config,
+        strategy=strategy,
+        checkpoint_and_state_module=checkpoint_and_state_module,
+        accept_failures=False,
+    )
+
+    # grpc_max_message_length is reset here so the entire model can be exchanged between the server and clients.
```
emersodb, line 94: Maybe repeat the comment about how the lengths need to be the same between the clients and server?

emersodb (commented):
Overall, looks great. Just a few small comments 🙂

examples/bert_finetuning_example/README.md
```diff
@@ -0,0 +1,54 @@
...
+To launch the server and clients, run:
+
+```bash
+sbatch examples/bert_finetuning_example/run.sh \
+  path_to_config.yaml \
+  path_to_folder_for_artifacts/ \
+  path_to_folder_for_dataset/ \
+  path_to_desired_venv/ \
+  client_learning_rate_value \
+  server_address
```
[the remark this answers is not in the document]
fatemetkl: Yes, I tested the example with this script and it works fine!

examples/bert_finetuning_example/README.md
fatemetkl: I initially set this argument for client-side checkpointing, but I later removed client-side checkpointing to keep the example simple. So, `client_number` is not used anywhere now.
I am removing this argument from the client code and `run.sh` scrip if that's alright.

emersodb (approved):
Good to go!

PR #360, opened 2025-03-24
Some small fixes for W and B integration

emersodb:
# PR Type
Fix

# Short Description

Our example of Weights and Biases integration/logging was a bit broken by some refactors that we put into place. As such, I've taken a pass at fixing them properly. We used to create the client-side logging using the config that the server sends to the clients. However, this has become cumbersome and a bit convoluted to implemented with how clients are instantiated and initialized. So I've migrated the example to use client-side W and B reporter construction and gotten rid of the configuration component.

# Tests Added

I ran the FedProx example with W and B logging and it worked great. These are kind of hard to test automatically but I'm open to suggestions 🙂


examples/fedprox_example/client.py
```diff
@@ -65,7 +66,23 @@ def get_criterion(self, config: Config) -> _Loss:
     # Set the random seed for reproducibility
     set_all_random_seeds(args.seed)
 
-    client = MnistFedProxClient(data_path, [Accuracy()], device, reporters=[JsonReporter()])
+    json_reporter = JsonReporter()
+    reporters: list[BaseReporter] = [json_reporter]
+
+    # # Uncomment to log results to W and B
```
emersodb, line 72: Comments aren't idea for this type of thing, but without a real entity this will throw errors and kill things like smoke tests etc.

examples/fedprox_example/server.py
```diff
@@ -75,11 +76,21 @@ def main(config: dict[str, Any], server_address: str) -> None:
...
     json_reporter = JsonReporter()
     client_manager = SimpleClientManager()
-    if "reporting_config" in config:
-        wandb_reporter = WandBReporter("round", **config["reporting_config"])
-        reporters = [wandb_reporter, json_reporter]
-    else:
-        reporters = [json_reporter]
+    reporters: list[BaseReporter] = [json_reporter]
+
+    # Uncomment to log results to W and B
```
emersodb, line 81: Comments aren't ideal for this type of thing, but without a real entity this will throw errors and kill things like smoke tests etc.

examples/fedprox_example/config.yaml
```diff
@@ -14,12 +14,3 @@ proximal_weight_patience : 5 # The number of rounds to wait before increasing or
 n_clients: 3 # The number of clients in the FL experiment
 local_epochs: 1 # The number of epochs to complete for client
 batch_size: 128 # The batch size for client training
-
```
emersodb, line 17: Dropping the functionality using the config for enabling logging as it isn't super viable now. W and B reporters are instantiated explicitly for the clients.

fl4health/clients/basic_client.py
```diff
@@ -77,7 +77,10 @@ def __init__(
         self.device = device
         self.metrics = metrics
         self.progress_bar = progress_bar
+
         self.client_name = client_name if client_name is not None else generate_hash()
```
emersodb, line 81: We weren't logging this before, but it feels like something worth logging.

fl4health/servers/base_server.py
```diff
@@ -79,7 +79,10 @@ def __init__(
                 model=None, parameter_exchanger=None, model_checkpointers=None, state_checkpointer=None
             )
         self.on_init_parameters_config_fn = on_init_parameters_config_fn
+
         self.server_name = server_name if server_name is not None else generate_hash()
+        log(INFO, f"Server Name: {self.server_name}")
```
emersodb, line 84: We weren't logging this before, but it feels like something worth logging.

fl4health/servers/tabular_feature_alignment_server.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         self,
         client_manager: ClientManager,
         config: Config,
-        initialize_parameters: Callable[..., Parameters],
```
emersodb, line 32: We strictly use this function with a two int inputs, so making that explicit.

fl4health/servers/tabular_feature_alignment_server.py
```diff
@@ -48,20 +48,19 @@ def __init__(
                 In most cases it should be the "source of truth" for how FL alignment should proceed.
 
                 **NOTE:** This config is **DISTINCT** from the Flwr server config, which is extremely minimal.
-            strategy (Strategy | None, optional): The aggregation strategy to be used by the server to handle.
+            initialize_parameters (Callable[[int, int], Parameters]): Function used to initialize the model to be
```
emersodb, line 51: Fixing outdated docs here for funsy.

examples/fedprox_example/client.py
lotif: Can we have a flag or a param instead?

tests/reporting/test_wandb_reporter.py
```diff
@@ -0,0 +1,72 @@
...
+from fl4health.reporting import WandBReporter, WandBStepType
+
+
+def test_wandb_reporter_init(tmp_path: Path) -> None:
+    # Create with various step types
+    wandb_reporter = WandBReporter("round")
+    wandb_reporter = WandBReporter("step")
+    wandb_reporter = WandBReporter("epoch")
+    wandb_reporter = WandBReporter(WandBStepType.STEP)
+    wandb_reporter = WandBReporter(WandBStepType.EPOCH)
```
lotif, line 13: Not sure if I understand this correctly, but aren't these variables just overwriting each other like this?

tests/reporting/test_wandb_reporter.py
```diff
@@ -0,0 +1,72 @@
...
+    # Should set id and name, as neither was specified at the start.
+    wandb_reporter_with_none.initialize(id="id2", name="name2")
+
+    assert wandb_reporter_with_none.id == "id2"
+    assert wandb_reporter_with_none.name == "name2"
+
+    assert wandb_reporter_with_name_id.initialized
+    assert wandb_reporter_with_name.initialized
+    assert wandb_reporter_with_id.initialized
+    assert wandb_reporter_with_none.initialized
```
lotif, line 72: I would move those asserts to their respective blocks for clarity and readability.

examples/fedprox_example/client.py
emersodb: Oh...that's a good idea. Silly me 😂

.github/workflows/static_code_checks.yaml
```diff
@@ -44,9 +44,12 @@ jobs:
           # Ignoring vulnerabilities in cryptography
           # Fix is 43.0.1 but flwr 1.9 depends on < 43
           # GHSA-cjgq-5qmw-rcj6 is a Keras vulnerability that has no fix yet
+          # GHSA-4vmg-rw8f-92f9 is a torch vulnerability that arises from the lack of security verification
+          # during the deserialization process of PythonUDF objects and has no fix yet
```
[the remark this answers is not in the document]
emersodb: Ah weird. I'll see if I can just upgrade torch then. I don't know why they didn't include a fix.

examples/fedprox_example/server.py
emersodb: Will address with a flag.

tests/reporting/test_wandb_reporter.py
emersodb: Yeah they are. I'm essentially checking to make sure the objects are created without throwing and exception. Basically, if the rewrites happen seamlessly, the inits are executing without issue.

tests/reporting/test_wandb_reporter.py
emersodb: Sounds good.

examples/fedprox_example/client.py
emersodb: Done

examples/fedprox_example/server.py
emersodb: Done

.github/workflows/static_code_checks.yaml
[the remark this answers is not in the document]
emersodb: Done

fl4health/model_bases/masked_layers/masked_normalization_layers.py
```diff
@@ -144,7 +144,7 @@ def __init__(
         self,
         num_features: int,
         eps: float = 1e-5,
-        momentum: float = 0.1,
```
emersodb, line 147: Mypy typing issue popped up when upgrading torch to 2.4.1

pyproject.toml
```diff
@@ -17,7 +17,7 @@ numpy = "^1.24"
 pandas = "^2.0"
 flwr = "1.9.0"
 opacus = "^1.3.0"
-torch = "^2.3.0"
+torch = "^2.4.1"
```
emersodb, line 20: Addresses a vulnerability surfaced by pip-audit. Torchvision needs an upgrade as well if we do this.

research/picai/fl_nnunet/start_client.py
```diff
@@ -17,8 +17,7 @@
 import flwr as fl
 import torch
 from flwr.common.logger import log, update_console_handler
-from torchmetrics.classification import Dice
-from torchmetrics.segmentation import GeneralizedDiceScore
+from torchmetrics.segmentation import DiceScore, GeneralizedDiceScore
```
emersodb, line 20: When upgrading torch to 2.4.1 torchmetrics also gets updated. They moved Dice to DiceScore in segmentation.

tests/checkpointing/test_opacus_checkpointers.py
```diff
@@ -148,7 +148,10 @@ def test_fix_of_loss_stateless_model_exception(tmp_path: Path) -> None:
     with pytest.raises(AttributeError) as attribute_exception:
         torch_checkpointer.maybe_checkpoint(model, 0.54, {})
 
-    assert str(attribute_exception.value) == "Can't pickle local object 'vmap.<locals>.wrapped'"
+    assert (
```
emersodb, line 151: When upgrading torch to 2.4.1, Opacus logs a slighly different error, but its the same one that we're ensuring that we fix.

tests/utils/snapshotter_test.py
```diff
@@ -43,7 +43,7 @@ def test_optimizer_scheduler_model_snapshotter() -> None:
     input_data = torch.randn(32, 100)
     target_data = torch.randn(32, 2)
 
-    fl_client.optimizers = {"global": torch.optim.SGD(fl_client.model.parameters(), lr=0.001)}
+    fl_client.optimizers = {"global": torch.optim.SGD(fl_client.model.parameters(), lr=0.001, momentum=0.1)}
```
emersodb, line 46: When upgrading torch to 2.4.1, apparently their SGD optimizers no longer store state if you don't specify a momentum. So doing it here to force that storage to ensure test continuity.

lotif (approved):
Thanks for addressing the comments! Especially the "upgrade torch" one :)

PR #361, opened 2025-03-25
Changing the way we log process outputs in the smokes.

emersodb:
# PR Type
Other | Fix

# Short Description

The way we log some of the failures in our smoke tests obscures the true issues behind the failure that the smoke test is experiencing. A notable example, when the server experiences a timeout issue due to a failure on the client-side, the server will fail with a timeout error and dump its logs. However, the client logs, as they were previously processed, would not be printed out anywhere. This would make understanding what happened during the failure.

In this change, I'm forcing us to print all stdout logs as they come in with an attribution to which process produced it. The downside of this approach is that the logs will be intertwined together, which isn't ideal. However, these logs can be post-processed to split these. The upside is that everything that is piped to stdout for all processes before the failure will be available for inspection. 

Addressing the con mentioned above will be deferred to future iterations. The way we collect logs and raise things like timeout exceptions means logging all process outputs is a little more complicated and would require more thoughtful refactors.

# Tests Added

This actually is a test 🙂


tests/smoke_tests/run_smoke_test.py
```diff
@@ -481,6 +454,10 @@ async def run_fault_tolerance_smoke_test(
             stderr=asyncio.subprocess.STDOUT,
         )
 
+        # Reads lines from the server output in search of the startup log message. Times out after 20s of inactivity
```
emersodb, line 457: This change, and the one below, are implemented to fix the flakiness that we were experiencing `test_basic_server_client_checkpoint` smoke. Essentially, we weren't waiting almost any time at all for the server to come online. If the server took slightly longer to come up than one of the clients, it would kill the whole process, because the client would try to reach out to a server that wasn't there. This forces us to wait 20 seconds for the server to come up before the clients try to start.

tests/smoke_tests/test_smoke_tests.py
```diff
@@ -80,6 +80,7 @@ async def test_nnunet_config_2d(tolerance: float) -> None:
         config_path="tests/smoke_tests/nnunet_config_2d.yaml",
         dataset_path="examples/datasets/nnunet",
         tolerance=tolerance,
+        read_logs_timeout=450,
```
emersodb, line 83: In reading the smoke tests for the `test_nnunet_config_2d` smoke (which has been flaky)...it looked to me as it we were essentially just running over the timeout due to some fluctuation in the time the entire process takes. I think if we just increase the timeout for this specific test to a touch more (+150s) we should eliminate this flakiness...remains to be tested rigorously though.

emersodb: I ran the smoke tests for this PR 5 times in a row without a single failure. While that isn't definitive proof that the fixes in this PR worked for the flakiness of the tests, at the very least, it indicates that their flakiness has been reduced by a fair bit.

tests/smoke_tests/run_smoke_test.py
sanaAyrml: Oh that totally makes sense, amazing addition.

tests/smoke_tests/test_smoke_tests.py
sanaAyrml: amazing finding again, seeing these solved makes my day :)

tests/smoke_tests/run_smoke_test.py
```diff
@@ -23,6 +23,23 @@
 DEFAULT_TOLERANCE = 0.0005
 DEFAULT_READ_LOGS_TIMEOUT = 300
 
+SERVER_STARTUP_MESSAGES = [
+    # printed by fedprox, apfl, basic_example, fedbn, fedper, fedrep, and ditto, FENDA, fl_plus_local_ft & moon
+    # Update this is no longer in output, examples are actually triggered by the [ROUND 1] startup message
```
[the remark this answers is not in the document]
emersodb: Let me look into it quickly. I didn't change the comment here, just made it a global variable (with the comments that were previously there) to make it accessible to other functions.

sanaAyrml (approved):
This PR looks good for me, thanks for changing logging and putting time to solve some of the existing flakiness.

fatemetkl (approved):
LGTM! :rocket:

PR #366, opened 2025-04-01
Creating a Metrics Module (1/5)

emersodb:
# PR Type

Refactor

# Short Description

This is one of a series of PRs meant to facilitate refactoring of our metrics. **This first PR is purely moving functionality from our utils folder into a proper metrics folder.** and creating a metrics_base.py file to hold the abstract Metrics base class.

Because we use metrics everywhere in the library, just doing this touches a ton of files. The only real thing to review what now appears in `fl4health/metrics` which is what used to be in `utils/metrics.py`. I have separated out the `MetricManager` from our metrics implementations and the `metrics_aggregation` functionality.

# Tests Added

Existing tests have also been moved to their own testing module under tests/metrics/. Again, I've separated out the metric manager tests, but their contents remains intact.


fl4health/metrics/metrics.py
nerdai, line 1: I think it's okay for us at this stage, but maybe we should still retain `utils.metrics` and reimport from the new `metrics.metrics` module for backwards compat.

`utils.metrics.py`
```
# keep for backwards compatibility
from fl4health.metrics.metrics import *

import warnings

warnings.warn("<some warning message about the move and when this import path will no longer be valid.>")
```

fl4health/metrics/metrics_base.py
```diff
@@ -0,0 +1,67 @@
...
+        respective list.
+
+        Args:
+            input (torch.Tensor): The predictions of the model to be evaluated.
+            target (torch.Tensor): The ground truth target to evaluate predictions against.
+
+        Raises:
+            NotImplementedError: To be defined in the classes extending this class.
+        """
+        raise NotImplementedError
```
nerdai, line 40: as mentioned in some other PRs, at least as an FYI this `raise` doesn't actually get hit since we're using `~abc.ABC`. I think this case where this does get raised if we do `super().update()`.

If we've adopted this throughout the codebase, then I'm good with remaining consistent.

fl4health/servers/base_server.py
```diff
@@ -13,12 +13,12 @@
 from flwr.server.strategy import Strategy
 
 from fl4health.checkpointing.server_module import BaseServerCheckpointAndStateModule
+from fl4health.metrics.metrics_base import TEST_LOSS_KEY, TEST_NUM_EXAMPLES_KEY, MetricPrefix
```
nerdai, line 16: nit: just cause I see `base_reporter` underneath, which makes me wonder if we should name this `base_metrics` instead?

nerdai (approved):
LGTM!

I left a comment about backwards compat if we care to address that here.

fl4health/metrics/__init__.py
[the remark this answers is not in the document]
nerdai: oh yea that's a bit jank...

`fl4health/metrics/__init__.py`
```
from .metrics import ...

__all__ = [
    "...",
]
```

fl4health/metrics/metrics_base.py
```diff
@@ -0,0 +1,67 @@
+from abc import ABC, abstractmethod
```
lotif, line 1: Would it make sense to rename this file to `base.py` since it's already inside the `metrics` folder, which gives it enough context?

emersodb: Sure will do that

fl4health/metrics/metrics_base.py
emersodb: Yup, that makes sense. I think my thought was to throw on super calls in these situations. Is there something else you would suggest instead? I'm looking around for a standard, but it isn't obvious what the most "pythonic" thing to do here is (at least to me)

fl4health/metrics/metrics_base.py
emersodb: Actually, given your comment below, I'm going to go with `base_metrics.py `instead 🙂

fl4health/servers/base_server.py
emersodb: Done

fl4health/metrics/__init__.py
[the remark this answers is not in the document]
emersodb: Done.

fl4health/metrics/metrics.py
emersodb: Done.

PR #368, opened 2025-04-01
Creating Compound Metrics and Expanding our checkpointers (2/5)

emersodb:
# PR Type

Refactor

# Short Description

This PR does a few fairly small things. 

1. It slightly expands our checkpointing classes to have a class for checkpointing based on a metric. It also allows for naming the scoring function for a bit more informative logging.
2. It moves our TransformsMetric into a new file called `compound_metrics.py`, which will hold metrics that wrap metrics (hence the compound). `EMAMetric`, which will be added in the next PR, is another example of this kind of object.

# Tests Added

Most of the changes are refactors, but I added extensive testing of the `TransformsMetric` class, as we didn't seem to have any prior to this.


tests/metrics/transforms_metric_test.py
```diff
@@ -0,0 +1,147 @@
...
+    preds_1 = torch.Tensor([0.9, 0.7, 0.6, 0.9, 0.1, 0.2, 0.99])
+    targets_1 = torch.Tensor([1.0, 0.0, 1.0, 1.0, 0.0, 1.0, 0.0])
+
+    preds_2 = torch.Tensor([0.19, 0.75, 0.26, 0.49, 0.12, 0.92, 0.99])
+    targets_2 = torch.Tensor([1.0, 0.0, 1.0, 1.0, 0.0, 1.0, 0.0])
+
+    preds_3 = torch.Tensor([0.19, 0.75, 0.26, 0.49])
+    targets_3 = torch.Tensor([1.0, 0.0, 1.0, 0.0])
+
+    accuracy_metric = Accuracy("accuracy")
```
scarere, line 49: nit: Just brainstorming here, is there a way to test the transforms metric without depending on another metric in the library (ie. Accuracy)? Changes to the accuracy metric would affect these tests. Can we create a dummy metric of some sort?

scarere (approved):
Looks good to me

lotif (approved):
LGTM

tests/metrics/transforms_metric_test.py
emersodb: Yeah I think I can do that. Probably a worthwhile idea.

PR #369, opened 2025-04-01
Adding EMA Metric (3/5)

emersodb:
# PR Type

Feature

# Short Description

Creating an Exponential Moving Average Metric that can wrap any existing metric and simply keeps track of past metric values to be used in EMA calculations.

# Tests Added

Tests covering the EMA functionality.


fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,88 @@
 T = TypeVar("T", bound=Metric)
 
 
+class EmaMetric(Metric, Generic[T]):
+    def __init__(self, metric: T, smoothing_factor: float = 0.1, name: str | None = None):
+        """
+        Exponential Moving Average (EMA) metric wrapper to apply EMA to the underlying metric.
+
+        NOTE: If the underlying metric accumulates batches during update, then updating this metric without clearing
```
scarere, line 21: I'm just noticing that we take an instance of a metric as an argument here. So if the user creates a metric that they are actually using, and then creates an EMA metric from that instance, then the EMA metric is going to mess with the original metric. Should we create a deepcopy of the metric? Or have the user pass the metric class instead of an instance which EmaMetric can than instantiate itself?

fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,88 @@
 T = TypeVar("T", bound=Metric)
 
 
+class EmaMetric(Metric, Generic[T]):
```
scarere, line 16: nit: EMAMetric? since EMA is an acronym

scarere: but truthfully EmaMetric looks nicer. Whats best practice in these sorts of scenarios?

fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,88 @@
...
+    def __init__(self, metric: T, smoothing_factor: float = 0.1, name: str | None = None):
+        """
+        Exponential Moving Average (EMA) metric wrapper to apply EMA to the underlying metric.
+
+        NOTE: If the underlying metric accumulates batches during update, then updating this metric without clearing
+        in between will result in previously seen inputs and targets being a part of subsequent computes
+        (i.e. present in both previous scores and current score).
```
scarere, line 23: nit: Can we clarify this a bit? Maybe specify that it is present twice in the compute? Or maybe just add present in both previous score and current score used in the computation of the EMA?

fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,88 @@
...
+        for key, previous_score in self.previous_score.items():
+            current_score = metrics_dict[key]
+            if not isinstance(current_score, (str, bytes)) and not isinstance(previous_score, (str, bytes)):
+                self.previous_score[key] = (
+                    self.smoothing_factor * current_score + (1 - self.smoothing_factor) * previous_score
+                )
+
+        return copy.deepcopy(self.previous_score)
+
+    def _drop_str_or_bytes_scores_and_store(self, metrics_dict: Metrics) -> None:
```
scarere, line 81: Why are we supporting Metrics to be type bytes or str in the first place? I feel like enforcing them to be some sort of number might actually be beneficial

fl4health/metrics/compound_metrics.py
emersodb: From what I can tell, there isn't a canonical naming convention for acronyms. Keeping them upper-cased is an acceptable stylistic choice while camel-casing is also accepted. We've chosen to do camel-casing for acronyms throughout the library as much as possible. There are a few cases that we've not been consistent, but this is the approach for the most part.

fl4health/metrics/compound_metrics.py
emersodb: I see what you mean. That does open up the potential for unintended side-effects if used in somewhat of an odd way. Will move to deep copying for compound metrics like this.

fl4health/metrics/compound_metrics.py
emersodb: This is a really good question and the answer isn't all that great. The reason we're using Metrics as including `bytes` and `str` types is because that's the typing that Flower expects them to have. See the underlying typing of Metrics in this:
```
from flwr.common.typing import Metrics
```
We could technically do a little bit of dancing to have our metrics not produce a dictionary of that type, but we'd have to do conversion on the way in and out of the client interfaces so that Flower is happy. In most cases it isn't an issue, but a EMA of such types is not well-defined.

In looking at this a bit more, I realized scalar includes more values than it used to and EMA is really only applicable to float and int therein. So I'll switch this up a bit.

tests/metrics/transforms_metric_test.py
```diff
@@ -1,26 +1,8 @@
 import torch
-from flwr.common.typing import Scalar
 from pytest import approx
-from sklearn import metrics as sklearn_metrics
 
 from fl4health.metrics.compound_metrics import TransformsMetric
-from fl4health.metrics.metrics import SimpleMetric
-
-
-class AccuracyForTest(SimpleMetric):
```
emersodb, line 10: Moving this guy to a utility file, as it is also used for the ema metric

fl4health/metrics/compound_metrics.py
scarere: Just out of curiosity, can you explain to me why you're using a TypeVar? If Type T is only bound to Metric and no other types, then why not just make the type Metric?

fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,113 @@
...
+
+            ema = EmaMetric(Accuracy(), 0.1)
+
+            preds_1 = torch.Tensor([1, 0, 1]), targets_1 = torch.Tensor([1, 1, 1])
+
+            ema.update(preds_1, targets_1)
+
+            ema.compute() -> 0.667
+
+            preds_2 = torch.Tensor([1, 0, 1]), targets_1 = torch.Tensor([0, 0, 1])
```
scarere, line 37: Is the math on this example correct? It seems like preds_2 would also have an accuracy of 0.667, so wouldn't ema update be 0.9(0.667) + 0.1(0.667)? Not sure if I'm have a brain fart here or not though. Also nit: can you make the targets for both preds_1 and preds_2 be [1, 1, 1] so that its easier to figure out the accuracy in my head by just counting the number of 1's in the predictions?

fl4health/metrics/compound_metrics.py
```diff
@@ -10,6 +13,113 @@
...
+            smoothing_factor (float, optional): Smoothing factor in range [0, 1] for the EMA. Smaller values increase
+                smoothing by weighting previous scores more heavily. Defaults to 0.1.
+            name (str | None, optional): Name of the EMAMetric. If left as None will default to 'EMA_{metric.name}'.
+
+        """
+        # Create a copy of the metrics object so that we do not inadvertently change the provided object elsewhere
+        self.metric = copy.deepcopy(metric)
+        assert 0.0 <= smoothing_factor <= 1.0, f"smoothing_factor should be in [0, 1] but was {smoothing_factor}"
+        self.smoothing_factor = smoothing_factor
+        self.previous_score: Metrics | None = None
```
scarere, line 60: nit: Not something to be addressed in this PR but something about having a Metrics type for the dict and a Metric type for the classes rubs me the way.

emersodb: Yeah I get it. We could alias the Flower `Metrics` type here to be called something else, but we can discuss the write thing to do in a follow up.

fl4health/metrics/compound_metrics.py
emersodb: No you're not losing your mind. That's my bad. I just didn't quite set this up correctly. I've fixed this so that it is both accurate and takes your suggestion of the target tensors being [1, 1, 1]. I've also added a test of this behavior to the ema tests, because it was not covered like it should have been. Thanks for catching this!

fl4health/metrics/compound_metrics.py
scarere: Like when you hover over them vscode will know what Metric subsclass is being used if you typed it with a typevar but if you defined the type as Metric directly then it won't?

scarere (approved):

fl4health/metrics/compound_metrics.py
emersodb: Yup. Right now, that's the main advantage with the generic typing, but it can make other things we want to do later easier as well 🙂

PR #371, opened 2025-04-01
Implementing efficient multiclass metrics base and Dice (5/5)

emersodb:
# PR Type

Feature

# Short Description

Implementing the multiclass versions of the efficient classifications metric base class and Dice scores. These are a somewhat straightforward extension of the binary versions, but with their own sets of "gotchas." 

# Tests Added

A lot of tests have been put into place for both the efficient multiclass base class and the Multiclass Dice score class to ensure the correctness of the implementation.


fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    assert (
+        tensor1.shape != tensor2.shape
+    ), f"Could not infer the label dimension of tensors with the same shape: {tensor1.shape}"
+
+    assert 0 <= (tensor1.ndim - tensor2.ndim) <= 1, (
+        f"Could not infer the label dimension of tensors with shapes: tensor1: {tensor1.shape}), tensor 2: "
+        f"({tensor2.shape}). Expected tensor1 to be larger than tensor2 by at most 1 dimension."
+    )
+
+    # Infer label dimension.
```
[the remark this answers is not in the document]
scarere: nit: if tensor1 and tensor2 are the same size, then we are looking for matching dims with diff size, but one of them must have size 1 technically for that to make sense, that check is done afterwards

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+            "The inferred candidate dimension has different sizes on each tensor, was expecting one to be empty."
+        )
+    return label_dim
+
+
+def map_label_index_tensor_to_one_hot(
+    label_index_tensor: torch.Tensor, target_shape: torch.Size, label_dim: int
+) -> torch.Tensor:
+    """
+    Maps the provided ``label_index_tensor``, which has label indices at the provided ``label_dim``. This function
```
scarere, line 75: Worth explaining that the "label_dim" is expected to be empty. Also if we're being particular about the term one-hot-encoding then we probably shouldn't use it here. We're technically "vectorizing" the tensor or turning it from a dense tensor into a sparse tensor

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    one_hot_encoded_tensor.scatter_(label_dim, label_index_tensor.to(torch.int64), 1)
+    return one_hot_encoded_tensor
+
+
+def align_pred_and_target_shapes(
+    preds: torch.Tensor, targets: torch.Tensor, label_dim: int | None = None
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """
+    If necessary, attempts to correct shape mismatches between the given tensors by inferring which one
+    to one-hot-encode. Note that if both preds and targets have the same shape then nothing needs to be modified. If
```
scarere, line 112: Same note here, i think we spoke about not overloading the ohe term.

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    there is a mismatch, it is assumed that the labels in one of the predictions or the targets are vector encoded
+    in some way (either one-hot or soft) while the other is label index encoded.
+
+    If one is vector encoded but not the other, then both are returned as vector encoded tensors.
+
+    For example, consider a problem with 3 label classes with the preds vector encoded and the targets label encoded
```
scarere, line 118: Examples are a great addition

scarere: side note, if we reintroduce the generic reduction axes specification for metrics, I think one or two examples will go a long way in clarifying whats going on

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+        targets = torch.Tensor([[2], [1]])
+
+    preds has shape (2, 3) and targets has shape (2, 1). This function will convert targets to a one-hot-encode tensor
+    with contents
+
+    .. code-block:: python
+
+        targets = torch.Tensor([0, 0, 1], [0, 1, 0])
+
+    NOTE: This function ASSUMES label-index encoding if the shapes are misaligned. This assumption doesn't necessarily
```
scarere, line 133: I feel like we can add this functionality back in. I had the following code block in the old implementation. Maybe we can add this if statement int `map_label_index_to_one_hot`. Probably apt to rename the function like `map_dense_to_sparse`
```python
# One-hot-encode tensor2. We know at this point that it has an empty dim along label axis
    if torch.is_floating_point(tensor2) and torch.frac(tensor2).sum() != 0:
        # If tensor is continuous it must be binary classification and elements must be probabilities in range [0, 1].
        t2_ohe = torch.cat([1 - tensor2, tensor2], dim=label_dim)
    else:
        # If tensor2 is not continuous then it must contain class labels. One hot encode the tensor.
        t2_ohe = torch.zeros(tensor1.shape, device=tensor2.device)
        t2_ohe.scatter_(label_dim, tensor2.to(torch.int64), 1)
```

tests/metrics/multi_classification_metric_no_batch_test.py
scarere, line 1: How many test files are we at in `test/metrics`, should we add another level of nesting for organization?

scarere (commented):
Overall looks good. I think there are still a couple questions on what the best way to implement things will be. I personally think there's some opportunity to refactor a bit here but lets wait and see what andre thinks

tests/metrics/efficient_dice_metric_batch_test.py
```diff
@@ -0,0 +1,214 @@
...
+    dice_target_12 = (2 * tp1[0][1]) / (2 * tp1[0][1] + fp1[0][1] + fn1[0][1])
+    dice_target_13 = (2 * tp1[0][2]) / (2 * tp1[0][2] + fp1[0][2] + fn1[0][2])
+    dice_target_21 = (2 * tp1[1][0]) / (2 * tp1[1][0] + fp1[1][0] + fn1[1][0])
+    dice_target_22 = (2 * tp1[1][1]) / (2 * tp1[1][1] + fp1[1][1] + fn1[1][1])
+    dice_target_23 = (2 * tp1[1][2]) / (2 * tp1[1][2] + fp1[1][2] + fn1[1][2])
+    assert result["DICE"] == approx(
+        (1.0 / 6.0)
+        * (dice_target_11 + dice_target_12 + dice_target_13 + dice_target_21 + dice_target_22 + dice_target_23),
+        abs=1e-4,
+    )
```
lotif, line 125: Optional, but lines 115 to 125 can be replaced with:
```python
dice_target = [[(2 * tp1[i][j]) / (2 * tp1[i][j] + fp1[i][j] + fn1[i][j]) for j in range(len(tp1[i]))] for i in range(len(tp1))]
assert result["DICE"] == approx((1.0 / 6.0) * sum(sum(d) for d in dice_target), abs=1e-4)
```
Whether that's simpler can be up for debate, but I think it is 😅

lotif (changes requested):

fl4health/metrics/efficient_metrics.py
```diff
@@ -46,6 +50,109 @@ def compute_dice_on_count_tensors(
...
+        )
+        self.zero_division = zero_division
+
+    def compute_from_counts(
+        self,
+        true_positives: torch.Tensor,
+        false_positives: torch.Tensor,
+        false_negatives: torch.Tensor,
+        true_negatives: torch.Tensor,
+    ) -> Metrics:
```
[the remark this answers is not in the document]
emersodb: Added

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+        User defined convenience method that calculates the desired metric given the predictions and target without
+        accumulating it into the class itself
+
+        Raises:
+            NotImplementedError: User must define this method.
+        """
+        raise NotImplementedError
+
+
+class MultiClassificationMetric(ClassificationMetric):
```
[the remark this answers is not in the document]
emersodb: I've separated these out, in a similar way to Tensorflow, as a matter of convenience because, despite them being heavily related (and hence the base class housing a lot of shared functionality, there are subtle differences that make the classes distinct enough that they do feel better as separate entities. Some basic examples include: 
- the binary class allowing for implicit class predictions (i.e. 0.8 implying [0.2, 0.8], which we do not allow in the multiclass setting. This requires a lot of special treatment.
- the binary class not having a label dimension (related to the implict class prediction setting again).
- Handling predictions by label index (out of the box for binary, needs special treatment in multi-class)
- A meaningful notion of background class in the multiclass setting
- Enforcing reporting related to a single label in the binary setting (so that it doesn't matter how you provide predictions/target tensors, you always get consistent behavior).

Because of the above, I think it makes sense to have these be separate, for our purposes, to make things smooth.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+    def __init__(
+        self,
+        name: str,
+        label_dim: int,
+        batch_dim: int | None = None,
+        dtype: torch.dtype = torch.float32,
+        threshold: float | int | None = None,
+        ignore_background: int | None = None,
```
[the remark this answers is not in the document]
emersodb: I think removing the background is perhaps less common (though I know you use it for your stuff). I was following your implementation here to make it as generic as possible. However, if you think tying it strongly to the `label_dim` makes sense, I'm good to do that.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+            preds.shape == targets.shape
+        ), f"Preds ({preds.shape}) and targets ({targets.shape}) should have the same shape but do not."
+
+        indices = torch.arange(1, preds.shape[ignore_background], device=preds.device)
+        preds = torch.index_select(preds, ignore_background, indices)
+        targets = torch.index_select(targets, ignore_background, indices.to(targets.device))
+        return preds, targets
+
+    @classmethod
+    def _transpose_matrix_unless_empty(cls, matrix: torch.Tensor) -> torch.Tensor:
```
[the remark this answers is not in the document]
emersodb: Good point. Will change.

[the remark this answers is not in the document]
emersodb: In a lot of places in the library, we assume the batch dimension is first, but it isn't always the case. There are different conventions that are used. I think the most common is batch first, but it isn't universal. So I think supporting it being something potentially different is okay. 

I know you liked the flexibility of the specification of reduce along and over and I thought it was really nice. However, before getting quite deep into the implications of the arguments, I found it difficult to know what the implications of specifying the dimensions had for what I was trying to measure. I went with this approach, specifying two dimensions (optionally in the case of `batch_dim`) because they have meaningful names for most people and the implications of specifying them was easier to concretely explain.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+
+        Returns:
+            tuple[torch.Tensor, torch.Tensor]: Potentially transformed predictions and targets tensors, in that order
+        """
+
+        preds, targets = super()._transform_tensors(preds, targets)
+
+        # Attempt to automatically match pred and target shape.
+        # This supports any combination of hard/soft, vector/not-vector encoded
+        preds, targets = align_pred_and_target_shapes(preds, targets, self.label_dim)
```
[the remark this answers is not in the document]
emersodb: It's a nice function

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
+import torch
+
+
+def infer_label_dim(tensor1: torch.Tensor, tensor2: torch.Tensor) -> int:
+    """Infers the label dimension given two related tensors of different shapes.
+
+    Generally useful for inferring the label dimension when one tensor is vector-encoded and the other is not. The
```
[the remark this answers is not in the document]
emersodb: I know what you mean, but I'm not really sure where that would be?

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    label dimension is inferred by looking for dimensions that either are not the same size, or are not present in
+    tensor 2.
+
+    Args:
```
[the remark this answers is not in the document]
emersodb: I think I might have just taken your docs here 😂. I don't remember to be honest.

fl4health/metrics/utils.py
```diff
@@ -0,0 +1,181 @@
...
+    Raises:
+        AssertionError: If the the label dimension cannot be inferred without ambiguity. For example if a dimension
+            next to the label dimension has the same size.
+
+    Returns:
+        int: Index of the dimension along tensor 1 that corresponds to the label dimension.
+    """
+    assert (
+        tensor1.shape != tensor2.shape
+    ), f"Could not infer the label dimension of tensors with the same shape: {tensor1.shape}"
```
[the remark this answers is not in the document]
emersodb: Good call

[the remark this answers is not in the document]
emersodb: Err...actually, this will only throw if their shape is equal. So it would just be a repeat. I think that's why I just reported one.

fl4health/metrics/utils.py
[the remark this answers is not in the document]
emersodb: Sure, let me see if I can work these in.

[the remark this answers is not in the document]
emersodb: I think I might be missing where the nit above is for? I'm sure it makes sense somewhere, but I'm being slow 😕

fl4health/metrics/utils.py
emersodb: I do think we mean one-hot-encoding here. Below, in the comment we mention that if there is a mismatch we assume that there the mismatch is induced by label-index encoding, which needs to be one-hot-encoded specifically. 

I might be missing something though. So let me know if I am :-)

fl4health/metrics/utils.py
emersodb: I made the simplifying assumption that, in the binary case, the user will have properly aligned their tensors before asking for evaluation. I think that's pretty reasonable to force one to do that. It's a bit easier to handle misalignment in the multi-class setting (fewer edge cases due to not allowing implied classes). So I decided we could try to support it.

I'm also a little worried 
```
torch.is_floating_point(tensor2) and torch.frac(tensor2).sum() != 0:
```
might be brittle if the floating points are subject to machine precision fluctuations. That is, representing 1.0 numerically might not always be identically zero when calling frac? Perhaps just paranoid.

tests/metrics/efficient_dice_metric_batch_test.py
```diff
@@ -0,0 +1,214 @@
+import math
+
+import torch
+from pytest import approx
+
+from fl4health.metrics.efficient_metrics import MultiClassDice
+
+
+def test_multi_dice_metric_with_threshold() -> None:
+    torch.manual_seed(42)
```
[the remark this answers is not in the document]
emersodb: Yup!

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
scarere: Oh I understand now the purpose behind this now. If binary was a subclass of multiclass then technically anything that required a multiclass metric would accept a binary metric class instance, which would not be compatible. In that sense I see the necessity for having 3 base classes here. 

With that in mind, is there any way to move more of the common functionality into the more generic `ClassificationMetric` base class? I think the benefit of this is not only less lines of code, but it will make the differences in functionality between Binary and Multiclass base classes clearer. Some Ideas:
- An easy change would be moving the not implemented methods (`__call__`, `compute_from_counts`) into the `ClassificationMetric`. 
- If you move the call to `self._transform_tensors` to the top of `self.update`, then you can call `self._assert_correct_ranges` before `self.count_tp_fp_fn_tn`. This removes the need for `BinaryClassificationMetric` to override `self._transform_tensors`.  Also it makes a call to the assertion a default to the base class which is probably good.
- There is a lot of duplicated functionality in `self.count_tp_fp_fn_tn` between the two subclasses. Moving this into `ClassificationMetric` is less straigh-forward but I think if you can think of a way to do it then that will improve the clarity of the code.

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
scarere: One thing to note is that we do note reduce along the class/label dimension by default. Are there any situations in which we would?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+                those two expected settings
+        """
+        if matrix.ndim == 2:
+            return matrix.transpose(0, 1)
+        elif matrix.numel() == 0:
+            return matrix
+        else:
+            raise ValueError(f"Expected tensor to either be 2D or empty but has shape {matrix.shape}")
+
+    def _transform_tensors(self, preds: torch.Tensor, targets: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
```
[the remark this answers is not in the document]
scarere: Refer to my other comment. There is a way to move the function calls around where BinaryClassificationMetric doesn't need to override this method

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -569,7 +570,265 @@ def compute_from_counts(
...
+
+        # Remove the background channel from the axis specified by ignore_background_axis
+        if self.ignore_background is not None:
+            preds, targets = self._remove_background(self.ignore_background, preds, targets)
+
+        super()._assert_correct_ranges(preds, targets)
+
+        return preds, targets
+
+    def count_tp_fp_fn_tn(
```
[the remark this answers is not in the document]
scarere: I think i missed a lot of the differences in my last review. It's clearer now the differences between the two classes in this method. One way to unify these methods is to define the base ClassificationMetric method as:
```python
def count_tp_fp_fn_tn(self, preds, targets):
    # Compute counts. If we're ignoring a count, set it as an empty tensor to avoid downstream errors.
    true_positives, false_positives, false_negatives, true_negatives = self._prepare_counts_from_preds_and_targets(
        preds, targets
    )

    axes_to_ignore: set[int] = {self.label_dim} if self.label_dim is not None else set()
    if self.batch_dim is not None:
        axes_to_ignore.add(self.batch_dim)
    sum_axes = tuple([i for i in range(preds.ndim) if i not in axes_to_ignore])

    # If sum_axes is empty, then we have nothing to do.
    if len(sum_axes) != 0:
        true_positives, false_positives, false_negatives, true_negatives = self._sum_along_axes_or_discard(
            sum_axes, true_positives, false_positives, false_negatives, true_negatives
        )

    return true_positives, false_positives, false_negatives, true_negatives
```
Then the method for BinaryMetric would be:
```python
def count_tp_fp_fn_tn(self, preds, targets):
    # Make sure after transformation the preds and targets have the right shape
    assert (
        preds.shape == targets.shape
    ), f"Preds and targets must have the same shape but got {preds.shape} and {targets.shape} respectively."

    # Assert that the label dimension for these tensors is of size 2 at most.
    if self.label_dim is not None:
        assert preds.shape[self.label_dim] <= 2, (
            f"Label dimension for preds tensor is greater than 2 {preds.shape[self.label_dim]}. This class is "
            "meant for binary metric computation only"
        )
        assert targets.shape[self.label_dim] <= 2, (
            f"Label dimension for targets tensor is greater than 2 {targets.shape[self.label_dim]}. This class is "
            "meant for binary metric computation only"
        )

    true_positives, false_positives, false_negatives, true_negatives = super().count_tp_fp_fn_tn(preds, targets)

    # Postprocess
    true_positives = self._post_process_count_tensor(true_positives)
    false_positives = self._post_process_count_tensor(false_positives)
    false_negatives = self._post_process_count_tensor(false_negatives)
    true_negatives = self._post_process_count_tensor(true_negatives)

    if self.pos_label == 0:
        # Need to flip the label interpretations
        return true_negatives, false_negatives, false_positives, true_positives

    return true_positives, false_positives, false_negatives, true_negatives
```

Something similar could be done for MultiClassificationmetric. My argument for this approach would be that all the code in the overriden methods is now unique to that class (for the most part). Note that the call to transform_tensors is moved out of the function but that was already my suggestion from a previous comment. Also as to the differences with axes_to_ignore, I believe it should still work the same. If mypy complains then maybe just check if label_dim is an int. As an additional note, if you take this approach then you can remove the `_prepare_counts_from_preds_and_targets` and `_sum_along_axes_or_discard` methods by making them part of the base method.

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
scarere: I realize now that the purpose behind having comput_from_counts in the subclasses is to ovverride the docstring

fl4health/metrics/utils.py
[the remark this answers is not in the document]
scarere: I don't exactly remember but I think what I was trying to say was that if tensor1 and tensor2 have the same number of dimensions but different shapes, then the mismatched dimension must have size 1 for one of the tensors. For example `preds: (5, 1, 10, 10), targets: (5, 3, 10, 10)` makes sense since we can assume targets are ohe and preds are not. But `preds: (5, 4, 10, 10), targets: (5, 3, 10, 10)` does not make sense. We would still identify axis 1 as the mismatched axis, but there is no way to one-hot-encode either tensor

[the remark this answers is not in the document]
scarere: So i think i was trying to say that we could add the nuance of that we are not just looking for axes with different sizes but in fact one of them must be size 1

fl4health/metrics/utils.py
scarere: I honestly can't remember what I meant by that comment either as I can't tell the difference as well.

scarere: On a separate note, do we need the whole target_shape or can we just have the use specify the number of classes?

fl4health/metrics/utils.py
scarere: So I think the reason this is maybe not making sense is because before in my implementation I was using this for the BinaryClassificationMetric class as well. So i was covering the case in which we have continuous binary predictions that implicitly imply the class, and we "vectorize" them into a sparse tensor of continuous predictions. Correct me if I'm wrong but it seems we're assuming the predictions and targets will always be multiclass and therefore it will be impossible to have this scenario come up. I think that is what I initially meant but honestly don't remember and not totally sure

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: That's part of it. I think also, I'm technically not allowed to **not** have an implementation of some kind of implementation in the inheriting classes since it is marked as an `abstractmethod` in `ClassificationMetric`. The overriding of the docstrings is a nice bonus, but also not the only reason I have to have it 😂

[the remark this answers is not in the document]
emersodb: Thanks for the suggestions. I'll incorporate these and also take another pass through `count_tp_fp_fn_tn` to see if I can think of any additional clever ways for unification 🙂

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: I think there are situations where you might do that which aren't too exotic. For example, if you wanted to compute pure accuracy for a multiclass (single true label) problem, you might want to reduce along the label dimension to get all of the TPs regardless of the label. While we don't offer that reduction by default in these classes, it isn't hard to implement that in the `compute_from_counts`. Our counts will always either be or shape (n_classes, ) or (num_samples, n_classes). So the reduction can be implemented by an end user easily.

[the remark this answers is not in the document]
emersodb: I've added this comment 
```
NOTE: A user can implement further reduction along the label dimension (summing TPs across labels for example), if desired. It just needs to be handled in the implementation of this function
```

To the documentation for `compute_from_counts`.

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Sounds good. Will implement!

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Cool. Will give this a shot! Thanks for the suggestion.

fl4health/metrics/utils.py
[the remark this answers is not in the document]
emersodb: Right I see. Basically just updating the docs to reflect this nuance. Will do.

fl4health/metrics/utils.py
emersodb: It isn't strictly necessary, but in the calls to this function, we already have the shape that we want the target to have. So it was easier just to have this function take the predefined shape than to reconstruct it using the original shape, `label_dim`, and number of classes again within this function.

fl4health/metrics/utils.py
emersodb: Yes that makes sense. I see what you mean there. I've simplified this function to strongly assume a multi-class setting. So the case that you're referring to where you need to vectorize an implicit 2-class pred/target is not longer admissible. 

There is a note in the doc string for the function that tries to call this out. However, it was a bit of a ways down in the docs. I've moved it up above the example to make it more prominent.

fl4health/metrics/utils.py
emersodb: I tried to look into this too when I was working on converting your implementation and couldn't come up with a better implementation that yours above. I can see what it would be hard, but it does feel like we can't be the only people to want to do this 😂. 

In any case, because of the nuance in binary pred/target representations, I just thought it would be easier for us to say, please align your stuff before hand ha

tests/metrics/multi_classification_metric_no_batch_test.py
emersodb: Fair question. There are quite a few. Let me compartmentalize them a bit.

fl4health/metrics/efficient_metrics.py
```diff
@@ -4,7 +4,11 @@
...
 
-from fl4health.metrics.efficient_metrics_base import BinaryClassificationMetric
+from fl4health.metrics.efficient_metrics_base import (
+    BinaryClassificationMetric,
+    MetricOutcome,
+    MultiClassificationMetric,
+)
 
 
 def compute_dice_on_count_tensors(
```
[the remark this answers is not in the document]
emersodb: Moved to a utility file now along with another function that made sense to do this for.

tests/metrics/multi_classification_metric_no_batch_test.py
emersodb: Done

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Made the `_transform_tensors` changes. Good call. A nice compacting of code there 🙂. Will work on the `count_tp_fp_fn_tn` stuff. I think you made a suggestion below. So will try that now.

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Done!

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Okay...there is actually something bad about doing the `_transform_tensors` change that you suggested. It means that when we implement __call__ for the dice metrics in the binary setting, we have to remember to do the transformation and assertions before calling for the counts. I don't think that's a great setting to implement a convenience function because it differs so much from implementing the compute function.

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
emersodb: Done

fl4health/metrics/efficient_metrics_base.py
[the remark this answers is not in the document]
scarere: Ok so moving `_transform_tensors` back into `count_tp_fp_fn_tn` will mess with that other suggestion I made. One solution that feels kind of hacky is to just call `_transform_tensors` at the start of the overridden method (since it is being overriden anyways for both subclasses), then all the custom assertions are run, then when the `super().count_tp_fp_fn_tn` is called it will technically call `_transform_tensors` again which should do nothing. This feels like a bad solution. We could also just not call it at all in the base class and then the subclasses have to call it but that also feels suboptimal. Maybe we should just wrap the `__call__` function and have the user define a `call` function? Downside to that is that they would have to know not to use `call` and instead to use the implicit `__call__`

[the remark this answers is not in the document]
scarere: So I just realized that we're not calling `_transform_tensors` in `count_tp_fp_fn_tn` method for the base class. So what if we add the call to `assert_ranges` to the end of `tranform_tensors` in the base class, then we can actually get rid of the need to override it in the binary subclass and everything else stays the same. `_transform_tensors` is still being called in the subclass anyways.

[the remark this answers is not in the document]
scarere: One downside to both the way it is now and my suggested approach is that it feels weird to have a method defined in the base class but not used in the base class at all. Is that bad or is that common practice?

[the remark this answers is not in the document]
emersodb: For the first comment, the reason I don't do this now is because, in the multi-class case, we need to run `align_pred_and_target_shapes` before we check the ranges. This is required, for example, if the targets or predictions are label-index encoded (i.e. you could have a 3 as a legitimate value in the tensor). These first need to be one-hot encoded to fit into the right ranges.

For the second, as a matter of practice, I think that's okay. Base-classes can be places where functionality is centralized, even if it is only ever used in the inheriting child classes.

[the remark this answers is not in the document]
scarere: Oh I didn't notice that, you're right we can't just add it to the end of `_transform_tensors`. What about calling the two methods separately inside the overriden definition of `count_tp_fp_fn_tn`?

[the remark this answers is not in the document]
scarere: ie 
```python
def count_tp_fp_fn_tn(self, preds, targets):
    preds, targets = self._transform_tensors(preds, targets)
    self._assert_correct_ranges(preds, targets)
    ...
```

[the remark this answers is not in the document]
emersodb: I'll give it a shot!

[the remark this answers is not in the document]
emersodb: Done

fl4health/metrics/efficient_metrics_base.py
scarere, line 1: nit: In the ClassificationMetric docstring I would just mention something about the fact that it is memory efficient in the initial one-line description

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -262,10 +262,49 @@ def compute(self, name: str | None = None) -> Metrics:
...
 
     @abstractmethod
-    def count_tp_fp_fn_tn(
+    def count_tp_fp_tn_fn(
         self, preds: torch.Tensor, targets: torch.Tensor
     ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
-        raise NotImplementedError
+        """
+        Given two tensors containing model predictions and targets, returns the number of true positives (tp), false
+        positives (fp), true negatives (tn), and false negatives (fn), respecting ``self.batch_dim`` and
```
scarere, line 270: nit: I don't think the word respecting is particularly clear here. I think for one of the subclasses you're a little more explicit and give an example of the shapes that tp, fp, fn and fn will be based on these arguments

fl4health/metrics/efficient_metrics_base.py
scarere, line 1: nit: Also for the methods in ClassificationMetric that are not actually called and are only used by subclasses, does it make sense to note that in their documentation. Say something like "this method is meant to be used by subclasses..."

scarere, line 1: I know its mentioned in the init docstring, but since theres some tensor preprocessing going on, in the _prepare_counts docstring maybe explicitly state that preds and targets must be in range [0, 1]

scarere, line 1: Just looking at the postprocessing that is done in count_tp_fp_tn_fn and it seems like both Binary and Multiclass subclasses seem to be doing the same thing, which is just ensuring the batch_dimension is first. I noticed that in the base classification metric class it mentions that it makes this assumption and that subclasses must ensure this is the case. Can this functionality be moved into the base class? I feel like you might be able to just copy the `_postprocess_count_tensors` method and it should work for both subclasses (unless I'm missing something). This reducess the amount of code eliminating the need for each subclass needing to handle this separately and makes sense as something for the base class to handle.

fl4health/metrics/efficient_metrics.py
```diff
@@ -271,7 +271,7 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
         Returns:
             Scalar: Mean dice score for the provided tensors
         """
-        true_positives, false_positives, false_negatives, _ = self.count_tp_fp_fn_tn(input, target)
+        true_positives, false_positives, _, false_negatives = self.count_tp_fp_tn_fn(input, target)
```
scarere, line 274: We're not using the true negatives here but they are not specified to be discarded in the init of this class. 25% more memory efficient by not accumulating tn's

fl4health/metrics/efficient_metrics.py
```diff
@@ -271,7 +271,7 @@ def __call__(self, input: torch.Tensor, target: torch.Tensor) -> Scalar:
...
         dice = compute_dice_on_count_tensors(true_positives, false_positives, false_negatives, self.zero_division)
```
scarere, line 275: nit: You could instead pass tp, fp, tn, tn to `self.compute_from_counts` here. Doesn't save any lines of code but in the future if any changes are made, they'll be reflected implicitly in the call method

fl4health/metrics/utils.py
```diff
@@ -142,10 +146,6 @@ def align_pred_and_target_shapes(
 
         targets = torch.Tensor([0, 0, 1], [0, 1, 0])
 
-    NOTE: This function ASSUMES label-index encoding if the shapes are misaligned. This assumption doesn't necessarily
```
scarere: So I was looking into if there was a way to support not making this assumption (ie support the implicit labelling in the binary case) in a way that was more robust than what I was doing before. Interestingly enough, google search  AI overview suggested the following for determining if a float tensor contains only integers:
```python
int_tensor = float_tensor.int()
torch.equal(int_tensor, float_tensor)
```
Does this seem more robust to you than the previous approach? I'm not super familiar with the fine details of how pytorch stores integer values of floats in memory or how it converts floats into integers.

scarere: Doing this removes the need for the binary classification metric to make the assumption that preds and targets are the same shape

tests/metrics/efficient_metrics/compound_metrics/__init__.py
scarere, line 1: Seems like compound metrics should be outside the scope of efficient metrics

scarere (commented):
Mostly minor comments on documentation. Noted one potential opportunity to refactor again by merging the postprocessing functionality of the binary and multiclass subclasses

fl4health/metrics/efficient_metrics_base.py
emersodb: Done.

fl4health/metrics/efficient_metrics_base.py
emersodb: Added more commentary. For this class, I don't want to be extremely explicit, because the shapes are going to vary depending on the implementing class. That is, in the binary and multiclass settings, the shapes are subtly different.

fl4health/metrics/efficient_metrics_base.py
emersodb: I don't think that is necessary. It's a base class (an ABC in fact). So this is strongly implied just by structure alone.

emersodb: Done

emersodb: mmm...I may be missing something here in your suggestion, but the post process for the binary metrics class is quite a bit more nuanced due to the variations in shape that is taken care of in the post processing. There is some overlap when the count matrices are 2d, but a lot of the code remains unique to the 2D case. 

There is likely a bit of abstraction we can do here to save a touch of code, but I don't think it's going to save us a lot

fl4health/metrics/efficient_metrics.py
emersodb: That's fair. Let me enforce that

emersodb: Ah...now I remember why that's a bad idea. If the user sets `pos_label = 0`, i.e. they want the statistics with respect to the negative class rather than the positive class, we need to know the true negatives as this is the "true positives" from the perspective of the negative class.

fl4health/metrics/efficient_metrics.py
emersodb: Yeah I think that's a good suggestion.

fl4health/metrics/utils.py
emersodb: We still support implicit labeling, but your preds also have to be implicit in that case.

This conversion may still have fluctuation issues. That is, `torch.Tensor([0.99, 1.01]).int()` produces `tensor([0, 1], dtype=torch.int32)`. I may be being way too cautious here with worrying about numerical fluctuations for this. However, I still don't think its a big deal for the first implementation to assume you have everything aligned.

If we want to investigate this more thoroughly in a follow up to loosen this restriction, let's do it. My recommendation for now is that we leave as is though.

tests/metrics/efficient_metrics/compound_metrics/__init__.py
emersodb: Yeah I think you're right. My apologies. Was refactoring too fast.

fl4health/metrics/efficient_metrics.py
scarere: Hmmmn I see. One workaround would be `discard = 'tn' if pos_label else 'tp'`. However that's a bit nuanced and might not be super clear. I think at the very least a comment can be added stating why 'tn' can't be discarded.

emersodb: mmm...yeah we can do that. Since this is a specific metric implementation, that should be straightforward. I'll add that and document why we have the nuance in a comment.

fl4health/metrics/efficient_metrics_base.py
emersodb: Adding in a note here after discussion with you and Andrei in the call 🙂

fl4health/metrics/efficient_metrics.py
emersodb: Done!

fl4health/metrics/efficient_metrics_base.py
scarere: Ok i see whats happening now, the binary `_post_process` has the default behaviour of always removing the background class (It ony returns the counts for the pos_label class). I was really confused at first but I just realized that the cases you describe are the possible shapes of the incoming count_tensor (not the outgoing one returned by the function). Maybe that can be made clearer. Also if I'm correct on that, then its not really explained what the "appropriate format" is for the outgoing tensor or what the purpose of the post processing is. From reading the code it seems like you only want to return the counts for the positive class and you always want the output to be 2d with an empty dim in the second dimension, is that correct? Looking at the compute_from_counts method I see that this is in fact the case. What is the purpose of the empty second dimension? Anyways I see why we can't just use this for the multiclass subclass as we don't want results with respect only to a single class. Happy to leave the code as is but think some extra documentaion in the `_post_process` method might be warranted

emersodb: Ah gotcha. I'll be more clear in the function documentation that those are incoming shapes. 

Your interpretation of it always being a 2D tensor is almost correct. When the batch dimension isn't specified, everything collapses to a tensor with a single element (which is mentioned in the docs of `compute_from_counts` that you observed). I'm being a bit pedantic, I suppose, in forcing them two be of shape `(n_samples, 1)`, but in my head they are essentially column vectors. So I forced them to be that, if it made sense (i.e. `batch_dim` was specified). Some naturally have that shape, others I'm doing it artificially.

I strongly suspect that this has extremely low (if not no) overhead from a tensor perspective (if pytorch is doing their job). So I think this is fine. However, if you disagree, let me know.

I will 100% add more docs on the output shapes as well!

emersodb: Alright. More documentation added. Feel free to take a look at the additions.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -28,12 +29,13 @@ def __init__(
...
-        classes inheriting from this class to form the basis of efficient classification metrics:
-        BinaryClassificationMetric and MultiClassificationMetric. These handle implementation of the
-        ``count_tp_fp_fn_tn`` method.
+        A Base class for efficiently computing classification metrics that can be calculated using the true positives
+        (tp), false positive (fp), false negative (fn) and true negative (tn) counts.
+
+        How these values are counted is left to the inheriting class along with how they are composed together for the
+        final metric score. There are two classes inheriting from this class to form the basis of efficient
+        classification metrics: BinaryClassificationMetric and MultiClassificationMetric. These handle implementation
+        of the ``count_tp_fp_fn_tn`` method.
```
lotif, line 38: I think this function name changed based on one of my comments?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -157,15 +166,15 @@ def _prepare_counts_from_preds_and_targets(
         false_positives = (preds * (1 - targets)) if not self.discard_fp else torch.tensor([])
         false_negatives = ((1 - preds) * targets) if not self.discard_fn else torch.tensor([])
         true_negatives = ((1 - preds) * (1 - targets)) if not self.discard_tn else torch.tensor([])
-        return true_positives, false_positives, false_negatives, true_negatives
+        return true_positives, false_positives, true_negatives, false_negatives
```
lotif, line 169: The docstring on line 162 must change in order to reflect this change as well.

lotif (approved):
A few minor comments related to one of my previous comments, approving anyways it to avoid blocking :)

fl4health/metrics/efficient_metrics_base.py
emersodb: Yup...good catch. Will find an replace this!

fl4health/metrics/efficient_metrics_base.py
emersodb: Yes. Good call. I tried to do this everywhere, but clearly missed pieces.

PR #374, opened 2025-04-17
Upgrading flower to ~1.17.0

lotif:
# PR Type
Fix

# Short Description

Clickup Ticket(s): NA

Upgrading `flwr` to `~1.17.0`. All tests pass so no further changes should be necessary right now.

# Tests Added

NA


pyproject.toml
```diff
@@ -15,7 +15,7 @@ classifiers = [
 python = ">=3.10.0,<3.11"
 numpy = "^1.24"
 pandas = "^2.0"
-flwr = "1.9.0"
+flwr = "~1.17.0"
```
emersodb, line 18: Do we want to use the ~ here? It would mean we might upgrade with minor versions right? Just don't want to accidentally upgrade to something that breaks.

lotif: The `~` would allow patch updates, with the last digit like `1.17.1` and such (see https://python-poetry.org/docs/dependency-specification/#compatible-release-requirements). Minor versions (the middle one) and major versions (the first one) would both be blocked by this.

That statement is equivalent to `>=1.17.0, <1.18.0`.

fatemetkl (approved):
LGTM!

pyproject.toml
nerdai: Yeah patches should be non-breaking, if they follow semvar which I would assume that they do. I'm good with this. If stuff breaks on a patch, then we can always pin to a specific version after.

nerdai (approved):
LGTM!

This is non-breaking right since we're just in the deprecation period  of `start_client` and `start_server`?

lotif:
@nerdai yep, non breaking. Those functions are deprecated but not removed so we only get a warning for now.

Actually, `1.18.0` has just been released today, I'll update this PR to this version once the tests pass in my local and if everything is green here I'll merge.

nerdai:
Sgtm! @lotif

PR #377, opened 2025-04-22
Fixing rxrx1 experiments model

sanaAyrml:
# PR Type
[Fix]

# Short Description

Fixed the model for RxRx1 experiments by modifying the final layer of ResNet-18 to perform 50-way classification.

# Tests Added

NA


nerdai (commented):
looks good, just wondering if it would be better to be a bit more dry here and outsource the changing of model head to a utils function?

`research/rxr1/utils.py`
```
def change_head(model: nn.Module, ...) -> nn.Module:
        model = models.resnet18(pretrained=True)
        num_classes = 50
        in_features = model.fc.in_features
        model.fc = nn.Linear(in_features, num_classes)
        return model
```

and we just use this util function in all of these spots.

nerdai (approved):
LGTM!

PR #380, opened 2025-04-28
Small MMD experiment edits

emersodb:
# PR Type
Other

Small edits to the experiments. Also adding peft to the main dependencies because it's in the core utils

Also separating out the peft extraction functionality from the main extraction due to deepspeed being brought in. Even if you're not using the peft extraction, it gets imported because it's in the same utils file, which causes issues for all our examples on GPU if the CUDA_HOME env variable is not set.

# Tests Added

N/A


emersodb:
Note: I had to inject some `type: ignores` due to this issue with transformers: https://github.com/huggingface/transformers/issues/37339

research/rxrx1/ditto_deep_mmd/client.py
```diff
@@ -179,17 +179,11 @@ def get_model(self, config: Config) -> nn.Module:
...
-    post_aggregation_last_checkpoint_name = f"post_aggregation_client_{args.client_number}_last_model.pkl"
     checkpoint_and_state_module = ClientCheckpointAndStateModule(
         pre_aggregation=[
             BestLossTorchModuleCheckpointer(checkpoint_dir, pre_aggregation_best_checkpoint_name),
             LatestTorchModuleCheckpointer(checkpoint_dir, pre_aggregation_last_checkpoint_name),
         ],
-        post_aggregation=[
-            BestLossTorchModuleCheckpointer(checkpoint_dir, post_aggregation_best_checkpoint_name),
-            LatestTorchModuleCheckpointer(checkpoint_dir, post_aggregation_last_checkpoint_name),
-        ],
```
sanaAyrml, line 192: Here, you’ve only removed the extra checkpointing from deep_mmd on RxRx1—should we apply this change across all models?

emersodb:
> One thing I suggest adding to x.slrm files of cifar10 is: `export CUBLAS_WORKSPACE_CONFIG=:4096:8` Otherwise, it results in an error caused by setting the random seed.

Does it? I don't think I experienced this error when I was running the experiments.

research/rxrx1/ditto_deep_mmd/client.py
emersodb: Sure. I can apply that across the board.

sanaAyrml:
Oh I always get it :(

sanaAyrml (approved):
LGTM! Thanks for edits.

PR #383, opened 2025-04-30
First attempt at narrowing permissions for the workflows.

emersodb:
# PR Type
Fix

# Short Description

We have some CodeQL warnings/failures around not narrowing permissions for our github workflows. This is my first attempt at doing that. The documentation for this stuff is not great. So I might be trial and erroring this a bit.

# Tests Added

N/A


emersodb:
> Interesting. How can we check that this works? If CODEQL doesn't complain? This looks right to me...

Lol...good question. The first step for me was making sure all the workflows pass. The second step is figuring out how to run CodeQL on this to see if the issues are clean before merging to main

nerdai:
Sgtm!

PR #385, opened 2025-05-08
[Feature] Add `DittoPersonalizedMixin`

nerdai:
# PR Type
Feature

# Short Description

Continuation of using mixins to implement pFL methods. This PR adds the `DittoPersonalizedMixin` which subclasses the previously merged `AdaptiveDriftConstrainedMixin`.

# Tests Added

### DIFF
--- a/fl4health/mixins/personalized/__init__.py
+++ b/fl4health/mixins/personalized/__init__.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 from fl4health.clients.basic_client import BasicClient
-from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin
+from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin, DittoPersonalizedProtocol
 
 
 class PersonalizedModes(str, Enum):
@@ -30,4 +30,10 @@ def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedMode
         raise ValueError("Unrecognized personalized mode.")
 
 
-__all__ = ["DittoPersonalizedMixin", "PersonalizedModes", "PersonalizedMixinRegistry", "make_it_personal"]
+__all__ = [
+    "DittoPersonalizedMixin",
+    "DittoPersonalizedProtocol",
+    "PersonalizedModes",
+    "PersonalizedMixinRegistry",
+    "make_it_personal",
+]
--- a/fl4health/mixins/personalized/__init__.py
+++ b/fl4health/mixins/personalized/__init__.py
@@ -1,0 +1,32 @@
+from enum import Enum
+
+from fl4health.clients.basic_client import BasicClient
+from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin
+
+
+class PersonalizedModes(str, Enum):
+    DITTO = "ditto"
+
+
+PersonalizedMixinRegistry = {"ditto": DittoPersonalizedMixin}
+
+
+def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes) -> type[BasicClient]:
+    if mode == "ditto":
+
+        return type(
+            f"Ditto{client_base_type.__name__}",
+            (
+                PersonalizedMixinRegistry[mode],
+                client_base_type,
+            ),
+            {
+                # Special flag to bypass validation
+                "_dynamically_created": True
+            },
+        )
+    else:
+        raise ValueError("Unrecognized personalized mode.")
+
+
+__all__ = ["DittoPersonalizedMixin", "PersonalizedModes", "PersonalizedMixinRegistry", "make_it_personal"]
--- a/fl4health/mixins/personalized/__init__.py
+++ b/fl4health/mixins/personalized/__init__.py
@@ -11,7 +11,8 @@ class PersonalizedModes(str, Enum):
 PersonalizedMixinRegistry = {"ditto": DittoPersonalizedMixin}
 
 
-def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes) -> type[BasicClient]:
+def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes | str) -> type[BasicClient]:
+    """A mixed class factory for converting basic clients to personalized versions."""
     if mode == "ditto":
 
         return type(
--- a/fl4health/mixins/personalized/__init__.py
+++ b/fl4health/mixins/personalized/__init__.py
@@ -4,16 +4,16 @@ from fl4health.clients.basic_client import BasicClient
 from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin, DittoPersonalizedProtocol
 
 
-class PersonalizedModes(Enum):
+class PersonalizedMode(Enum):
     DITTO = "ditto"
 
 
-PersonalizedMixinRegistry = {PersonalizedModes.DITTO: DittoPersonalizedMixin}
+PersonalizedMixinRegistry = {PersonalizedMode.DITTO: DittoPersonalizedMixin}
 
 
-def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes) -> type[BasicClient]:
+def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedMode) -> type[BasicClient]:
     """A mixed class factory for converting basic clients to personalized versions."""
-    if mode == PersonalizedModes.DITTO:
+    if mode == PersonalizedMode.DITTO:
 
         return type(
             f"Ditto{client_base_type.__name__}",
@@ -33,7 +33,7 @@ def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedMode
 __all__ = [
     "DittoPersonalizedMixin",
     "DittoPersonalizedProtocol",
-    "PersonalizedModes",
+    "PersonalizedMode",
     "PersonalizedMixinRegistry",
     "make_it_personal",
 ]
--- a/fl4health/mixins/personalized/__init__.py
+++ b/fl4health/mixins/personalized/__init__.py
@@ -4,16 +4,16 @@ from fl4health.clients.basic_client import BasicClient
 from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin, DittoPersonalizedProtocol
 
 
-class PersonalizedModes(str, Enum):
+class PersonalizedModes(Enum):
     DITTO = "ditto"
 
 
-PersonalizedMixinRegistry = {"ditto": DittoPersonalizedMixin}
+PersonalizedMixinRegistry = {PersonalizedModes.DITTO: DittoPersonalizedMixin}
 
 
-def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes | str) -> type[BasicClient]:
+def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes) -> type[BasicClient]:
     """A mixed class factory for converting basic clients to personalized versions."""
-    if mode == "ditto":
+    if mode == PersonalizedModes.DITTO:
 
         return type(
             f"Ditto{client_base_type.__name__}",
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -138,6 +138,7 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         config["for_global"] = True
         return self.get_model(config).to(self.device)
 
+    @ensure_protocol_compliance
     def get_optimizer(self: DittoPersonalizedProtocol, config: Config) -> dict[str, Optimizer]:
         if self.global_model is None:
             # try set it here
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -118,10 +118,22 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
 
     @property
     def optimizer_keys(self: DittoPersonalizedProtocol) -> list[str]:
-        """Returns the optimizer keys."""
+        """Property for optimizer keys.
+
+        Returns:
+            list[str]: list of keys for the optimizers dictionary.
+        """
         return ["local", "global"]
 
     def _copy_optimizer_with_new_params(self: DittoPersonalizedProtocol, original_optimizer: Optimizer) -> Optimizer:
+        """Helper method to make a copy of the original optimizer for the global model.
+
+        Args:
+            original_optimizer (Optimizer): original optimizer of the underyling `BasicClient`.
+
+        Returns:
+            Optimizer: a copy of the original optimizer to be used by the global model.
+        """
         OptimClass = original_optimizer.__class__
         state_dict = original_optimizer.state_dict()
 
@@ -159,6 +171,12 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
 
     @ensure_protocol_compliance
     def get_optimizer(self: DittoPersonalizedProtocol, config: Config) -> dict[str, Optimizer]:
+        """
+        Returns a dictionary with global and local optimizers with string keys "global" and "local" respectively.
+
+        Args:
+            config (Config): The config from the server.
+        """
         if self.global_model is None:
             # try set it here
             self.global_model = self.get_global_model(config)  # is this the same config?
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -481,13 +481,13 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
             log(INFO, "Using '_special_compute_loss_and_additional_losses' to compute loss")
             global_loss, _ = self._special_compute_loss_and_additional_losses(global_preds, features, target)
 
-            # Compute local model loss + ditto constraint term
+            # Compute local model loss
             local_loss, _ = self._special_compute_loss_and_additional_losses(local_preds, features, target)
 
         else:
             global_loss = self.criterion(global_preds, target)
 
-            # Compute local model loss + ditto constraint term
+            # Compute local model loss
             local_loss = self.criterion(local_preds, target)
 
         additional_losses = {"local_loss": local_loss.clone(), "global_loss": global_loss}
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -132,13 +132,11 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         # store initial_lr to be used with schedulers
         initial_lr = param_group.pop("initial_lr", param_group["lr"])
 
-        optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params", "lr")}
+        optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params")}
         assert self.global_model is not None
-        global_optimizer = OptimClass(
-            self.global_model.parameters(), lr=param_group["lr"], **optimizer_kwargs
-        )  # type:ignore [call-arg]
+        global_optimizer = OptimClass(self.global_model.parameters(), **optimizer_kwargs)
 
-        # If you need to maintain the initial_lr for schedulers:
+        # maintain initial_lr for schedulers
         for param_group in global_optimizer.param_groups:
             param_group["initial_lr"] = initial_lr
 
@@ -157,7 +155,6 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         Returns:
             nn.Module: The PyTorch model serving as the global model for Ditto
         """
-        config["for_global"] = True
         return self.get_model(config).to(self.device)
 
     @ensure_protocol_compliance
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -409,7 +409,7 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
             retval.update(**{f"local-{k}": v for k, v in local_preds.items()})
             return retval, {}
         else:
-            raise ValueError(f"Unsupported pred type: {type(global_preds)}.")
+            raise ValueError(f"Unsupported pred types. Global: {type(global_preds)}, local: {type(local_preds)}")
 
     def _extract_pred(self, kind: str, preds: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
         if kind not in ["global", "local"]:
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -26,19 +26,19 @@ class DittoPersonalizedProtocol(AdaptiveDriftConstrainedProtocol, Protocol):
     optimizer_keys: list[str]
 
     def get_global_model(self, config: Config) -> nn.Module:
-        pass
+        pass  # pragma: no cover
 
     def _copy_optimizer_with_new_params(self, original_optimizer: Optimizer) -> Optimizer:
-        pass
+        pass  # pragma: no cover
 
     def set_initial_global_tensors(self) -> None:
-        pass
+        pass  # pragma: no cover
 
     def _extract_pred(self, kind: str, preds: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
-        pass
+        pass  # pragma: no cover
 
     def safe_global_model(self) -> nn.Module:
-        pass
+        pass  # pragma: no cover
 
 
 class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -103,9 +103,14 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         )
 
     def safe_global_model(self: DittoPersonalizedProtocol) -> nn.Module:
-        """Convenience method for accessing optional `global_model`.
+        """Convenient accessor for the global model.
 
-        Raises: ValueError if global model has not yet been set.
+        Raises:
+            ValueError: If the `global_model` attribute has not yet been set, we
+            will raise an error.
+
+        Returns:
+            nn.Module: the global model if it has been set.
         """
         if self.global_model:
             return self.global_model
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -43,7 +43,7 @@ class DittoPersonalizedProtocol(AdaptiveDriftConstrainedProtocol, Protocol):
 
 class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
 
-    def __init__(self, *args: Any, **kwargs: Any):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         # Initialize mixin-specific attributes
         self.global_model: torch.nn.Module | None = None
 
@@ -57,7 +57,7 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         if not isinstance(self, BasicClientProtocolPreSetup):
             raise RuntimeError("This object needs to satisfy `BasicClientProtocolPreSetup`.")
 
-    def __init_subclass__(cls, **kwargs: Any):
+    def __init_subclass__(cls, **kwargs: Any) -> None:
         """This method is called when a class inherits from AdaptiveMixin"""
         super().__init_subclass__(**kwargs)
 
--- a/fl4health/mixins/personalized/ditto.py
+++ b/fl4health/mixins/personalized/ditto.py
@@ -130,9 +130,9 @@ class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
         param_group = state_dict["param_groups"][0]
 
         # store initial_lr to be used with schedulers
-        initial_lr = param_group.pop("initial_lr", param_group["lr"])
+        initial_lr = param_group.get("initial_lr", param_group["lr"])
 
-        optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params")}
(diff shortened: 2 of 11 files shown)


nerdai:
@emersodb: as discussed here is the WIP PR for adding the ditto mixin. Will add unit tests to get coverage checks, but in the meantime you can take a look around the PR and add your review.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        Returns:
+            nn.Module: The PyTorch model serving as the global model for Ditto
+        """
+        config["for_global"] = True
+        return self.get_model(config).to(self.device)
+
+    def get_optimizer(self: DittoProtocol, config: Config) -> dict[str, Optimizer]:
+        if self.global_model is None:
+            # try set it here
+            self.global_model = self.get_global_model(config)  # is this the same config?
```
emersodb, line 144: Any reason we're doing this here? It seems a bit odd to mix this responsibility in here, but I might be missing why this is required.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            log(INFO, f"global model set: {type(self.global_model).__name__} within `get_optimizer`")
+
+        # Note that the global optimizer operates on self.global_model.parameters()
+        optimizer = super().get_optimizer(config=config)  # type: ignore[safe-super]
+        if isinstance(optimizer, dict):
+            try:
+                original_optimizer = next(el for el in optimizer.values() if isinstance(el, Optimizer))
```
emersodb, line 151: Is this necessary? That is, the return value of get_optimizer should be `Optimizer | dict[str, Optimizer]`. Some I'm not sure why we would fail to get one in this case?

fl4health/mixins/personalized/ditto.py
emersodb: The in-line comment is a bit curious as well 😂

fl4health/mixins/personalized/ditto.py
emersodb: I'm also slightly confused why we're doing a copy in the dictionary setting. Should we not just try matching keys and if not present thow?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+    def _copy_optimizer_with_new_params(self: DittoProtocol, original_optimizer: Optimizer) -> Optimizer:
+        OptimClass = original_optimizer.__class__
+        state_dict = original_optimizer.state_dict()
+
+        # Extract hyperparameters from param_groups
+        # We only take the first group's hyperparameters, excluding 'params' and 'lr'
+        param_group = state_dict["param_groups"][0]
+
+        # store initial_lr to be used with schedulers
+        initial_lr = param_group.pop("initial_lr", param_group["lr"])
```
emersodb, line 111: Any reason to use pop here? Do we anticipate the `initial_lr` to not be present in some settings?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+
+        optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params", "lr")}
+        assert self.global_model is not None
+        global_optimizer = OptimClass(
```
emersodb, line 115: Implicitly, this forces the optimizers to have the same hyper-parameters. However, the most general form of Ditto does not force that restriction.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            self.global_model.parameters(), lr=param_group["lr"], **optimizer_kwargs
+        )  # type:ignore [call-arg]
```
emersodb, line 117: Is the typing issue here related to the kwargs?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        # We only take the first group's hyperparameters, excluding 'params' and 'lr'
+        param_group = state_dict["param_groups"][0]
+
+        # store initial_lr to be used with schedulers
+        initial_lr = param_group.pop("initial_lr", param_group["lr"])
+
+        optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params", "lr")}
+        assert self.global_model is not None
+        global_optimizer = OptimClass(
+            self.global_model.parameters(), lr=param_group["lr"], **optimizer_kwargs
```
emersodb, line 116: I'm not an expert on optimizer state, but this feels like we're assuming that the `lr` should be the same across all param groups, which isn't necessarily the case if we've taken any steps. However, if we haven't taken any steps, why bother using anything other than the initial LR? I may be wrong though.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        The global model should be the same architecture as the local model so we reuse the ``get_model`` call. We
+        explicitly send the model to the desired device. This is idempotent.
+
+        Args:
+            config (Config): The config from the server.
+
+        Returns:
+            nn.Module: The PyTorch model serving as the global model for Ditto
+        """
+        config["for_global"] = True
```
emersodb, line 138: Is this guy used anywhere?

fl4health/mixins/personalized/utils.py
```diff
@@ -0,0 +1,16 @@
...
+from typing import Any
+
+import wrapt
+
+from fl4health.clients.basic_client import BasicClient
+
+
+@wrapt.decorator
+def ensure_protocol_compliance(func: Callable, instance: Any | None, *args: Any, **kwargs: Any) -> None:
+    # validate self is a BasicClient
```
emersodb, line 11: Any reason we bother doing this reassignment to self?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        setting up the global model.
+
+        Args:
+            config (Config): The config from the server.
+        """
+        try:
+            self.global_model = self.get_global_model(config)
+            log(INFO, f"global model set: {type(self.global_model).__name__}")
+        except AttributeError:
+            log(INFO, "Couldn't set global model before super().setup_client(). Will try again within that setup.")
```
emersodb, line 188: I'm definitely missing something, but why wouldn't this work and what might change to make it work later?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            pass
+        # The rest of the setup is the same
+        super().setup_client(config)  # type:ignore [safe-super]
+        # Need to setup the global model here as well. It should be the same architecture as the local model.
```
emersodb, line 192: Should this comment move somewhere rather than sort of dangling at the end of the function?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+    def initialize_all_model_weights(self: DittoProtocol, parameters: NDArrays, config: Config) -> None:
+        """
+        If this is the first time we're initializing the model weights, we initialize both the global and the local
+        weights together.
+
+        Args:
+            parameters (NDArrays): Model parameters to be injected into the client model
+            config (Config): The config is sent by the FL server to allow for customization in the function if desired.
+        """
+        parameter_exchanger = cast(FullParameterExchanger, self.parameter_exchanger)
```
emersodb, line 275: Is there a reason you need cast here? Casting is always a pretty last resort for me, akin to ignores.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+from fl4health.mixins.personalized.utils import ensure_protocol_compliance
+from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
+from fl4health.utils.config import narrow_dict_type
+from fl4health.utils.losses import EvaluationLosses, TrainingLosses
+from fl4health.utils.typing import TorchFeatureType, TorchInputType, TorchPredType, TorchTargetType
+
+
+@runtime_checkable
+class DittoProtocol(AdaptiveDriftConstrainedProtocol, Protocol):
+    global_model: torch.nn.Module | None
```
emersodb, line 25: Is there a reason we've added the `global_model` as optional here? It isn't really optional for Ditto.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            tuple[TorchPredType, TorchFeatureType]: A tuple in which the first element contains predictions indexed by
+            name and the second element contains intermediate activations index by name. For Ditto, we only need the
+            predictions, so the second dictionary is simply empty.
+
+        Raises:
+            ValueError: Occurs when something other than a tensor or dict of tensors is returned by the model
+                forward.
+        """
+
+        if hasattr(self, "_predict"):
```
emersodb, line 368: Where is this function ever defined? Does the user define this?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+                # If input is a dictionary, then we unpack it before computing the forward pass.
+                # Note that this assumes the keys of the input match (exactly) the keyword args
+                # of the forward method.
+                global_preds = self.safe_global_model()(**input)
+                local_preds = self.model(**input)
+
+        # Here we assume that global and local preds are simply tensors
+        # TODO: Perhaps loosen this at a later date.
+        # assert isinstance(global_preds, torch.Tensor)
+        # assert isinstance(local_preds, torch.Tensor)
```
emersodb, line 387: The above comments are out of date?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        if isinstance(global_preds, torch.Tensor) and isinstance(local_preds, torch.Tensor):
+            return {"global": global_preds, "local": local_preds}, {}
+        elif isinstance(global_preds, dict) and isinstance(local_preds, dict):
+            retval = {f"global-{k}": v for k, v in global_preds.items()}
+            retval.update(**{f"local-{k}": v for k, v in local_preds.items()})
+            return retval, {}
+        else:
+            raise ValueError(f"Unsupported pred type: {type(global_preds)}.")
```
emersodb, line 395: This only reports the `global_preds` type. I know the models are essentially copies, but on the off chance the return values are different should we return both?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def _extract_pred(self, kind: str, preds: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
```
emersodb, line 397: Perhaps some docs?

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            - The tensor for the model loss
+            - A dictionary with ``local_loss``, ``global_loss`` as additionally reported loss values.
+        """
+
+        global_preds = self._extract_pred(kind="global", preds=preds)
+        local_preds = self._extract_pred(kind="local", preds=preds)
+
+        # Compute global model vanilla loss
+
+        if hasattr(self, "_special_compute_loss_and_additional_losses"):
```
emersodb, line 433: Is the defined anywhere? I'm not 100% sure of its envisioned use.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            log(INFO, "Using '_special_compute_loss_and_additional_losses' to compute loss")
+            global_loss, _ = self._special_compute_loss_and_additional_losses(global_preds, features, target)
+
+            # Compute local model loss + ditto constraint term
```
emersodb, line 437: I know you took this from my docs and its also out of date there too, but we don't actually compute the ditto constraint here.

emersodb: Same on line 443 below

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            TrainingLosses: An instance of ``TrainingLosses`` containing backward loss and additional losses indexed by
+            name. Additional losses includes each loss component and the global model
+            loss tensor.
+        """
+        # Check that both models are in training mode
+        assert self.safe_global_model().training and self.model.training
+
+        # local loss is stored in loss, global model loss is stored in additional losses.
+        loss, additional_losses = self.compute_loss_and_additional_losses(preds, features, target)
+        additional_losses = additional_losses or {}  # make mypy happy
```
emersodb, line 478: What was the issue for mypy here?

emersodb (commented):
A few comments throughout. I think some of them might be related to this PR setting up some scaffolding for follow ups, which might have spurred a few of the comments.

fl4health/mixins/personalized/__init__.py
```diff
@@ -0,0 +1,32 @@
+from enum import Enum
```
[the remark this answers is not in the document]
nerdai: Yeah, I have seen this adopted in practice, by engineers more experienced than me; and in other libraries. That is, I am treating this as a factory/registry, which I have seen be put in these module-level init files. I think so long as we're not baking in the actual logic here, then we're okay... but i'm also not married to this. If you really prefer to outsource this to some other module, say `factory` or `registry`, then I'm not picky. :)

fl4health/mixins/personalized/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+class PersonalizedModes(str, Enum):
+    DITTO = "ditto"
+
+
+PersonalizedMixinRegistry = {"ditto": DittoPersonalizedMixin}
+
+
+def make_it_personal(client_base_type: type[BasicClient], mode: PersonalizedModes) -> type[BasicClient]:
+    if mode == "ditto":
```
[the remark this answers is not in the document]
nerdai: The main advantage that I have used it for is code reduction, but you can also apply str operations to the enum directly if needed. For our users, I think code reduction is pretty nice, as most are lazy to import an enum when they  pass in a str instead. In our lib, we can create validations to ensure the passed in str is a correct member of the enum, which is what we do in FedRAG. And, if I'm not mistaken, if using Pydantic, then I think they've got some nice validations for this as well.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
+"""Ditto Personalized Mixin"""
```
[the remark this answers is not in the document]
nerdai: Yeah two reasons:

1. I used to have static checkers check for this, so I just go used to doing them.
2. Some automatic docs tools, will use this as the header/title in the api reference for this module.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+
+        Raises: ValueError if global model has not yet been set.
+        """
+        if self.global_model:
+            return self.global_model
+        raise ValueError("Cannot get global model as it not yet been set.")
+
+    @property
+    def optimizer_keys(self: DittoProtocol) -> list[str]:
+        """Returns the optimizer keys."""
```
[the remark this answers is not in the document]
nerdai: No worries. I'm more than happy to conform to our docs structure. Google, and for all methods? Or what's the standard?

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
nerdai: The main reason I think that these factories/registries are created here is because often all of the classes you need are already being re-imported (i.e., and being exposed in the public api "__all__")

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
nerdai: BTW, if we handle renames like that, then we should keep in mind to always add some backwards compatibility functionality so that we don't break things like that so easily.

fl4health/mixins/personalized/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+from fl4health.clients.basic_client import BasicClient
+from fl4health.mixins.personalized.ditto import DittoPersonalizedMixin
+
+
+class PersonalizedModes(str, Enum):
+    DITTO = "ditto"
+
+
+PersonalizedMixinRegistry = {"ditto": DittoPersonalizedMixin}
```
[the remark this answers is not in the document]
emersodb: True...I think if we want to limit to the enum types here, it's a little more "rigid" if we use the enum explicitly. If you disagree that's alright.

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
emersodb: Yeah that's fair. I get the argument for code reduction, but I also think the advantage of enums is "forced" adherence to some structure. So I tend to not "bail people out" with the string escape 😂

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
emersodb: That's fair. I honestly have no idea what best practices apply to `__init__.py` files. Seems kind of wild west ha.

fl4health/mixins/personalized/ditto.py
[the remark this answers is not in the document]
emersodb: Ah I see...I guess the later would be a reason to do it. Just kind of seems like "redundant" comments is all 🙂

fl4health/mixins/personalized/ditto.py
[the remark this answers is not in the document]
emersodb: Yeah google style docs is what we've been using. There's a nice vs code integration that you can setup if you look at the `## Code Documentation` section of the contributing.md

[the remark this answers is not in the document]
nerdai: Yeah, I got autodoc string already installed :) and its also set to Google already on my code. Do we doc all methods in this way or just ones that we want to have reference in the API docs, or some other criteria.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+        try:
+            super().__init__(*args, **kwargs)
+        except TypeError:
+            # if a parent class doesn't take args/kwargs
+            super().__init__()
+
+        if not isinstance(self, BasicClientProtocolPreSetup):
+            raise RuntimeError("This object needs to satisfy `BasicClientProtocolPreSetup`.")
+
+    def __init_subclass__(cls, **kwargs: Any):
```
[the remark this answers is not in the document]
nerdai: Oh i see, we also event annotate `__init__`. No problem, will add.

(In some other projects, I haven't annotated init or init_subclass and mypy i think by default was good with it).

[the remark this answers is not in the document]
nerdai: done!

fl4health/mixins/personalized/ditto.py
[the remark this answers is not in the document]
emersodb: I've been documenting all of them in this way unless they are trivial. However, I'm open to establishing a different norm.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+    def _extract_pred(self, kind: str, preds: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
+        pass
+
+    def safe_global_model(self) -> nn.Module:
+        pass
+
+
+class DittoPersonalizedMixin(AdaptiveDriftConstrainedMixin):
+
+    def __init__(self, *args: Any, **kwargs: Any):
```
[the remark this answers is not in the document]
nerdai: done!

fl4health/mixins/personalized/ditto.py
nerdai: NNunet.... we don't have the model at setup, and have that step to configure the architecture. So we have to do some of this funky logic.

nerdai: The in-line comment was actually a question I had on this. I think its the same config that we keep propagating forward as needed? Do we have any other config types that we pass around?

fl4health/mixins/personalized/ditto.py
nerdai: Nnunet. It's not set first.

fl4health/mixins/personalized/ditto.py
```diff
@@ -0,0 +1,527 @@
...
+            f"base classes implement BasicClient. This may cause runtime errors."
+        )
+        log(WARN, msg)
+        warnings.warn(
+            msg,
+            RuntimeWarning,
+        )
+
+    def safe_global_model(self: DittoProtocol) -> nn.Module:
+        """Convenience method for accessing optional `global_model`.
```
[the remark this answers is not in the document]
nerdai: done!

fl4health/mixins/personalized/ditto.py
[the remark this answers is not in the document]
nerdai: Trivial in the sense that the method is self-documenting i.e., perhaps by its obvious naming suggesting what its doing?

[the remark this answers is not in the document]
emersodb: Yeah. Everyone has their own definition of trivial I suppose, but that's what I meant 😂. I find "self-documenting" is a problematic adjective. Everyone thinks their code is self-documenting ha

fl4health/mixins/personalized/ditto.py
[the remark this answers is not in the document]
emersodb: Yeah mypy is good without it there, we've just been adding it I guess.

fl4health/mixins/personalized/ditto.py
emersodb: Gotcha. I see what you mean now. I don't really remember how we handled this for nnunet, because presumably this is also an issue for the local model optimizer?

Yes, that is **the** config. It packs in everything and is received by each function. Technically, we can add to it dynamically, but that gets messy fairly quickly because it relies on knowing the full flow etc. So we avoid that quite heavily.

fl4health/mixins/personalized/ditto.py
emersodb: This might be an artifact of the way inheritance works vs. mixins etc., but the NnUnetClient seems cool with the BasicClient not having the self.model property be optional? There's probably a complexity I'm missing though.

fl4health/mixins/personalized/ditto.py
nerdai: i think we could run into a problem if we include it as part of the `optimizer_kwargs` when creating the copy of the optimizer.

fl4health/mixins/personalized/ditto.py
nerdai: nnunet :)

fl4health/mixins/personalized/ditto.py
nerdai: thanks, I don't think this comment is needed, tbh.

fl4health/mixins/personalized/ditto.py
nerdai: Yeah, to make mypy happy lol. I'm also just confused with whether it should be full param exchanger with or without packing. I think in our ditto/adaptive clients we make use of both and set it to the parameter exchanger.

fl4health/mixins/personalized/ditto.py
nerdai: This is needed to be able to delegate the prediction method to the underlying `BasicClient` that is being personalized. This was definitely needed to work with the NNunet client implementation.

fl4health/mixins/personalized/ditto.py
nerdai: Yeah, I think I inherited this from dittoclient. But commented code is dead code. Removing it.

nerdai: removed!

fl4health/mixins/personalized/ditto.py
nerdai: done!

fl4health/mixins/personalized/ditto.py
nerdai: done!

fl4health/mixins/personalized/ditto.py
nerdai: This will be added in the `nnunet_client`. I am thinking to make this into a protocol, but needed because theirs some special business logic in `nnunet_client` that required delegation of `compute_loss_and_additional_losses` to the underlying class. Though, I do think this is a sensible thing to do in general. The form factor might need polishing, but this shouldn't affect the user api. In the next PR I'll take care of modifying the Nnunet client as such. This PR is already too big.

fl4health/mixins/personalized/ditto.py
nerdai: Ah okay, should be tho? Is this an issue with missing logic or just a mismatch in the comment?

fl4health/mixins/personalized/ditto.py
nerdai: I think `nnunet_client` or the other way around broke the interface for `compute_loss_and_additional_losses`? Specifically:

```
# from nnunet client
    def compute_loss_and_additional_losses(
        self,
        preds: TorchPredType,
        features: dict[str, torch.Tensor],
        target: TorchTargetType,
    ) -> tuple[torch.Tensor, dict[str, torch.Tensor] | None]:

# from ditto client
    def compute_loss_and_additional_losses(
        self,
        preds: TorchPredType,
        features: TorchFeatureType,
        target: TorchTargetType,
    ) -> tuple[torch.Tensor, dict[str, torch.Tensor]]:
```

I went with `nnunet_client` here cause that's what I was trying to make work during development. And so `additional_losses` can be `None` but we treat it as a dict. So, in the event it is `None` I fallback to `dict` so that mypy recognizes this as a dict and doesn't complain.

fl4health/mixins/personalized/ditto.py
emersodb: Oh weird...it seems like every optimizer would have that argument, so it would be in the base class. I guess not.

fl4health/mixins/personalized/ditto.py
emersodb: Lol...right.

fl4health/mixins/personalized/ditto.py
emersodb: Gotcha

fl4health/mixins/personalized/ditto.py
emersodb: Just a mismatch in the comment. The ditto constraint is added elsewhere.

fl4health/mixins/personalized/ditto.py
emersodb: mmm. I see.

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
nerdai: Yea me too... I just go with things I have seen before in libraries i've worked in, or have read source code in, and how past teammates have used `__init__.py`.

fl4health/mixins/personalized/ditto.py
emersodb: Gotcha...it's been a while since I pulled an optimizer apart like this.

fl4health/mixins/personalized/__init__.py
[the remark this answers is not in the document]
nerdai: Ahhhh i still think we have the forced adherence, but we just provide convenience at the same time. With proper validation we provide convenience to the user (and developer) without losing any of the benefits of using enums in Python?

I'm not married to any sort of way tbh for this library. So, I think I'll just use enum everywhere if that's the convention...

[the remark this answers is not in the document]
nerdai: made the mode a strict enum!

fl4health/mixins/personalized/ditto.py
nerdai: Okay, I think that's a good way to go about it -- it's simpler.

fl4health/mixins/personalized/ditto.py
nerdai: Honestly, I don't remember the last time I had to do this. I asked Claude and ChatGPT for help and re-jigged it to what made sense to me. That plus some good unit tests that I've added is enough, I hope.

fl4health/mixins/personalized/ditto.py
nerdai: Yeah, exactly. I just bundled everything in the `optimizer_kwargs` instead of explicitly spelling it out.

fl4health/mixins/personalized/ditto.py
nerdai: Yeah I saw that.

TBH, I haven't encountered code in this way before where class vars are defined/typed but not set and it be fine at instantiation. Like this would definitely break using something modern like Pydantic nowadays.

Typical patterns I've seen for this is have some internal attribute which is optional and set to None at init. And, when appropriate the property is set.

```
class MyClass

_model: torch.nn.Module | None = None

@property
def model(self) -> torch.nn.Module:
    if self._model:
        return self._model
    else:
        # either set it and return it, or raise an error

@model.setter
def model(self) -> None:
     ....
```

(I think in a future refactor, we probably want to consider adopting something more standard like this, and I think easier to develop with, if i'm being honest 🙂)

fl4health/mixins/personalized/ditto.py
nerdai: Gotcha. Adjusted the comment by removing the bit on the "+ dittto constraint term"

fl4health/mixins/personalized/ditto.py
nerdai: Okay, so this logic needs confirmation. The thing I'm aiming to do is copy the optimizer to be used for the global model. If it's a dictionary of optimizers, which one should I copy? Would there be a case where there are multiple optimizers for an underlying client?

This code just takes the first optimizer it encounters in the dict of optimizers and copies that....

emersodb: Perfectly fine.

emersodb (approved):
LGTM!

nerdai:
> LGTM!

wooohoo! thanks for the review -- sorry it's a bit nuanced.

PR #386, opened 2025-05-12
[chore] Fix `docs_publish.yml` permmissions

nerdai:
# PR Type
Other

# Short Description

Looks like our permissions were a bit too restrictive for our github pages publish action. The reason why we probably weren't seeing this fail in the last couple of weeks since making the change to permissions is that our docs didn't have anything new to upload. Our action only publishes to the `github_pages` branch if there is something new/different in the docs.

# Tests Added

I manually triggered the workflow using cli, and can confirm the workflow now works.

image

### DIFF
--- a/.github/workflows/docs_publish.yml
+++ b/.github/workflows/docs_publish.yml
@@ -22,7 +22,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


emersodb (approved):

emersodb:
Thanks Andrei

PR #390, opened 2025-05-21
Nerdai/ditto nnunet

nerdai:
# PR Type

# Short Description

# Tests Added


fl4health/clients/nnunet_client.py
```diff
@@ -228,7 +229,7 @@ def train_step(self, input: TorchInputType, target: TorchTargetType) -> tuple[Tr
 
         # As in the nnUNetTrainer, we implement mixed precision using torch.autocast and torch.GradScaler
         # Clear gradients from optimizer if they exist
-        self.optimizers["global"].zero_grad()
+        self.optimizers["local"].zero_grad()
```
scarere, line 232: Were changes made to the base client to ensure that this is the correct key even when not using ditto?

fl4health/clients/nnunet_client.py
```diff
@@ -315,7 +316,12 @@ def get_data_loaders(self, config: Config) -> tuple[DataLoader, DataLoader]:
         return train_loader, val_loader
 
     def get_model(self, config: Config) -> nn.Module:
-        return self.nnunet_trainer.network
+
+        for_global = config.get(FOR_GLOBAL_MODEL_KEY, False)
```
scarere, line 320: Is there a way we can abstract this into the base class if all we need to do is make a copy? Otherwise every time a new client is made and you want it to be compatible with ditto you'll have to do this check

fl4health/clients/nnunet_client.py
```diff
@@ -606,27 +612,14 @@ def setup_client(self, config: Config) -> None:
...
-        Additionally if device type is cuda, loss computed in mixed precision.
-
-        Args:
-            input (TorchInputType): The model inputs
-
-        Returns:
-            tuple[TorchPredType, dict[str, torch.Tensor]]: A tuple in which the first element model outputs indexed by
-            name. The second element is unused by this subclass and therefore is always an empty dict
-        """
+    def _predict(self, model: torch.nn.Module, input: TorchInputType) -> tuple[TorchPredType, dict[str, torch.Tensor]]:
```
scarere, line 615: Are there refactors we can make to basic client such that subclasses are automatically compatible to be mixed-in with ditto? This feels more like a workaround

fl4health/clients/nnunet_client.py
```diff
@@ -642,26 +635,28 @@ def predict(self, input: TorchInputType) -> tuple[TorchPredType, dict[str, torch
...
 
         Returns:
-            tuple[torch.Tensor, dict[str, torch.Tensor] | None]: A tuple where the first element is the loss and the
-            second element is an optional additional loss
+            tuple[TorchPredType, dict[str, torch.Tensor]]: A tuple in which the first element model outputs indexed by
+            name. The second element is unused by this subclass and therefore is always an empty dict
         """
+        return self._predict(self.model, input)
+
+    def _special_compute_loss_and_additional_losses(
```
scarere, line 654: Why the need for the new function here if the function signature is the same?

fl4health/clients/nnunet_client.py
```diff
@@ -742,8 +759,13 @@ def update_metric_manager(
             target (TorchTargetType): the targets generated by the dataloader to evaluate the preds with
             metric_manager (MetricManager): the metric manager to update
         """
+        preds = {k: v for k, v in preds.items() if "local" in k}
```
scarere, line 762: Maybe add comments explaining what is going on here. Without the context that this is to support pFL its kind of confusing.

fl4health/mixins/personalized/ditto.py
```diff
@@ -15,7 +15,7 @@
 from fl4health.mixins.core_protocols import BasicClientProtocolPreSetup
 from fl4health.mixins.personalized.utils import ensure_protocol_compliance
 from fl4health.parameter_exchange.full_exchanger import FullParameterExchanger
-from fl4health.utils.config import narrow_dict_type
+from fl4health.utils.config import FOR_GLOBAL_MODEL_KEY, narrow_dict_type
```
scarere, line 18: Going to assume DittoMixin has already been reviewed

examples/nnunet_pfl_example/README.md
```diff
@@ -0,0 +1,69 @@
+# NnUNetClient Example
```
scarere, line 1: Documentation in this readme needs to be updated. I would remove most stuff and just refer to the basic nnunet_example. Then have this readme focus mainly on the pfl stuf

examples/nnunet_pfl_example/client.py
```diff
@@ -0,0 +1,245 @@
...
+        default=None,
+        help="[OPTIONAL] Name of the client used to name client state checkpoint. \
+        Defaults to None, in which case a random name is generated for the client",
+    )
+    parser.add_argument(
+        "--personalized-strategy",
+        type=str,
+        required=False,
+        default=None,
+        help="[OPTIONAL] Personalized strategy to use. Can be 'ditto' or 'mr-mtl' \
```
scarere, line 212: technically doesn't yet support mt-mtl but i'm going to assume thats planned

scarere (commented):
I don't think I got to review the Ditto Mixin stuff so I'm missing a bit of context, but overall looks good. My main comments are just with the aim of making all BasicClients personalizable by default without any special consideration needed.

examples/nnunet_pfl_example/client.py
nerdai: Yeah, that's coming right after "ditto" lands. But thanks for catching, will remove the `mr-mtl`.

examples/nnunet_pfl_example/README.md
nerdai: Great idea!

fl4health/mixins/personalized/ditto.py
nerdai: Yeah, sorry you didn't get in there. You can take a look at the code tho and always submit a PR/issue if you see something wonky.

This `FOR_GLOBAL_MODEL_KEY` is needed to pass thru to `get_model(config: Config)`. In nnunet the `get_model()` returns `self.nnunet_trainer.network`, but we need a copy of this for the global model ditto and mr-mtl.

An alternative could be to consider the call stack and see if the call includes `.get_global_model()`, but I thought that might be to complex and passing in a var in Config seemed like a better idea.

fl4health/clients/nnunet_client.py
nerdai: This is cause in ditto mixin we will override `compute_loss_and_additional_losses`, unless the object has a `_special_compute_loss_and_additional_losses` method. Essentially, what its saying is that the parents `compute_loss_and_additional_losses` is what should be invoked.

I think this can be polished up a bit though. I just don't know yet how best to do that. So, I just made a "private" method that the mixin calls until then.

Though I'm to changing this if we like a better way. For example, should we define another mixin? Or define a new ClassVar on `BasicClient` like `__DONT_OVERRIDE = [compute_loss_and_additional_losses]`.

fl4health/clients/nnunet_client.py
nerdai: I think `BasicClient` should support these helper functions that allow for dependency injection of `model`. And we should probably refactor `BasicClient` in a subsequent PR, and move all clients to this. Though I think that's its own separate PR.

fl4health/clients/nnunet_client.py
nerdai: Yah, so the main thing here is that the original interface is fine so long as the model that's being returned is a net new instantiation, i.e. if its a factory method.

So, we could either change this to `model_factory()`, which is in many cases how we use it versus `get_model()` which seems more like an accessor than a factory. I think these are valuable refactors to consider, though I'm unsure if we should do them in this PR.

My approach here was because there's a lot of moving parts between existing `DittoClient`,`BasicClient` and `NnunetClient` that I didn't want to break things and in the existing interfaces too much when implementing this new "feature".

I tried to preserve the option to clean up internal implementations without breaking things for the user... and i think we have that?

fl4health/clients/nnunet_client.py
nerdai: Yeah... I'm a bit confused on how to best approach this. Our smoke tests fail because of this name change apparently, so need to investigate that. I didn't know this key was so embedded in the base client?

fl4health/clients/nnunet_client.py
nerdai: Yea good point. I'm going to wait and see how to resolve the "global" vs "local" optimizer key stuff.

Ditto refers to the "global" as for the global model and "local" is for the clients local model...

fl4health/clients/nnunet_client.py
scarere: hmmn maybe you modify the base client methods to accept a key as an argument? And then set the default for that argument to be "global". But the ditto mixins could change the calls to the train_step methods to ensure the correct model is being trained? Or maybe the default key should be changed to local or some other key for all clients? Technically even in FedAvg they're not optimizing the global model, they're optimizing their local copy of it.

fl4health/clients/nnunet_client.py
scarere: I'm fine with leaving it as is for now. I think once you guys get confirmation that FL4Health is part of the Vector roadmap moving forward it'd be worth finally doing a giant library refactor/overhaul and updating the major from 0 to 1.

fl4health/clients/nnunet_client.py
scarere: Unless I'm missing something about how the inheritance works, but is there a way to just call the methods of the client that ditto is mixed in with? If so there should not be much need to modify any of the clients, we can just wrap all of their methods as necessary to enable ditto

fl4health/clients/nnunet_client.py
scarere: Again maybe rather than just having a assumed key in all of our clients, the key should be an argument?

fl4health/clients/nnunet_client.py
nerdai: Yeah, the super call could work, but I'm not sure we should even be doing this. I think the ditto mixin's `train_step` should just call the supers `train_step` two times, once for local and another for global. If invoking the supers train step, then we shouldn't be overriding `compute_loss_and_additional_losses` nor should we override any methods within the train step logic.

I need to clean up the mixin to delegate only the required methods and not override everything, I think.

fl4health/mixins/personalized/ditto.py
scarere: Reiterating on the idea of the mixin just wrapping client methods, why not just have the dittomixin retrieve the model using get_model(), check if the second copy has been made, if not it makes it and then it returns both models? Or i see you already have a get_global_model method for Ditto, why not:
```python
#pseudocode
class DittoMixin():
    def get_global_model(self, config: Config) -> nn.Module:
        if self.global_model is not None:
            self.global_model = deepcopy(self.get_model(config))
        return self.global_model
```

fl4health/mixins/personalized/ditto.py
```diff
@@ -74,6 +74,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         if not isinstance(self, BasicClientProtocolPreSetup):
             raise RuntimeError("This object needs to satisfy `BasicClientProtocolPreSetup`.")
 
+        #
```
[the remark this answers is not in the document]
nerdai: artifact of me moving things around lol. This PR is def still in Draft :)

fl4health/mixins/adaptive_drift_constrained.py
```diff
@@ -148,7 +155,26 @@ def set_parameters(
 
         super().set_parameters(server_model_state, config, fitting_round)  # type: ignore[safe-super]
 
-    def compute_training_loss(
+    def train_step(
```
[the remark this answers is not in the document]
nerdai: I would say for now, given the snags I've run into, probably best not to look at this PR. I may even open a new one once I have this working in the correct way. If you look now unfortunately a lot of it won't make sense, as I'm still trying to make sense of it...

fl4health/mixins/adaptive_drift_constrained.py
```diff
@@ -168,6 +194,7 @@ def compute_training_loss(
             TrainingLosses: An instance of ``TrainingLosses`` containing backward loss and additional losses indexed
             by name. Additional losses includes penalty loss.
         """
+        log(DEBUG, "COMPUTE TRAINING LOSS CALL FROM ADAPTIVE MIXIN")
```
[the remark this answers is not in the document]
emersodb: No worries. I'll take your advice and wait a touch to do anymore reviewing as you do your debugging etc. 🙂

fl4health/clients/nnunet_client.py
nerdai: done!

nerdai: this is unchanged now.

fl4health/clients/nnunet_client.py
nerdai: "global" is retained.

fl4health/clients/nnunet_client.py
nerdai: outdated and no longer existing in latest iteration, resolving as a result.

fl4health/mixins/personalized/ditto.py
nerdai: resolving as this cleaner logic has been implemented.

fl4health/mixins/adaptive_drift_constrained.py
```diff
@@ -29,6 +29,13 @@ class AdaptiveDriftConstrainedProtocol(BasicClientProtocol, Protocol):
 
     def compute_penalty_loss(self) -> torch.Tensor: ...  # noqa: E704
 
+    def __compute_training_loss(  # noqa: E704
```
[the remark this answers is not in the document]
nerdai: Yeah, you'll probably see it more often now ha — it's def a Pythonic thing to do and one that confused the hell out of me for a while as to when to use "_" vs "__".

fl4health/clients/basic_client.py
```diff
@@ -975,34 +986,19 @@ def get_parameter_exchanger(self, config: Config) -> ParameterExchanger:
...
-                it is assumed that the keys of input match the names of the keyword arguments of
-                ``self.model.forward().``
-
-        Returns:
-            tuple[TorchPredType, TorchFeatureType]: A tuple in which the first element contains a dictionary of
-            predictions indexed by name and the second element contains intermediate activations indexed by name. By
-            passing features, we can compute losses such as the contrastive loss in MOON. All predictions included in
-            dictionary will by default be used to compute metrics separately.
+    def _predict(self, model: torch.nn.Module, input: TorchInputType) -> tuple[TorchPredType, TorchFeatureType]:
+        """Helper predict method.
```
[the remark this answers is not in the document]
nerdai: Resolving for now to clean things up and since this has been done in latest iteration. Feel free to open a new one should the rename and new doc string need further edits.

fl4health/mixins/adaptive_drift_constrained.py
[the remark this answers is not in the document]
nerdai: Resolving for now, to clean things up. I have removed this debug statement as well.

fl4health/mixins/personalized/ditto.py
```diff
@@ -369,73 +372,50 @@ def train_step(
...
-                # Note that this assumes the keys of the input match (exactly) the keyword args
-                # of the forward method.
-                global_preds = self.safe_global_model()(**input)
-                local_preds = self.model(**input)
+        training_losses = TrainingLosses(backward=local_loss + penalty_loss, additional_losses=additional_losses)
 
+        # aggregate preds
         if isinstance(global_preds, torch.Tensor) and isinstance(local_preds, torch.Tensor):
-            return {"global": global_preds, "local": local_preds}, {}
+            agg_preds = {"global": global_preds, "local": local_preds}
```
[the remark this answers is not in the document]
nerdai: done! Resolving to clean up the PR.

examples/nnunet_pfl_example/config.yaml
```diff
@@ -0,0 +1,8 @@
+# Parameters that describe the server
+n_server_rounds: 1
+
+# Parameters that describe the clients
+n_clients: 1
```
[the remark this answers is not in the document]
nerdai: Yeah, this just a mirror copy of our existing nnunet example, but we can switch to 2 clients for both.

examples/nnunet_pfl_example/client.py
```diff
@@ -0,0 +1,245 @@
...
+    dataset_path: Path,
+    msd_dataset_name: str,
+    server_address: str,
+    fold: int | str,
+    always_preprocess: bool = False,
+    verbose: bool = True,
+    compile: bool = True,
+    intermediate_client_state_dir: str | None = None,
+    client_name: str | None = None,
+    personalized_strategy: Literal["ditto"] | None = None,
```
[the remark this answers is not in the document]
nerdai: Yeah, I haven't used pure enum with argparser before. So this was a quick and easy way to do this, which I felt was okay given this is just an example anyway versus being in the library code itself. But can look to see how to do this for enum if we really want to do it here.

emersodb (commented):
@nerdai: I've only gotten part of the way through the PR, but I think we might want to talk a bit about the design. I'm a bit worried that the functionality with all the private functions is getting a bit fractious and knotted. I just want to make sure I understand the ideas a bit better before we commit to this.

fl4health/clients/basic_client.py
```diff
@@ -559,38 +559,103 @@ def update_metric_manager(
...
+            a dictionary of any predictions produced by the model prior to the
+            application of the backwards phase
         """
         # Clear gradients from optimizer if they exist
-        self.optimizers["global"].zero_grad()
+        optimizer.zero_grad()
 
         # Call user defined methods to get predictions and compute loss
-        preds, features = self.predict(input)
+        preds, features = self._predict_with_model(model, input)
```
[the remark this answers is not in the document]
nerdai: Yeah this is for added flexibility, which I think is actually a nice (standard) way to handle what we needed to do.

1. The private methods are internal and more flexible now with the injection of model dependency. This is flexible in the sense should we need to change any of these private methods, then we can and can do so in a way that doesn't break the public API, namely `predict()`. Its not so much circumventing from the user perspective, but giving us library developers more flexibility.

2. We needed to go this route to be able to essentially re-use the already established logic contained in our existing clients when apply ditto and future mixins. With this internal method, we (as library devs) can make use of our private methods and supply local/global models. Without this, we can't do this and we have to resort to the alternative which is DittoClient and then subclasses of DittoClient and the other client, which I think leads to a lot of repeat code and class explosion.

3. What should eventually happen is that we move all existing clients to implement their private methods. There is no rush to do this as this isn't a breaking change, but we should move it at some point so its clear to devs in the future on how to create clients—namely, if the private method exists, implement those helpers.

Happy to discuss more on it, but I think this is okay and from my experience a pretty typical thing to do.

PR #394, opened 2025-06-03
State checkpointer refactor

fatemetkl:
# PR Type
Fix

# Short Description

Clickup Ticket(s): https://app.clickup.com/t/868d4w5rt

Unifying our PerRoundStateCheckpointer and the Snapshotting functionality that supports early stopping.
This PR adds `checkpointing.state_checkpointer` modules with classes that can be used to checkpoint the state of the server and clients. 

# Tests Added
* `tests/checkpointing/test_state_checkpointer.py`
* `tests/utils/test_early_stopper.py`

I also manually tested our `nnunet_example` to make sure `NnUnetServerCheckpointAndStateModule` works as expected. We have a smoke test (load_from_checkpoint_example) for `ServerPerRoundStateCheckpointer`.

### DIFF
--- a/tests/utils/snapshotter_test.py
+++ b/tests/utils/snapshotter_test.py
@@ -35,10 +35,9 @@ def compare_mixed_dictionaries(
             if dict1_value != narrow_dict_type(dict2, key, list):
                 return False
         elif isinstance(dict1_value, dict):
-            for k, v in dict1_value.items():
-                if not compare_mixed_dictionaries(v, narrow_dict_type(dict2, k, dict)):
-                    return False
-                return True
+            if not compare_mixed_dictionaries(dict1_value, narrow_dict_type(dict2, key, dict)):
+                return False
+            return True
         else:
             raise TypeError(f"Unsupported type in dictionary: {type(dict1_value)}")
 


tests/smoke_tests/run_smoke_test.py
```diff
@@ -529,8 +529,7 @@ async def run_fault_tolerance_smoke_test(
 
         # client assertions
         client_errors = []
-        for i in range(len(client_processes)):
-            client_errors.extend(_assert_metrics(MetricType.CLIENT, client_metrics, tolerance))
+        client_errors.extend(_assert_metrics(MetricType.CLIENT, client_metrics, tolerance))
```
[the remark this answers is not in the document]
emersodb: Even though we're comparing against the same set of metrics for each client (since they're operating on the same dataset and off the same random seed), I think we still want to do this assertion for each of them to make sure they are all functioning as expected? That is, if, for some reason one of the clients has something go wrong for some reason, we'll detect it.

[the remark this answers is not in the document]
fatemetkl: I might be wrong! based on what I understood ` _assert_metrics` goes over all the clients (`for file in DEFAULT_METRICS_FOLDER.iterdir()`), and for each client type metric (`if metrics["host_type"] == metric_type.value`) performs ` errors.extend(_assert_metrics_dict(metrics_to_assert, metrics, tolerance))`. Since we are already checking for all the clients, I removed `for i in range(len(client_processes)):`
In my test case, only one of the clients (out of two) had issues, but the size of client_errors was twice what it should have been (the error message was repeated twice). After I made this change, the size of the error messages was correct.
Overall, I think this is minor, and I can revert it if we are not sure.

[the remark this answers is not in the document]
emersodb: Oh I see. You're saying we don't need to loop because we already run through elsewhere. I'll look a little deeper to confirm when I do a full review. I was thinking you were suggesting we just check one clients logs against the metrics since they will all be the same, which would be fine most of the time, but not 100% of the time 🙂.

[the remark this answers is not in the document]
fatemetkl: Yes! We are repeating the loop over all the clients 2 times. Thanks! 🙂

fl4health/checkpointing/state_checkpointer.py
```diff
@@ -0,0 +1,594 @@
...
+        # Setting default name if one doesn't exist. If we're here and it doesn't exist yet, user is expecting
+        # a default name for loading anyway.
+        self.maybe_set_default_checkpoint_name()
+        if self.checkpoint_exists():
+            self.load_state(attributes)
+            log(INFO, f"State checkpoint successfully loaded from: {self.checkpoint_path}")
+            self.client = None
+            return True
+
+        # Clear the client after we are done updating its attributes.
```
emersodb, line 383: Maybe update this comment, since this only happens if the checkpoint didn't exist? Something like "Clear the client since no checkpoint exists."

fl4health/checkpointing/state_checkpointer.py
```diff
@@ -0,0 +1,594 @@
...
+                "train_metric_manager": (
+                    SerializableObjectSnapshotter(),
+                    MetricManager,
+                ),
+            }
+
+        super().__init__(checkpoint_dir, checkpoint_name, snapshot_attrs)
+        self.client: BasicClient | None = None
+
+    def maybe_set_default_checkpoint_name(self, object_name: str | None = None) -> None:
```
emersodb, line 328: I don't think we need this `object_name` arg anymore with our redesign. This function can be simplified to 
```
    def maybe_set_default_checkpoint_name(self) -> None:
        """
        Potentially sets a default name for the checkpoint to be saved. If ``checkpoint_dir`` is set but
        ``checkpoint_name`` is None then a default ``checkpoint_name`` based on the underlying name of the client to
        be checkpointed will be set of the form ``f"client_{self.client.client_name}_state.pt"``.
        """
        assert self.client is not None, "Attempting to save client state but client is None"
        # Set the checkpoint name based on client's name if not already provided.
        if self.checkpoint_name is None:
            # If checkpoint_name is not provided, we set it based on the client name.
            self.checkpoint_name = f"client_{self.client.client_name}_state.pt"
            self.set_checkpoint_path(self.checkpoint_dir, self.checkpoint_name)
```

fl4health/checkpointing/state_checkpointer.py
```diff
@@ -0,0 +1,594 @@
...
+                    ReportsManager,
+                ),
+                "server_name": (StringSnapshotter(), str),
+                "history": (HistorySnapshotter(), History),
+            }
+        super().__init__(checkpoint_dir, checkpoint_name, snapshot_attrs)
+        self.server: FlServer | None = None
+        self.server_model: nn.Module | None = None
+
+    def maybe_set_default_checkpoint_name(self, object_name: str | None = None) -> None:
```
emersodb, line 456: With the refactor, I don't think we need `object_name` anymore? This can be simplified to
```
    def maybe_set_default_checkpoint_name(self) -> None:
        """
        Potentially sets a default name for the checkpoint to be saved. If ``checkpoint_dir`` is set but
        ``checkpoint_name`` is None then a default ``checkpoint_name`` based on the underlying name of the server to
        be checkpointed will be set of the form ``f"server_{self.server.server_name}_state.pt"``.
        """
        assert self.server is not None, "Attempting to save server state but server is None"
        # Set the checkpoint name based on server's name if not already provided.
        if self.checkpoint_name is None:
            # If checkpoint_name is not provided, we set it based on the server's name.
            self.checkpoint_name = f"server_{self.server.server_name}_state.pt"
            self.set_checkpoint_path(self.checkpoint_dir, self.checkpoint_name)
```

fl4health/checkpointing/state_checkpointer.py
```diff
@@ -0,0 +1,594 @@
...
+        if self.checkpoint_exists():
+            self.load_state(attributes)
+            log(INFO, f"State checkpoint successfully loaded from: {self.checkpoint_path}")
+            # Clear the server after we are done updating its attributes.
+            self.server = None
+            # Server model is saved and returned separately for parameter extraction.
+            return self.server_model
+
+        log(INFO, f"No state checkpoint found at: {self.checkpoint_path}")
+        # Clear the server objects after checkpointing.
```
emersodb, line 522: Maybe we can update this comment to "Clear the server objects if checkpoint is not found"?

fl4health/utils/early_stopper.py
```diff
@@ -61,94 +40,25 @@ def __init__(
...
-                WARNING,
-                "Checkpointing directory is not provided. Client best state will be kept in the memory.",
-            )
-            self.snapshot_ckpt = copy.deepcopy(self.snapshot_ckpt)
+        self.best_score: float | None = None
 
     def load_snapshot(self, attributes: list[str] | None = None) -> None:
         """
-        Load checkpointed snapshot dict consisting to the respective model attributes.
+        Load the best snapshot of the client state from the checkpoint directory or memory.
```
emersodb, line 54: We don't need the "or memory" part of this docstring anymore I don't think?

fl4health/utils/snapshotter.py
```diff
@@ -220,8 +159,8 @@ def load_attribute(
             attribute[key] = attribute_snapshot[key]
 
 
-class NumberSnapshotter(AbstractSnapshotter[int | float]):
-    def save_attribute(self, attribute: dict[str, int | float]) -> dict[str, Any]:
+class NumberSnapshotter(AbstractSnapshotter[int | float | bool]):
```
emersodb, line 162: If we're going to allow "non-numbers" i.e. booleans, perhaps we should rename this to something like "SingletonSnapshotter" or something else?

emersodb: This would also mean updating the mention of numbers in the docstrings below 🙂

tests/utils/snapshotter_test.py
```diff
@@ -1,183 +0,0 @@
-import copy
```
emersodb, line 1: I might be wrong, but it feels like there may be a few tests in here that we can keep around to test out the more sophisticated snapshotters like `LRSchedulerSnapshotter` etc? I know you made a fair few changes to the classes, but covering those would be useful I think?

emersodb: Some of them are now covered in the state checkpointer tests. So maybe we just add `LRSchedulerSnapshotter` and any others that are losing some coverage to there?

emersodb (commented):
Nearly ready to go I think a couple of small questions and suggested refactors, but looks great!

examples/nnunet_example/client.py
```diff
@@ -140,7 +140,7 @@ def main(
             get_msd_dataset_enum function",
     )
     parser.add_argument(
-        "--always-preprocess",
```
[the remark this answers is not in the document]
fatemetkl: I don't remember doing these changes myself. Perhaps these are from pre-commit checks.
In this file, the previous argument (`--msd_dataset_name`) also has underscores. In nnunit_example.server, `--config_path` also has an underscore, which is a bit inconsistent to the rest of the args, but I can revert all my changes here.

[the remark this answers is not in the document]
emersodb: If they are pre-commit checks, let me know. We can leave them in. Just figured we might as well just leave them be 😂

fl4health/utils/early_stopper.py
fatemetkl: Nice catch!

fl4health/checkpointing/state_checkpointer.py
```diff
@@ -0,0 +1,649 @@
...
+
+    def load_state(self, attributes: list[str] | None = None) -> None:
+        """
+        Load checkpointed state dictionary either from the checkpoint or from the memory (self.snapshot_attrs).
+
+        Args:
+            attributes (list[str] | None): List of attributes to load from the checkpoint. If None, all attributes
+                specified in ``snapshot_attrs`` are loaded. Defaults to None.
+        """
+        assert self.checkpoint_exists() or self.snapshot_ckpt != {}, "No state checkpoint to load."
```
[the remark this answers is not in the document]
fatemetkl: This is also addressed in your PR.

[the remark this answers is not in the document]
emersodb: Yeah sorry. These are legacy comments. Not sure why they were dragged along.

examples/nnunet_example/client.py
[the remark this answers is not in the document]
fatemetkl: I shouldn't have blamed the pre-commit checks, my memory should be blamed 😂.

emersodb (approved):
Good to go!

PR #396, opened 2025-06-05
Creating FedBN client and adding test

emersodb:
# PR Type
Feature

# Short Description

Really small PR to add in a sanity check layer for FedBN clients. This allows us to ensure that someone wanting to use BN has a bit of a check on their definition of the parameter exchanger. FedBN really only differs from the Basic Client in a single way (uses a layer exchanger with exclusions), but we weren't really enforcing that anywhere and the example was just "doing so properly" so to speak. This PR just adds in a layer where, if the user wants to use a FedBN client, it checks that they have an exchanger of the correct type.

# Tests Added

Test that the enforcement fails if the user does not use the right class for parameter exchange.


fatemetkl (approved):
LGTM! :rocket:

PR #405, opened 2025-06-20
Loosening Some Ruff Ignores

emersodb:
# PR Type
Other

# Short Description

This PR removes some of the Ruff ignores that we have in place and fixes the issues that arise in our code base by doing so. Namely, applying the Ruff rules
- D107: No undocumented public functions
- D417: No undocumented parameters in the args section of docstrings
- D105: No undocumented public init functions
- PLR2004: No magic numbers

**Note**: We do not enforce various ruff rules in our examples, tests, and research folders. Below is a justification for the settings I chose in these folders.

```
# Ignoring undocumented public functions, public init, magic method, and magic numbers
"tests/*" = ["D103", "D105", "D107", "PLR2004"]
```
In the tests, there is good justification for having magic numbers. We also create a lot of helper functions, classes, and magic methods meant to facilitate testing that would be very annoying (and many would be unnecessary) to document formally.

```
# Ignoring undocumented, public init, magic method, and magic numbers
"research/*" = ["PLR2004", "D105", "D107"]
```
Our research folders are meant to be less formal. It's nice to have documentation where helpful to someone using the code, but things like public init documentation, magic method descriptions, and magic numbers aren't exceptionally important.

```
# Ignoring undocumented public init
"examples/*" = ["D107"]
```
Each time we want to perform FL training, we override the server and client base classes to produce customized training flows. Documenting these is unnecessary and we have README files that discuss the point of each example and how to run it.

# Tests Added

NA

### DIFF
--- a/fl4health/clients/nnunet_client.py
+++ b/fl4health/clients/nnunet_client.py
@@ -138,6 +138,8 @@ class NnunetClient(BasicClient):
                 No checkpointing (state or model) is done if not provided. Defaults to None.
             reporters (Sequence[BaseReporter], optional): A sequence of FL4Health reporters which the client should
                 send data to.
+            client_name (str | None, optional): Name of the client. If None the class will set a default and random
+                name. Defaults to None.
             nnunet_trainer_class (type[nnUNetTrainer]): A ``nnUNetTrainer`` constructor. Useful for passing custom
                 ``nnUNetTrainer``. Defaults to the standard nnUNetTrainer class. Must match the
                 ``nnunet_trainer_class`` passed to the ``NnunetServer``.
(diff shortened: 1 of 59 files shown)


fl4health/clients/nnunet_client.py
```diff
@@ -904,7 +912,7 @@ def update_before_train(self, current_server_round: int) -> None:
         gc.collect()  # Cleans up unused variables
         # As the linked issue above points out, calling gc.freeze() greatly reduces the
         # overhead of garbage collection. (from 1.5s to 0.005s)
-        if current_server_round == 2:
+        if current_server_round == 2:  # noqa: PLR2004
```
emersodb, line 915: This magic number is sort of hard to describe in a concise name... So applying an ignore

nerdai (approved):
Justifications seem sensible to me.

fatemetkl (approved):
LGTM!

PR #406, opened 2025-06-21
[chore] Add unit tests for protocols

nerdai:
# PR Type
Other

# Short Description

As a quick follow-up to the `FlexibleClient` PR, this PR introduces tests for our protocols.

Adds unit tests for protocols to assert methods and annotations exist.

### DIFF
--- a/tests/mixins/test_core_protocols.py
+++ b/tests/mixins/test_core_protocols.py
@@ -1,0 +1,53 @@
+from fl4health.mixins.core_protocols import (
+    FlexibleClientProtocol,
+    FlexibleClientProtocolPreSetup,
+    NumPyClientMinimalProtocol,
+)
+
+
+def test_numpy_client_protocol_attr() -> None:
+    """Test interface for NumPyClientMinimalProtocol."""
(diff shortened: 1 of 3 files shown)


tests/mixins/test_core_protocols.py
```diff
@@ -0,0 +1,53 @@
...
+    assert hasattr(NumPyClientMinimalProtocol, "fit")
+    assert hasattr(NumPyClientMinimalProtocol, "evaluate")
+    assert hasattr(NumPyClientMinimalProtocol, "set_parameters")
+    assert hasattr(NumPyClientMinimalProtocol, "update_after_train")
+
+
+def test_flexible_client_presetup_protocol_attr() -> None:
+    """Test interface for FlexibleClientProtocolPreSetup."""
+    annotations = FlexibleClientProtocolPreSetup.__annotations__
+    assert "device" in annotations
```
nerdai, line 20: Apparently class vars are not actual class attributes in `Protocols` so this is how we can test existence of these attributes.

emersodb (approved):

PR #412, opened 2025-06-26
[chore] Parallelize smoke tests in Github runners

nerdai:
# PR Type

Other

# Short Description
Let's speed up our dev cycles by parallelizing smoke tests... our biggest bottleneck for checks.

image

## Add

- pytest-test-groups dev dependency

## Changed

- github workflow for smoketests to split the smoketests into 4 groups and run them in parallel.


.github/workflows/smoke_tests.yaml
```diff
@@ -44,12 +48,13 @@ jobs:
 
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --with "dev, test"
+        run: |
```
emersodb, line 51: Is this pipe just for aesthetics? Just curious.

.github/workflows/smoke_tests.yaml
```diff
@@ -7,13 +7,17 @@ on:
...
 
 jobs:
   test:
     runs-on: ubuntu-22.04
     permissions:
       contents: read
       actions: write
+    strategy:
+      matrix:
+        group: [1, 2, 3, 4]
```
emersodb, line 20: I'll look this up a bit more later, but curious, is this literally just split the collected tests into 4 groups numbered 1,2,3,4? I assume that the unit tests don't get put into one of these groups and/or broken up is because you don't "assign" them to a group with the argument `--test-group`?

.github/workflows/smoke_tests.yaml
```diff
@@ -44,12 +48,13 @@ jobs:
...
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --with "dev, test"
+        run: |
+          poetry install --with "dev, test"
 
       - name: Run Script
         run: |
           source .venv/bin/activate
-          pytest -m "smoketest" -v --cov fl4health --cov-report=xml tests
+          pytest --test-group-count=4 --test-group=${{ matrix.group }} -v --cov fl4health --cov-report=xml tests/smoke_tests/test_smoke_tests.py
```
emersodb, line 57: A naive question, but when passing `--test-group ${{ matrix.group }` does this mean that it will just randomly send each individual smoke to one of the members of the matrix.group list?

emersodb (approved):
🔥

scarere:
Quick question, is there any way to group the smoketests together in the UI so that the list of workflows is more organized?

emersodb:
@nerdai: It seems like I might have approves this prematurely. I thought all tests had passed, but it seems like maybe they hadn't (13/17 checks passed) and we were still able to merge? I think maybe we need to investigate this a touch, because maybe the merge blocking is broken with this PR?

Also to Shawn's point, is there a way we can get better looking output? You screen shot looks nice, but then the checks UI looks a little wonky above.

nerdai:
> Quick question, is there any way to group the smoketests together in the UI so that the list of workflows is more organized?

I think at the very least we can do that manually and create separate "marks" for these groups, and just name the jobs accordingly. I'm not sure `pytest-test-groups` would allow anything fancier than just basic splitting. If we think its worth it, we should add a ticket to the board and prioritize the work tho :) 

I definitely thought it was becoming a priority to get the runtime of these smoketests down

.github/workflows/smoke_tests.yaml
nerdai: Yeah, just putting it here as well for convenience, but the groups are assigned deterministically via something like this:

```
num_jobs = 26
num_groups=4
group_1 = range(0,26,num_groups)
group_2 = range(1,26,num_groups)
group_3 = range(2,26,num_groups)
group_4 = range(3,26,num_groups)
```

The unit tests don't use the group params in their command so they run as they did before:

```
# command for unit tests
          pytest -m "not smoketest" -v --cov fl4health --cov-report=xml tests
```

nerdai: From the unit tests logs:

```
collected 482 items / 26 deselected / 456 selected
```

(26 for the tests marked as smoketests were deselected)

.github/workflows/smoke_tests.yaml
nerdai: oops lol -- so at first I was trying to install the `pytest-test-groups` dep here with `poetry add ...`, thus I had two commands.

I ended up just adding the dep to our dev group, but didn't revert back to the original syntax for this.

nerdai:
Ah I see what you mean. I'm not sure if we can modify the way the UI portrays workflow jobs that are run via a "matrix" strategy -- never looked. If you're interested in investigating it, then by all means :)

PR #413, opened 2025-06-26
Update nnunet example with efficient metrics

scarere:
# PR Type
[**Feature** | Fix | **Documentation** | Other ]

# Short Description

Set out to basically update the nnunet example with the efficient metrics and the compound EMA metric. This way they get coverage in the smoke tests and we have a good example on how to use them. Plus if you're training a nnunet model, you probably want to use the efficient metrics anyways given you'll probs OOM without them.

I also went ahead and tried to make some of the documentation for the efficient metrics a bit clearer. I have some ideas for refactors as well as some general python questions that I'm going to leave in the comments.

## Summary of Changes

- Updated `nnunet_example` to use `EmaMetric` and `efficient_metrics.BinaryDice` or `efficient_metrics.MultiClassDice`
- Updated documentation for efficient metrics
- Changed `MetricOutcome` Enum to `ClassificationOutcome`  
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


fl4health/metrics/efficient_metrics_base.py
```diff
@@ -17,7 +17,7 @@
 BINARY_COUNT_TENSOR_N_DIMS = 2
 
 
-class MetricOutcome(Enum):
+class ClassificationOutcome(Enum):
```
scarere, line 20: q: I was looking at some other python code and discovered the use of Python Literal types for cases like this. I know we tend to favour Enums in this library and I'm wondering what the tradeoffs/advantages of each are and what reason we would have to choose one over the other. At a quick glance, it seems using Literals has the advantage of 1. The user doesn't need to import the enum class, they can just pass a string 2. The allowed values of the literal appear in the docstring implicitly with the typing, whereas for enums the user needs to find the enum documentation to see what its members are

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -224,7 +228,7 @@ class have implemented `count_tp_fp_tn_fn` such that, if `self.batch_dim` is not
             return
 
         # If batch_dim has been specified, we accumulate counts for EACH INSTANCE seen throughout the updates, such
-        # that each of the counts is a 2D tensor of row length equal to the samples seen.
+        # that each of the counts is a tensor of row length equal to the samples seen.
```
scarere, line 231: Counts are not necessarily 2D, it's only 2D if `label_dim` was also specified. On that note, this class assumes that the inheriting classes have overrided `count_tp_fp_tn_fn` and ensured that batch dim always appears first. If this base class must assume that, then why not just have that functionality inside the base class? Then we can get rid of the `transpose_2d_matrix` method in the `MultiClassClassificationMetric` class.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -380,34 +397,39 @@ def __init__(
...
+                negative class label are discarded. Defaults to None.
+            batch_dim (int | None, optional): If None, the counts along the specified dimension (ie. for each sample)
+                are aggregated and the batch dimension is reduced. If specified, counts will be computed along the
+                dimension specified. That is, counts are maintained for each training sample INDIVIDUALLY. For example,
+                if batch_dim = 1 and label_dim = 0, then
 
                 .. code-block:: python
 
-                    p = torch.tensor([[[0, 0, 0, 1], [1, 1, 1, 1]]])
+                    p = torch.tensor([[[0, 0, 0, 1], [1, 1, 1, 1]]])  # Size([1, 2, 4])
```
scarere, line 422: It made sense to me for the example to have different sizes for the label and batch dims so that it was clear that they were transposed

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -380,34 +397,39 @@ def __init__(
         ("soft"). For example, with a target of 1, a prediction of 0.8 contributes 0.8 to the true_positives count and
         0.2 to the false_negatives.
 
-        NOTE: For this class, the predictions and targets passed to the update function MUST have the same shape
+        NOTE: For this class, the predictions and targets passed to the `update` function MUST have the same shape
 
         NOTE: Preds and targets are expected to have elements in the interval [0, 1] or to be thresholded, using
         that argument to be as such.
 
+        NOTE: For this class, only the counts for the positive label are accumulated.
```
scarere, line 405: This is a comment for the `_postprocess_tensor` method. Why are we enforcing an empty dim? Why not just return tensor of shape (batch_size) if `batch_dim` is specified and (1,) or even (,) otherwise. Whats the point of having the extra dim and returning tensors of shape (batch_size, 1).

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -656,28 +679,28 @@ def __init__(
         Args:
             name (str): The name of the metric.
             label_dim (int): Specifies which dimension in the provided tensors corresponds to the label
-                dimension. During metric computation, this dimension must have size of AT LEAST 2.
-            batch_dim (int | None, optional): If None, then counts are aggregated across the batch dimension. If
-                specified, counts will be computed along the dimension specified. That is, counts are maintained for
-                each training sample INDIVIDUALLY. For example, if batch_dim = 1 and label_dim = 0, then
+                dimension. During metric computation, this dimension must have size of AT LEAST 2. Counts are always
```
scarere, line 682: Why enforce that the label dim is not reduced? Why not just have the same functionality as the base class we reduce it automatically if its not specified?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -656,28 +679,28 @@ def __init__(
...
+                dimension specified. That is, counts are maintained for each training sample INDIVIDUALLY.
+                NOTE: If `batch_dim` is specified, then counts will be presented batch dimension
+                first, then label dimension. For example, if batch_dim = 1 and label_dim = 0, then
 
                 .. code-block:: python
 
-                    p = torch.tensor(
-                        [[[1.0, 1.0, 1.0, 0.0], [0.0, 0.0, 0.0, 0.0]], [[0.0, 0.0, 0.0, 1.0], [1.0, 1.0, 1.0, 1.0]]]
-                    )
+                    p = torch.tensor([[[1.0, 1.0, 1.0, 0.0]], [[0.0, 0.0, 0.0, 1.0]]])  # Size([2, 1, 4])
```
scarere, line 693: I found it tough to follow this examples given the number of dimensions. Size annotations help make it easier to understand, although i feel like it would be clearer of batch_dim and label_dim had different sizes. The only difficult part is ideally you want a minimum size of 2, which would make the smalles possible tensor for this example something with a shape like (2, 3, 4), which is big. So i left it for now

scarere:
Unit tests seem to be failing due to the image not having the pytest executable. Does anyone know anything about this?

emersodb (approved):
Overall, I think this looks good. I responded to your questions and comments as best I could remember. Let me know if they were helpful at all.

Have you run an example to make sure everything is working as expected and the EMA seems like its making sense?

scarere:
> Overall, I think this looks good. I responded to your questions and comments as best I could remember. Let me know if they were helpful at all.
> 
> Have you run an example to make sure everything is working as expected and the EMA seems like its making sense?

Yep I ran the example and numbers check out.

fl4health/metrics/efficient_metrics_base.py
scarere: Right but in both cases, after running `super().count_tp_fp_tn_fn()`, the the returned tensors are 2D, then there's a label and batch dimension, and we can use that info to transpose. For binary case with impicit binary predictions, they are only implicit if `label_dim` was left unspecified, in which case we would have only a 1D tensor and there is nothing to transpose. I might just see if i can quickly add this in and get it working.

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -327,11 +336,19 @@ def compute_from_counts(
...
+        Provided tensors associated with the various classification outcomes from predictions compared to targets in
+        the form of true positives, false positives, true negatives, and false negatives, returns a dictionary of
+        Scalar metrics. For example, one might compute recall as true_positives/(true_positives + false_negatives).
+
+        The count tensors will all have the same shape. This shape depends on whether `batch_dim` and or `label_dim`
+        were specified. For example, if the batch and label dimensions have sizes 2 and 3 respectively, then:
+                    - Both `batch_dim` and `label_dim` are specified: Size([2, 3])
+                    - Only `batch_dim` is specified: Size([2])
+                    - Only `label_dim` is specified: Size([3])
+                    - Neither specified: Size([])
```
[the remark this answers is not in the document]
scarere: Same comment as before. Note that in the documentation I add a note that inheriting classes may further modify the shapes of the count tensors that are provided as arguments. I guess it is sort of on the inherting class to update the documentation, which I believe they do

emersodb:
> I think we could refactor even more by making the ClassificationMetric base class the one to use for multiclass settings, and then having `BinaryClassificationMetric` be a subclass, but I might propose that in a separate PR.

Yes, let's leave that to a future PR

nerdai (approved):
Just a quick question if we want to support `MetricOutcome` for backwards compat. This PR presents a breaking change — if we had a ton of users, I'd definitely hold the brakes until we've addressed this issue, but I think we're less concerned here?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -11,13 +11,11 @@
...
-N_DIM_2D_TENSOR = 2
-EXPECTED_BINARY_LABEL_DIM_SIZE = 2
-BINARY_COUNT_TENSOR_MAX_SIZE = 2
-BINARY_COUNT_TENSOR_N_DIMS = 2
+MAX_COUNT_TENSOR_DIMS = 2  # Count tensors are never more than 2 dimensional
+N_LABELS_BINARY = 2  # The number of labels in binary segmentation.
 
 
-class MetricOutcome(Enum):
+class ClassificationOutcome(Enum):
```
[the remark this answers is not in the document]
scarere: However seeing the efficient metrics were a relatively recent addition, I don't think MetricOutcome would be in use by anyone so it should be safe just to deprecate it. But again happy to put an alias in to follow good practices

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -201,14 +203,40 @@ def _sum_along_axes_or_discard(
...
 
+    def _maybe_transpose_count_tensor(self, count_tensor: torch.Tensor) -> torch.Tensor:
+        """
+        Ensures that the batch dimension always appears first in the count tensor.
+
+        If the count tensor does not have two dimensions, or the batch dimension is already in front of the label
+        dimension, then it is returned unchanged. Otherwise it is transposed.
+
+        Args:
+            count_tensor (torch.Tensor): The count tensor after reducing unused dimensions
```
emersodb, line 214: Is "unused" the right term here? Maybe just "reducing dimensions"?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -201,14 +203,40 @@ def _sum_along_axes_or_discard(
...
+
+        Returns:
+            torch.Tensor: If count_tensor was 1D then it is returned unchanged. Otherwise this is the count_tensor
```
emersodb, line 217: perhaps wrap count_tensor in `` to make it pretty in the sphinx docs render?

emersodb: Same with batch_size and num_labels?

fl4health/metrics/efficient_metrics_base.py
```diff
@@ -201,14 +203,40 @@ def _sum_along_axes_or_discard(
...
     def update(self, preds: torch.Tensor, targets: torch.Tensor) -> None:
         """
         Updates the existing self.true_positive, self.false_positive, self.false_negative and self.true_negative
         counts with new counts computed from preds and targets.
 
-        NOTE: In the implementation, this function implicitly assumes that the child classes inheriting from this
-        class have implemented `count_tp_fp_tn_fn` such that, if `self.batch_dim` is not None, the counts are returned
-        with shapes such that the batch dimension comes FIRST for the counts.
+        NOTE: This function assumes that if `self.batch_dim` is not None, the counts are returned with shapes such that
+        the batch dimension comes FIRST for the counts. If `self.count_tp_fp_tn_fn` is overriden it must ensure that
```
emersodb, line 238: overridden I think?

emersodb (approved):
I really like the simplifications you came up with 🙂. Just a few cosmetic suggestions. Other than that, good to go.

fl4health/metrics/efficient_metrics_base.py
scarere: Yeah i need to learn what the sphinx formatting shortcuts are

PR #420, opened 2025-07-07
Some changes suggested to parameter initialization to better support client-side calls

emersodb:
# PR Type
Feature/Fix

# Short Description

Our support for client-side initialization for more complex parameter exchange paradigms is a bit weak due to our primary focus on server-side initialization. This PR helps strengthen that blind spot in order to better support workflows that require client-side initialization in many cases (nnunet is a principle example).

# Tests Added

Added a small test to make sure a function added works as expected.


nerdai (approved):
This looks good! Just one quick question: how come all of the clients don't need to use the new helper, `setup_client_and_return_all_model_parameters`?

emersodb:
> This looks good! Just one quick question: how come all of the clients don't need to use the new helper, `setup_client_and_return_all_model_parameters`?

That's a good question. There are a few reason that this doesn't propagate to all of the clients (but does to most of them). 
- Some clients don't actually use the basic client workflow (FedPcaClient). So it can't be inherited there. 
- Some clients have a slightly different workflow where it doesn't make sense to exchange the full model and have custom logic (FendaDittoClient)
- Some clients don't actually initialize a model at all (ModelMergeClient).

So there are some edge cases. However, if there is a client you see that should definitely use it. Let me know 🙂

PR #423, opened 2025-07-12
Some Docs Cleanup, Spelling Corrections, Dependency simplification.

emersodb:
# PR Type
Other

# Short Description

- Taking a full pass through our docs and making some rendering improvements.
- Correcting some minor spelling mistakes.
- Simplifying our dependencies in our toml, as they were a little unnecessarily complicated.

**NOTE**: For reviewing this PR, I'd suggest just taking my word for it on the documentation and spelling changes (which are the vast majority of these changes). They are all cosmetic. 

As such, you can filter out all of the python files in this PR. This will leave you with a much smaller set of changes to review. The only changes of real significant that I've made is in the toml to reduce the number of groups we have. There is still some thoughtful arrangement that needs to happen here but that will wait until later.

# Tests Added

NA


README.md
fatemetkl, line 1: Would you mind also fixing the GPFL documentation here? It is not properly placed inside the table 😢.

Everything else looks good to me! 🎉

fatemetkl (approved):

PR #429, opened 2025-07-21
[Feature] Adding MR-MTL Personalized Mixin

nerdai:
# PR Type
Feature

# Short Description

## Intro

Now that we have established the mixin/adapter pattern wtih Ditto, in this PR we adopt the same pattern to implement `MrMtlPersonalizedMixin` that works with `FlexibleClient` types.

## Summary

- Adds `MrMtlPersonalizedMixin` and associated `MrMtlPersonalizedProtocol` for typing
- Updates `make_it_personal` class factory method to include `MrMtlPersonalizedMixin`
- Adds `client_dynamic.py`  to `examples/mr_mtl_example` and confirm that it produces exact same result as `client.py` when setting seed — as we did with `examples/ditto_example`

## Testing of dynamically personalized and subclass `MrMtlClient`

image

- left (dynamic) and right (original client inheriting from MrMtlClient)
- examples/mr_mtl_example produces exact same results when setting seed

# Tests Added

- smoke test: `tests/smoke_tests/tests_smoke_tests.py::test_mr_mtl_flexible_mnist`
- unit tests: `tests/mixins/personalized/test_mr_mtl.py`


fl4health/mixins/personalized/mr_mtl.py
```diff
@@ -0,0 +1,212 @@
+"""MR MTL Personalized Mixin."""
+
+import copy
+import warnings
+from logging import INFO, WARN
```
emersodb, line 5: We've been using the full WARNING constant just to be consistent. I honestly don't know why there are two...

fl4health/mixins/personalized/mr_mtl.py
```diff
@@ -0,0 +1,212 @@
...
+        # Initialize mixin-specific attributes
+        self.initial_global_model: torch.nn.Module | None = None
+        self.initial_global_tensors: list[torch.Tensor] = []
+
+        # Call parent's init
+        try:
+            super().__init__(*args, **kwargs)
+        except TypeError:
+            # if a parent class doesn't take args/kwargs
+            super().__init__()
```
emersodb, line 59: I probably asked this before, but don't remember the answer. Under what circumstances do we get here but its okay? I.e. should we just raise the TypeError? Sorry if I have indeed asked this of you previously.

fl4health/mixins/personalized/mr_mtl.py
```diff
@@ -0,0 +1,212 @@
...
+        msg = (
+            f"Class {cls.__name__} inherits from MrMtlPersonalizedMixin but none of its other "
+            f"base classes implement FlexibleClient. This may cause runtime errors."
+        )
+        log(WARN, msg)
+        warnings.warn(msg, RuntimeWarning, stacklevel=2)
+
+    def get_global_model(self: MrMtlPersonalizedProtocol, config: Config) -> nn.Module:
+        """
+        Returns the global model to be used during Ditto training and as a constraint for the local model.
```
emersodb, line 91: during MR-MTL training? Or perhaps just "Returns the global model on client setup to be used as a constraint for the local model during training"?

emersodb: I wonder if we should rename this function to something more suggestive of the fact that we're just setting the initial global model?

fl4health/mixins/personalized/mr_mtl.py
```diff
@@ -0,0 +1,212 @@
...
+    @ensure_protocol_compliance
+    def set_parameters(
+        self: MrMtlPersonalizedProtocol, parameters: NDArrays, config: Config, fitting_round: bool
+    ) -> None:
+        """
+        The parameters being passed are to be routed to the initial global model to be used in a penalty term in
+        training the local model. Despite the usual FL setup, we actually never pass the aggregated model to the
+        **LOCAL** model. Instead, we use the aggregated model to form the MR-MTL penalty term.
+
+        NOTE; In MR-MTL, unlike Ditto, the local model weights are not synced across clients to the initial global
```
emersodb, line 136: Perhaps colon instead of semicolon? I imagine I made this typo in the MR-MTL client file as well. So if you could fix that would be great 🙂

tests/mixins/personalized/test_mr_mtl.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.train_loader, self.val_loader
+
+    def get_optimizer(self, config: Config) -> Optimizer | dict[str, Optimizer]:
+        return self.optimizers["local"]
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return torch.nn.CrossEntropyLoss()
+
+
+class _TestDittoedClient(MrMtlPersonalizedMixin, _TestFlexibleClient):
```
emersodb, line 39: Should this be "Dittoed"?

tests/mixins/personalized/test_mr_mtl.py
```diff
@@ -0,0 +1,72 @@
...
+from fl4health.mixins.core_protocols import FlexibleClientProtocol
+from fl4health.mixins.personalized import (
+    MrMtlPersonalizedMixin,
+    MrMtlPersonalizedProtocol,
+)
+from fl4health.parameter_exchange.packing_exchanger import FullParameterExchangerWithPacking
+from fl4health.parameter_exchange.parameter_packer import ParameterPackerAdaptiveConstraint
+
+
+class _TestFlexibleClient(FlexibleClient):
```
emersodb, line 25: Correct me if I'm wrong, but I think we have one of these classes for each of `test_mr_mtl.py`, `test_ditto.py`, and `test_adaptive_drift_constrained.py`. Could we consolidate them somewhere to reduce the duplicate code? We have a `tests/test_utils` for a few of these things.

tests/mixins/personalized/test_mr_mtl.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get_criterion(self, config: Config) -> _Loss:
+        return torch.nn.CrossEntropyLoss()
+
+
+class _TestDittoedClient(MrMtlPersonalizedMixin, _TestFlexibleClient):
+    pass
+
+
+class _DummyParent:
```
emersodb, line 43: Same thing here, I believe there is one of these guys for Ditto tests as well?

fl4health/mixins/personalized/mr_mtl.py
nerdai: lol I never even knew about WARNING but that's the main enum. I guess they have this alias for lazy people like me.

fl4health/mixins/personalized/mr_mtl.py
nerdai: Yeah, sorry about the copy-and-paste issues with Ditto mixin... I used your suggested summary statement.

nerdai: I think the name is technically correct though right, since we're getting the model and then using it to set the `self.initial_global_model`. Like this method returns a model, so it is in that sense a getter.

But, to your point, we don't need to make a copy of the global model here do we? Like could I just make use of `self.get_model(config)`

fl4health/mixins/personalized/mr_mtl.py
nerdai: changed in both

tests/mixins/personalized/test_mr_mtl.py
nerdai: Thanks. Changed this to something more appropriate.

tests/mixins/personalized/test_mr_mtl.py
nerdai: done!

fl4health/mixins/personalized/mr_mtl.py
nerdai: I vaguely remember this now. Will run some tests on this in the AM. I think this may just be a static type error that was causing the issue.

This would get encountered if there is a parent in the MRO that doesn't accept args, kwargs. It's probably not something though that we should encounter given these mixins are meant to be mixed with `FlexibleClient` only. So, I'd be okay with just raising the error to indicate something off to the developer.

tests/mixins/personalized/test_mr_mtl.py
nerdai: Note: in this instance with import statement style `from .conftest import ...` I actually like this here for two reasons:

1. its not really part of a library that we're importing from here
2. since multiple conftests can be defined and are automatically resolved with their hierarchical structure this indicates directly that its importing from the `conftest.py` at the same level as the test script

fl4health/mixins/personalized/mr_mtl.py
emersodb: If you're good with the name we can stick with it. The `deep_copy` that's being done in the MR-MTL client was strictly for "safety" just in case somehow the `get_model` implemented by the user actually returned an already existing object. We want the global and local models to be guaranteed separate objects. So I believe we implemented the deep copy to ensure that was the case regardless.

nerdai: Oh, I didn't see us using `deep_copy` in `MrMtlClilent`. This part came from my copy over from `DittoPersonalizedMixin` to kickstart the mixin/protocol approach for MR-MTL. We used this `deep_copy` there for Nnunet stuff, I believe.

In `MrMtlClient`, we call `self.get_model(config)` under the `setup_client()` method:

```
    def setup_client(self, config: Config) -> None:
        """
        Set dataloaders, optimizers, parameter exchangers and other attributes derived from these.
        Then set initialized attribute to True.

        Args:
            config (Config): The config from the server.
        """
        # Need to setup the init global model here as well. It should be the same architecture as the model so
        # we reuse the get_model call. We explicitly send the model to the desired device. This is idempotent.
        self.initial_global_model = self.get_model(config).to(self.device)
        ...
```

(I think in our library the `get_model()` gets implemented as a method that returns a single model i.e., in a single memory location, and other times as a method that produces a new model copy/instance i.e., in a new memory location)

emersodb (approved):

PR #558, opened 2026-05-13
Upgrading FL4Health Libraries

emersodb:
# PR Type
Other

# Short Description

Updating the libraries in the FL4Health library. In particular, this updates many of the libraries with vulnerabilities, but also finally upgrading Torch and Tensorflow with the HPC upgrades.

Note: This required a few small workarounds. One of them is a monkey patch for nnunet, of which both Marcelo and I aren't huge fans, but it seems as though they are not planning to move to newer torch implementations soon. So it felt like the least invasive approach at the moment.

# Tests Added

None added but made sure the unit and smoke tests all pass locally and on the cluster.

### DIFF
--- a/.github/workflows/static_code_checks.yaml
+++ b/.github/workflows/static_code_checks.yaml
@@ -47,31 +47,10 @@ jobs:
         uses: pypa/gh-action-pip-audit@v1.1.0
         with:
           virtual-environment: .venv/
-          # GHSA-3749-ghw9-m3mg and GHSA-887c-mr87-cxwp are pytorch vulnerabilities that require 2.7 and 2.8 but we're
-          # pinning to 2.6.0 for now.
-          # CVE-2025-53000 NBConvert issue, no fix yet.
-          # CVE-2026-21851 is a MonAI issue with no fix at the moment.
-          # CVE-2024-55459, CVE-2025-9906, CVE-2025-12058, CVE-2025-12060 are keras vulnerabilities that require
-          # keras>=3.12.0 which needs tensorflow>=2.16, but we're pinning to tensorflow 2.15 due to tensorflow-io
-          # compatibility constraints.
-          # CVE-2026-0994 is a protobuff vulnerability without a fix yet.
           # CVE-2026-26007 is a cryptography vulnerability that requires cryptography>=46.0.5, but flwr (flower)
           # requires cryptography<45.0.0, blocking the upgrade.
-          # GHSA-rf74-v2fm-23pw, CVE-2026-33230, CVE-2026-33231: NLTK vulnerabilities without a fix yet.
           ignore-vulns: |
-            GHSA-3749-ghw9-m3mg
-            GHSA-887c-mr87-cxwp
-            CVE-2025-53000
-            CVE-2026-21851
-            CVE-2024-55459
-            CVE-2025-9906
-            CVE-2025-12058
-            CVE-2025-12060
-            CVE-2026-0994
             CVE-2026-26007
-            GHSA-rf74-v2fm-23pw
-            CVE-2026-33230
-            CVE-2026-33231
 
       # Deleting some temporary files and useless folders to free up space
       # Deleting /usr/share/dotnet should clear ~4GB of space.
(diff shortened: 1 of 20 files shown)


fl4health/utils/nnunet_utils.py
```diff
@@ -518,12 +518,9 @@ def __init__(
         # Number of windows with constant LR across training
         self.num_windows = ceil(max_steps / self.steps_per_lr)
         self._step_count: int
-        super().__init__(optimizer, -1, False)
+        super().__init__(optimizer, -1)
 
-    # mypy incorrectly infers get_lr returns a float
```
emersodb, line 523: This has been fixed 🙂

fl4health/utils/nnunet_utils.py
```diff
@@ -518,12 +518,9 @@ def __init__(
         # Number of windows with constant LR across training
         self.num_windows = ceil(max_steps / self.steps_per_lr)
         self._step_count: int
-        super().__init__(optimizer, -1, False)
+        super().__init__(optimizer, -1)
```
emersodb, line 521: Torch no longer supports the verbose argument for schedulers (this is similar to the issue that was plaguing the nnunet schedulers).

.github/workflows/static_code_checks.yaml
```diff
@@ -47,31 +47,12 @@ jobs:
...
-            CVE-2024-55459
-            CVE-2025-9906
-            CVE-2025-12058
-            CVE-2025-12060
-            CVE-2026-0994
             CVE-2026-26007
-            GHSA-rf74-v2fm-23pw
-            CVE-2026-33230
-            CVE-2026-33231
+            CVE-2026-34073
```
lotif, line 56: That's good to see! 👏

fl4health/mixins/personalized/ditto.py
```diff
@@ -124,6 +124,13 @@ def _copy_optimizer_with_new_params(self: DittoPersonalizedProtocol, original_op
 
         optimizer_kwargs = {k: v for k, v in param_group.items() if k not in ("params", "initial_lr")}
         assert self.global_model is not None
+
+        # NOTE: This is a small workaround for torch back-compatibility in AdamW. Torch injects a key (that isn't part
+        # of the class signature) into the param groups called "decoupled_weight_decay" which cases an error in the
```
lotif: I think you meant "causes an error".

fl4health/mixins/personalized/ditto.py
```diff
@@ -124,6 +124,13 @@ def _copy_optimizer_with_new_params(self: DittoPersonalizedProtocol, original_op
...
+        # kwargs below. See: https://github.com/pytorch/pytorch/blob/v2.11.0/torch/optim/adamw.py#L57
+        if optim_class == torch.optim.AdamW:
+            optimizer_kwargs.pop("decoupled_weight_decay")
```
lotif, line 132: Weird that we need a workaround for their workaround. What is the case where this happens?

pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
     "pandas>=2.0,<3.0",
     "flwr>=1.18.0,<1.19.0",
     "opacus>=1.3.0,<2.0.0",
-    "torch==2.6.0",
+    "torch>=2.6.0",
```
lotif, line 22: nice 👏

lotif (approved):
Approving with a couple of minor comments :)

fl4health/mixins/personalized/ditto.py
emersodb: Indeed I do 😂

fl4health/mixins/personalized/ditto.py
emersodb: Basically, they got rid of this in the constructor, but if you follow the link, you'll see they "inject" it into the dictionary. To make this mixin work, Andrei was reconstructing the optimizer arguments to create a new one from the old one. The injection means that there is a key corresponding to the argument that used to be called `decoupled_weight_decay`. So it gets into the kwargs which are passed to the constructor which no longer wants `decoupled_weight_decay`.

The way Andrei was doing this is fairly brittle, but I don't want to mess with it much in this PR.

lotif: Yeah, sounds a little hacky, but I agree with you about not messing with it here.

